PR #99, opened 2014-09-16
pt_pt translations for LM master

gioman:


rldhont:
Apologize for this late response. This pull-request is always up-to-date ?

gioman:
The translated strings in this PR are ok.  I guess that since then that more are needed to be added, but this ones are good.

PR #181, opened 2015-07-10
AllowMulti=true for ValueRelation creates checkboxes control

nboisteault:
Here is a patch to create a checkboxes control when AllowMulti=true in QGIS for 'ValueRelation' QGISEditType.
I'm new to git and github so I hope my pull request is not to messy.


lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -797,6 +806,10 @@ public function saveFormDataToDb($form){
     foreach($fields as $ref){
       // Get and filter the posted data foreach form control
       $value = $form->getData($ref);
+      
+      if(is_array($value)){
+        $value = '{'.join($value,',').'}';
```
rldhont: It seems that join is not a PHP method. Can you correct this line with implode method.

rldhont:
Hi @nboisteault, thanks for your contribution.
With which data did you test your code ?

nboisteault:
Hi. You're welcome.
I've tested with a personal PostGIS database. My array colums are of type 'character varying' and not 'int[]' though.

PR #200, opened 2015-09-17
Cache - white list of parameters for md5 key

xcaeag:
Use a white list of parameters for the constitution of a unique  MD5 key
Between WWW and CLI cache generation, sometimes, the list differ ("exception" param)
Seen with René-Luc : The list still needed to be adapted.


lizmap/lizmap-modules/localiz/controllers/default.classic.php
```diff
@@ -0,0 +1,63 @@
...
+        return $rep;
+    }
+    
+    /**
+    *
+    */
+    function search() {
+        $localConfig = jApp::configPath('localconfig.ini.php');
+        $ini = new jIniFileModifier($localConfig);
+
```
laurentj: First, you shouldn't use jIniFileModifier to only read an ini file : it is very slow compare to parse_ini_file().

Second, all parameters into localconfig.ini.php are merged at run time with other configuration files, so in fact, you can access to your parameters with `jApp::config()->localiz_plugin` which is an array (ex: jApp::config()->localiz_plugin['db_host'])

lizmap/lizmap-modules/localiz/install/install.php
```diff
@@ -0,0 +1,49 @@
...
+        }
+        
+        // installer le JS
+        $jsFile = jApp::wwwPath('js/localiz.js');
+        if (! file_exists($jsFile)) {
+            $this->copyFile('js/localiz.js', $jsFile);
+        }
+
+        // conf DB
+        $localConfFile = jApp::configPath('localconfig.ini.php');
```
laurentj: You already have a such instantiated object in `$this->entrypoint->localConfigIni`

lizmap/lizmap-modules/localiz/install/install.php
```diff
@@ -0,0 +1,49 @@
...
+        if (file_exists($localConfFile)) {
+            $ini = new jIniFileModifier($localConfFile);
```
laurentj: You already have a such instantiated object in `$this->entrypoint->localConfigIni`.

lizmap/lizmap-modules/localiz/install/install.php
```diff
@@ -0,0 +1,49 @@
...
+            $ini->setValue('db_driver', 'pgsql', 'localiz_plugin');
+            $ini->setValue('db_host', 'pgpool', 'localiz_plugin');
+            $ini->setValue('db_port', '5432', 'localiz_plugin');
+            $ini->setValue('db_database', 'refgeo2', 'localiz_plugin');
+            $ini->setValue('db_user', 'visu', 'localiz_plugin');
+            $ini->setValue('db_password', 'visu', 'localiz_plugin');
```
laurentj: It's better to put these parameters into the profiles.ini.php (and so you don't have to create a "virtual" profil into search(), and it is more performant). And you should verify that there isn't already these parameters, to not overwrite values changed by the user.

To create a such profil, use instead the method `declareDbProfile`:

```
$this->declareDbProfile('localiz_plugin', array('driver'=>'pgsql', 'host'=>'...',...));
```

lizmap/lizmap-modules/localiz/install/install.php
```diff
@@ -0,0 +1,49 @@
...
+            $ini->setValue('db_query', "select oid, geometrytype, value, label, longlabel, xmin, ymin, xmax, ymax from MYTABLE where label like '%s%%'", 'localiz_plugin');
```
laurentj: It's ok for this parameter to put it into the localconfig.ini.php

`$this->entrypoint->localConfigIni->setValue(...);`

laurentj:
I did some comments about the use of jelix API.  @rldhont or @mdouchin will review the lizmap part :-)

lizmap/www/js/map.js
```diff
@@ -3754,9 +3754,9 @@ var lizMap = function() {
         }
       });
 
-      printLayers.reverse();
-      styleLayers.reverse();
-      opacityLayers.reverse();
+      //printLayers.reverse();
+      //styleLayers.reverse();
+      //opacityLayers.reverse();
```
rldhont: Hi @xcaeag you can't comment this lines, because this correction is available for QGIS 2.18 and not for 2.14.
The fix has to be more complex.

lizmap/lizmap-modules/localiz/controllers/default.classic.php
rldhont: @laurentj we have to check all the lizmap code for this point.

rldhont:
Hi @xcaeag,
I think some part of code has not to be part of this PR. For example the module `localiz` has to be published in a different PR.

xcaeag:
René-Luc, you can close this PR, the latest additions (localiz plugin ...) are quite specific to the water agency ...
Ignorance of git.
But take a look at the key md5 of the tiles of the cache...
thank you !

rldhont:
With WMTS support in QGIS Server 3.* we will probably changed the way to cached tiles

PR #277, opened 2016-03-30
Query external GetLegendGraphics for external WMS layers

nboisteault:


nboisteault:
I'm sorry but my master branch was not updated so I didn't see you already did that work. This pull request can be deleted so.

rldhont:
Thanks for your contribution

nboisteault:
Now I'm looking at adding GetFeatureInfo for external WMS layers. It seems that I need to add as many GetFeatureInfo control than different external url. What do you think of this idea?

PR #300, opened 2016-04-25
Proxy external WMS getFeatureInfo

nboisteault:
getFeatureInfo requests for external WMS are proxied via lizmap


nboisteault:
I'm sorry. I see that there are two commits not needed that I created to keep my master synced.
Let me know if you need a cleaner pull request.

laurentj:
Yes, it would be better to rebase your commits :-)

nboisteault:
Ok I think it is good now. See https://github.com/3liz/lizmap-web-client/pull/312

PR #305, opened 2016-05-05
Put data from qgs files to a cache

laurentj:
Reading the xml file of projects (*.qgs) takes many time (and many memory). And it is readed at each request, so for each tiles. 

Now the XML content is readed only one time, and needed data are put in a cache. Xml file is readed only if it has changed. And so, most of methods of LizmapProject don't read anymore the xml content, and just return data readed from the cache. Except some few methods that read the xml file everytime : getXmlLayers(), getXmlLayerByKeyword($key), getComposer( $title ), getXmlLayer()

### DIFF
--- a/lib/jelix-plugins/cache/file/file.cache.php
+++ b/lib/jelix-plugins/cache/file/file.cache.php
@@ -176,7 +176,6 @@ class fileCacheDriver implements jICacheDriver {
 
         $filePath = $this->_getCacheFilePath($key);
         $this->_createDir(dirname($filePath));
-
         if ($this->_setFileContent ($filePath,$var)) {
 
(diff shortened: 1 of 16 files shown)


laurentj:
- getXmlLayers(), getXmlLayerByKeyword($key), getComposer( $title ): it seems they are not used in Lizmap. Probably some methods used by some user's projects? Could we drop them? Will it break some other projects?
- getXmlLayer() : used by the edition feature. Many layer data are used by the edition feature, and I don't know yet very well what is useful to put needed data into the cache.

rldhont:
getXmlLayerByKeyword($key) is used by user's projects and could be used by us to assign properties to layers.

PR #315, opened 2016-05-18
Admin - Add parameters to choose Redis backend host and port

mdouchin:
@rldhont  locales have been created for english and french in this PR, please give guidelines to add other locales


laurentj:
Parameters in the form should be hidden if the storage type is not redis, IMHO

lizmap/modules/lizmap/classes/lizmapProxy.class.php
```diff
@@ -392,11 +392,18 @@ static public function createVirtualProfile( $repository, $project, $layers, $cr
...
+                $cacheRedisPort = trim($ser->cacheRedisPort);
+
             // Virtual cache profile parameter
             $cacheParams = array(
                 "driver"=>"redis",
-                "host"=>"localhost",
-                "port"=>"6379",
+                "host"=>$cacheRedisHost,
+                "port"=>$cacheRedisPort,
                 "ttl"=>$cacheExpiration
```
laurentj: a "db" parameter should be added (optional, integer >=0), in order to allow to select the database index in Redis.

laurentj: a "key_prefix" parameter should be added (optional, string), in order to indicate key_prefix.

mdouchin:
@laurentj  @rldhont updated with key prefix and db optional parameters + hide them in form if redis is not chosen

laurentj:
It ok for me :-)

PR #323, opened 2016-05-27
Fix issue in the admin menu: impossible to back to the parent page

laurentj:
Clicking on the link corresponding to a parent page of a section is not possible.

For example, when we click on a link on "Lizmap configuration",
then we go to modify the config or a repository, the link
"Lizmap configuration" becomes inactive and has the bad URL.

We cannot back to the lizmap configuration with the menu.


laurentj:
This behavior is not user friendly IMHO

PR #324, opened 2016-05-27
ajout du test property_exists() a proxy_external_wms

nboisteault:


rldhont:
Can you add more tests:
```
if( property_exists($pConfig->layers, $queryLayer) 
&& property_exists($pConfig->layers->$queryLayer, 'externalAccess') 
&& $pConfig->layers->$queryLayer->externalAccess == 'True' ){
```
Thanks

PR #377, opened 2016-09-20
 Popup - Warn user if no object has been found & dock not showed on small screen

josemvm:
popup.msg.no.result=Não foi encontrado nenhum objecto neste local.

also necessary for all other locales


josemvm:
sorry, according commit https://github.com/3liz/lizmap-web-client/commit/fd4e6e84cf5f5fbf61201ac51024107eda2a8db0

PR #420, opened 2016-11-26
minor fix in the default css theme

jonnyforestGIS:
Fix header url. The extra ' '  is broking the url header and use the default theme header instead of project theme header.

It's happening in the loading.gif i will try to see what is wrong.

Regards


rldhont:
Manually merged in release_3_0 with 6815eabfb0d8b828bb3aa09f354a585aa71b9f5c
Will be available in 3.0.5

PR #435, opened 2016-12-16
Update map.UTF-8.properties

majorxtrem:
Change the placeorder to "Recherche"
"Rue, Ville" is not ideal if you use quickfinder to search other things


rldhont:
Can you update it to change en_US locale ?
The en_US locale is used by other localizers on transifex.

majorxtrem:
Done !

rldhont:
Thanks!

PR #489, opened 2017-02-01
WIP: [FEATURE] Limit the amount of features per layer displayed in popup

mbernasocchi:
This is a work in progres.

The idea is to limit the ammount of features showed in the popup per each layer

@mdouchin what do you think of such a functionality? do you want me to code it further (UI part, ...)  or do you think we should not have this in Lizmap? if yes just let me know and I'll do

### DIFF
--- a/lizmap/modules/lizmap/controllers/service.classic.php
+++ b/lizmap/modules/lizmap/controllers/service.classic.php
@@ -768,8 +768,8 @@ class serviceCtrl extends jController {
 
       // Loop through the features
       // TODO make this configurable
-      $max_features_per_layer = 999999999;
-      $features_per_layer_counter = 0;
+      $popupMaxFeatures = 999999999;
+      $layerFeaturesCounter = 0;
       foreach($layer->Feature as $feature){
         $id = $feature['id'];
         // Optionnally filter by feature id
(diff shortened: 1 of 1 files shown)


mdouchin:
I like the idea, and some users have asked for it.
I think we would need to have
* a default max number, for example 10
* a way to define a max number for each layer. For example, set to 1 if the editor wants to show only one feature

These 2 vars should be configured through the plugin, in the .cfg (json) file.
There is a way in Lizmap php to use the options configured within the plugin (with a cache system).

I propose to name this vars popupMaxFeatures, one in the "options" prop and one foreach layers in the layers prop (only if needed)

Btw I need to give the plugin some love, and refactor some parts ( hanling of tables for example...)

mbernasocchi:
@mdouchin sorry I forgot this, do you have any pointers re the following?

> There is a way in Lizmap php to use the options configured within the plugin (with a cache system).

cheers
Marco

mbernasocchi:
I think a value per layer is a good idea

rldhont:
@mbernasocchi @mdouchin any idea how to fixe the popupMaxFeatures per layers ?
I'd like to introduce this feature in 3.2.

mdouchin:
I would add a new option in the plugin, in the Layers tab, in the subgroup "Popup", with a input number set to 10 by default, and a min and max value (min = 1, max = 20 ?).

rldhont:
@mbernasocchi Ihave integrated and completed your PR #716

mbernasocchi:
@rldhont wow great, sorry for the slow answer. I'm travelling.

a similar issue would be the search radius for each layer popup. is there some ideas yet?

PR #548, opened 2017-03-21
Fix visibility of some controllers methods

laurentj:
Some methods are not supposed to be called from a browser,
but are called only from other controllers methods. So they
must not be public.

### DIFF
--- a/lizmap/modules/admin/controllers/config.classic.php
+++ b/lizmap/modules/admin/controllers/config.classic.php
@@ -830,22 +830,4 @@ class configCtrl extends jController {
     return $rep;
   }
 
-  function cryptUsersPassword(){
-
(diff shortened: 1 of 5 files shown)


laurentj:
There are other methods inside lizmap~service which seem to be protected theorically: all Get* methods called from the index method. Should we keep public or not?

rldhont:
@laurentj you're right, except getProjectConfig the other Get* methods has to be protected

laurentj:
Ok, why not. But why do not create urls for each get*, instead of having a parameter for index() ?

rldhont:
Because each method matches OGC Service - Request parameters pair. These method are for convenience and has to be ported to the classes lizmap*Request.

Issue #597, opened 2017-05-24
Data Table - Field name not showing


GabrielePrestifilippo:
I've noticed that with the versions 2.18 of QGIS the field's name in the data table are not available.
![field](https://cloud.githubusercontent.com/assets/2738850/26397886/02372c1c-4078-11e7-94d5-bb5a72c55232.png)

However, selecting a field and clicking "Display Info" shows the name of the field.
Am I missing something or is it a bug?

jhcardoso:
Hi, you need to define in QGIS the alias in layer properties. Then, it will be showned in lizmap table.

GabrielePrestifilippo:
Thanks, I didn't find it anywhere in the documentation, maybe you should mention it somewhere.

gioman:
> Thanks, I didn't find it anywhere in the documentation, maybe you should mention it somewhere.

consider also doing a pull request with a patch/improvement for the docs.

rldhont:
It's an issue, if nothing is defined when alias is not defined, isn't it ?

jhcardoso:
That's right. It seems that there's no other way without the alias.

rldhont:
Which Lizmap version do you use ?

GabrielePrestifilippo:
I provided a temporary fix here:
https://github.com/3liz/lizmap-web-client/pull/598

rldhont:
manually merged in 3_0

PR #598, opened 2017-05-25
Show fields in data table

GabrielePrestifilippo:
Show fields name in data table even when aliases are not fixed
Referencing https://github.com/3liz/lizmap-web-client/issues/597


rldhont:
Does your fix has to be applied to 3.1 and 3.0 branch ?

GabrielePrestifilippo:
@rldhont Yes, it looks the problem was in both branches and the fix works on both.

rldhont:
manually merged

PR #604, opened 2017-05-31
Update jquery.combobox.js

josemvm:
- is not translated
show

- is intuitive and therefore unnecessary
- the combox of the other tools (selection, tooltip) do not have label


rldhont:
thanks

Issue #701, opened 2017-11-04
small graphical details between LWC desktop mode and mobile mode


josemvm:
hi,

I found some small graphical details:

**mobile mode:**

- the "Login" header truncates and overlaps the logo:
![image1](https://user-images.githubusercontent.com/10053874/32406325-5c83e580-c16e-11e7-854a-a9685bfe412f.jpg)
-  the "Administration" header overlaps the logo and once inside the administration menu, the strings for the logged user/"admin" and "Projects" have a different char font:
![image2](https://user-images.githubusercontent.com/10053874/32406326-5c9e16a8-c16e-11e7-8051-338c407f34d1.jpg)
- there's an empty space on left in the headers (Administration/image above, Main and Projects)
![image3](https://user-images.githubusercontent.com/10053874/32406327-5cbb2216-c16e-11e7-9426-5ca5320af3a0.jpg)
![proj](https://user-images.githubusercontent.com/10053874/32406324-5c66cf22-c16e-11e7-8457-f891ed768c33.jpg)

inside the map project the header it's ok:
![ok](https://user-images.githubusercontent.com/10053874/32406494-791578f0-c171-11e7-9d9b-f2dc3ddd48bb.jpg)

**desktop mode:**
- after login, drop down menu - behavior ok:
![login1](https://user-images.githubusercontent.com/10053874/32406750-2d560e3e-c176-11e7-9b84-dc9e5257aaab.jpg)
- after login, inside the Administration menu/My account - different char font:
![logadmi](https://user-images.githubusercontent.com/10053874/32406914-61104030-c178-11e7-9adc-a385e04abd8c.jpg)
-  **(master branch, after commit https://github.com/3liz/lizmap-web-client/commit/b83f1a59f119b8804e87b3ee5d0fbd10d2648dde)** inside the map project, after login - this behavior, 
![lll](https://user-images.githubusercontent.com/10053874/32407010-e23c30be-c179-11e7-990b-46912a02f527.jpg)

thanks in advance for your attention

rldhont:
Thanks @josemvm if you have some hack or propositions to fix them, thanks in advance.

mdouchin:
@josemvm @rldhont I just pushed a commit, not related to this, but concerning CSS. I was a bit annoyed by all the bold fonts in Lizmap UI. I removed many of them.
Do you like it ? I really do...

josemvm:
hi @mdouchin 
yes, that seems good to me!

[...]

PR #704, opened 2017-11-07
 [bugfix] small graphical details

josemvm:
a partial fix for https://github.com/3liz/lizmap-web-client/issues/701


josemvm:
- i can't solve the logo problem in this case imagem without affecting the other cases.
- i can't solve the the different font size for "Projects" and "logged user" strings inside the administration menu

i need your help
thanks

Issue #706, opened 2017-11-09
Tooltip doesn't let choose the layer and data load forever for one layer


Yolecomte:
Using LWC 3.1.0, I try to use a Tooltip on a PostgreSQL view.

I have two layer which come from a postgres view. This two layers shows different data based on filter set in the QGIS project on this view. 
The Tooltip works well for one layer called 'parcelle Vente' which get 3500 features.
But for the other layer based on the same view (called Actes administratif (6900 features)), the tooltip load forever and never let the user get the informations.

The GeoJSon send back is OK and come with a 200 status with the request send by the tooltip for both layer. 

The same issue occurs when I try to get the data attribute table for this two layers.
The attribute table show well for 'parcelle Vente' but never show up for 'Actes administratif'.

A second point is that when I click on the tool, it starts to load the first layer before I can choose for which layer I want to use the tooltip... While the layer is loading, I can't change to another layer. As the loading time may be long, it's uncomfortable that I can't choose the layer I want to use before starting loading anything.

Anything else in the project works very well and I'd like to thank's the lizmap team developpement for this great tool!

josemvm:
👍

josemvm:
![tooltip](https://user-images.githubusercontent.com/10053874/32604640-367719f8-c546-11e7-8a60-06eddfda2695.jpg)

i have tried to replace this code
`<option value="">---</option>`
by the following code
`<option disabled selected value> -- select an option -- </option>`

inside

>  lizmap-web-client/lizmap/modules/view/templates/map_tooltip.tpl

but it doesn't work...

Yolecomte:
Ok thank's for the feedback!
So I hope It will come in a next release ;)

josemvm:
as workaround, you can add

`<option style="display:none">`
after
`<option value="">---</option>`

Yolecomte:
Ok, that works well to avoid a default layer to be load when initializing the tool.
Thk's a lot for that. 

But I still have no idea about why the tool never finish to load when using my 'Actes administratifs' layer.

rldhont:
@Yolecomte have you tested with last versions ?

Yolecomte:
@rldhont Thk's for your feedback.
Answer is no... The company which I worked for when I face this problem isn't ready to upgrade for now. But I will make a try with a fresh installation on my own computer with the problematics datas. 
I give some feedbacks as soon as possible!

rldhont:
Related to #1123

[...]

PR #711, opened 2017-11-18
 tooltip - unwanted default layer

josemvm:
to prevent an unwanted default layer to be load when initializing the tool
https://github.com/3liz/lizmap-web-client/issues/706

### DIFF
--- a/lizmap/modules/view/locales/en_US/map.UTF-8.properties
+++ b/lizmap/modules/view/locales/en_US/map.UTF-8.properties
@@ -26,6 +26,7 @@ print.toolbar.format=Format
 print.toolbar.template=Template
 tooltip.toolbar.title=Tooltip
 tooltip.toolbar.layer=Highlight layer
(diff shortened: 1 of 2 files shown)


rldhont:
Thanks to trying to fixed #706, but your code is not enough to completely fixed it.
We firstly needs an option to disable or not: loading a layer at the tooltip dock displayed.
We have to enhance the JavaScript to well ended if the GeoJSON is not complete or if the request not well finished.

josemvm:
ok, for now it's just a workaround

rldhont:
Why did you close your PR ? I was thinking to reuse it and enhance. I'd like the idea to add a string in the select. Can you restore your branche and reopen it ? Thanks

josemvm:
sorry, you can reuse the code in a timely manner

rldhont:
I have opened an issue #1123 to save your solution to fix #706

PR #728, opened 2017-12-16
 attribute table - unnecessary help

josemvm:
the behavior is intuitive

i propose also this change to gain some space
imagem

or remove `<b>{@view~map.attributeLayers.options.title@}</b>`


rldhont:
@mdouchin what do you think about these commit and changes ?

mdouchin:
Ok for me, we can merge

PR #744, opened 2018-01-19
Josemvm patch 4

josemvm:
including LDAP authentication


josemvm:
also including `mainconfig.ini.php`

PR #745, opened 2018-01-19
Including ldap authentication

josemvm:
Including ldap authentication, if configured


josemvm:
i'll redo it one more time

Issue #475, opened 2017-01-19
Get Atlas for Lizmap from QGIS Print Composer


jhcardoso:
How can we configure an atlas configuration print layout to be used? 
Any extra server side code?

Thanks

rldhont:
This feature will probably need a QGIS Server plugin or a new QGIS Server feature

rldhont:
I close this issue because, the atlas is available in the popup if a plugin is installed server side. Document needs to be updated

PR #782, opened 2018-02-23
Atlas popup - Link to visualize the element in the popup for a layer with atlas

SenequeC:
This feature permits you to get a print composer for the element you selected as a link in the popup if your popup is activated for this layer.


rldhont:
Will fix #475

PR #784, opened 2018-03-02
[UI] Clear button is now in the header menu

### LINKED ISSUE #700: search address tool - clear button
the search address tool needs a clear button...

if the search tool is not configured untitled-1 the user has no way to clear the highlighted geometry

SenequeC:
The design of the clear button is now a 'refresh' icon because some feedback told us that the previous one was not always easy to understand.
There is a white version for desktop mode and black version for mobile because it's difficult to see the white one if your background is of a clear color, which is often the case. 
It should fixes #700


josemvm:
thanks @SenequeC !

PR #806, opened 2018-03-12
Add of the refresh button in the Embed controller

### LINKED ISSUE #797: carry [UI] Clear button to the [Feature] Embed controller
hi

please carry the new [UI] clear button https://github.com/3liz/lizmap-web-client/pull/784 to the [feature] embed controller https://github.com/3liz/lizmap-web-client/pull/699

imagem  ===>>> imagem

thanks

SenequeC:
Should fix #797 
I didn't changed the place of the button because its not only for the refresh of the things selected by adress but can be used with the locate too. I think people may think it was only for the adress search if it was in it, maybe i'm wrong.

### DIFF
--- a/lizmap/www/css/embed.css
+++ b/lizmap/www/css/embed.css
@@ -26,7 +26,7 @@
 #mapmenu .navbar-search {
   margin-top:1px;
   height:30px;
-  width:235px;
+  width:258px;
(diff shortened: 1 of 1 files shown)


josemvm:
that makes sense, you're right
thanks @SenequeC !

PR #822, opened 2018-03-28
Dataviz - enhancement of the pie 

SenequeC:
The pie will now work without need of rework on the table, you just have to select your X and Y field and chose an aggregate method (sum, avg, median, stddev, first, last, min, max, count) and it will automatically regroup the Y with same X according to your choice.

### DIFF
--- a/lizmap/modules/dataviz/classes/datavizPlot.class.php
+++ b/lizmap/modules/dataviz/classes/datavizPlot.class.php
@@ -375,26 +375,155 @@ class datavizPlot {
                 // Prepare an array to store features color (if any)
                 $featcolors = array();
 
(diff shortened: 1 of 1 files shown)


lizmap/modules/dataviz/classes/datavizPlot.class.php
```diff
@@ -375,23 +375,148 @@ public function fetchData($method='wfs', $exp_filter=''){
                 // Prepare an array to store features color (if any)
                 $featcolors = array();
 
+                if ( $this->type == 'pie' )
+                    {
+                        //Faire un count distinct maybe pour créer la liste si besoin uniquement, si count distinct X > count Y alors créer
```
rldhont: Translate this comment.

PR #841, opened 2018-05-02
Missing case for 'time' data type

nboisteault:


rldhont:
This update has to be done on release_3_0 branch. Then we will merge it in 3_1 then 3_2 and then master.

Issue #845, opened 2018-05-14
The geometry type does not match! Error


mathieubossaert:
Hi,

I face a bug I tried to understand by checking data in my PostGIS database... I Just want to edit PostGIS tables and for som of them, I can't save data because of this error message.
Here is my table definition. Table has been truncated.

```sql
CREATE TABLE saisie.bande_enherbee
(
    id integer NOT NULL DEFAULT nextval('saisie.objets_nid_seq'::regclass),
    login text COLLATE pg_catalog."default",
    nom text COLLATE pg_catalog."default",
    id_parcelle integer,
    geometrie geometry(LineString,2154),
    CONSTRAINT bande_enherbee_pkey PRIMARY KEY (id),
    CONSTRAINT bande_enherbee_id_parcelle_fkey FOREIGN KEY (id_parcelle)
        REFERENCES saisie.parcelle (id) MATCH SIMPLE
        ON UPDATE NO ACTION
        ON DELETE NO ACTION
);
```

Do you have any idea ? 

Huge thanks :-)

Mathieu

mathieubossaert:
Seems to be relative to this function but I can't find the problem.

https://github.com/3liz/lizmap-web-client/blob/ae24ae5b86a048cbce5acee536350c07c111b514/lizmap/modules/lizmap/classes/qgisVectorLayer.class.php#L346

mdouchin:
What is the error message please ?

mathieubossaert:
Hi @mdouchin,

"**The geometry type does not match!**". appars when I try to save the new feature.

mdouchin:
Do you have the query written in your `lizmap/var/logs/error.log` ? If so, you could try to copy paste that in your SQL tool and investigate further

mathieubossaert:
I do not have any message relative to this problem. Logs are "on" but nothing appears, neither on the database side...

mathieubossaert:
Maybe my logs are not well configured...

mdouchin:
it must an error inside Lizmap, not related to a wrong SQL syntax. Can you paste here the datasource of your layer ( from the General tab of the layer properties dialog)

mathieubossaert:
here it is : 
dbname='ecodiag' host=192.168.1.119 port=5432 user='dba' sslmode=disable key='id' srid=2154 type=LineString table="saisie"."bande_enherbee" (geometrie) sql=

mathieubossaert:
[...]

PR #849, opened 2018-05-16
Update edition.classic.php

mathieubossaert:
Fixes the issue https://github.com/3liz/lizmap-web-client/issues/845.
When two tables have the same name in two different schema, Lizmap ask the geometry_columns view to get the geometry type of the table... but get the first geometry type of both tables returned by the query, the the only one of the wanted table of the right schema.

### DIFF
--- a/lizmap/modules/lizmap/controllers/edition.classic.php
+++ b/lizmap/modules/lizmap/controllers/edition.classic.php
@@ -429,7 +429,7 @@ class editionCtrl extends jController {
           $sql.= " AND f_table_schema = " . $cnx->quote($schema);
           $sql.= " AND f_table_name = " . $cnx->quote($tableAlone);
           $res = $cnx->query($sql);
-	  $res = $res->fetch();
+          $res = $res->fetch();
           if( $res )
             $this->geometryType = strtolower($res->type);
         }
--- a/lizmap/modules/lizmap/controllers/edition.classic.php
+++ b/lizmap/modules/lizmap/controllers/edition.classic.php
@@ -364,6 +364,7 @@ class editionCtrl extends jController {
     // Set some private properties
     $this->table = $table;
     $this->tableName = $tableAlone;
+    $this->schemaName = $schema;
     $this->whereClause = $sql;
     $driver = $this->providerDriverMap[$this->provider];
 
@@ -424,7 +425,7 @@ class editionCtrl extends jController {
         // If postgresql, get real geometryType from geometry_columns (jelix prop gives 'geometry')
         if( $this->provider == 'postgres' and $this->geometryType == 'geometry' ){
           $cnx = jDb::getConnection($this->layerId);
-          $res = $cnx->query('SELECT type FROM geometry_columns WHERE f_table_name = '.$cnx->quote($this->tableName));
+          $res = $cnx->query('SELECT type FROM geometry_columns WHERE f_table_schema = '.$cnx->quote($this->schemaName).' AND f_table_name = '.$cnx->quote($this->tableName));
           $res = $res->fetch();
           if( $res )
             $this->geometryType = strtolower($res->type);
--- a/lizmap/modules/lizmap/controllers/edition.classic.php
+++ b/lizmap/modules/lizmap/controllers/edition.classic.php
@@ -426,7 +426,7 @@ class editionCtrl extends jController {
           $cnx = jDb::getConnection($this->layerId);
           $sql = "SELECT type FROM geometry_columns";
           $sql.= " WHERE 2>1";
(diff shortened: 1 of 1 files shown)


lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -364,6 +364,7 @@ private function getDataFields($datasource){
     // Set some private properties
     $this->table = $table;
     $this->tableName = $tableAlone;
+    $this->schemaName = $schema;
```
rldhont: This changes is not needed, there is already `$this->schema = $chema`

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -424,7 +425,7 @@ private function getDataFields($datasource){
         // If postgresql, get real geometryType from geometry_columns (jelix prop gives 'geometry')
         if( $this->provider == 'postgres' and $this->geometryType == 'geometry' ){
           $cnx = jDb::getConnection($this->layerId);
-          $res = $cnx->query('SELECT type FROM geometry_columns WHERE f_table_name = '.$cnx->quote($this->tableName));
+          $res = $cnx->query('SELECT type FROM geometry_columns WHERE f_table_schema = '.$cnx->quote($this->schemaName).' AND f_table_name = '.$cnx->quote($this->tableName));
```
rldhont: The better way to code it could be:
```php
        $sql = "SELECT type FROM geometry_columns";
        $sql.= " WHERE 2>1";
        $sql.= " AND (f_table_schema = ANY (current_schemas(false)) OR t_table_schema = " . $cnx->quote($schema) . ")";
        $sql.= " AND f_table_name=".$cnx->quote($tableAlone);
        $res = $cnx->query($sql);
```

lizmap/modules/lizmap/controllers/edition.classic.php
mathieubossaert: I was too tired yesterday.

lizmap/modules/lizmap/controllers/edition.classic.php
mathieubossaert: the test on the schema name `f_table_schema = ANY (current_schemas(false))` returns two rows because two tables shared the same name in two schema. This will cause the same error.
This works fine in my project : 
```php
 $sql = "SELECT type FROM geometry_columns";
        $sql.= " WHERE 2>1";
        $sql.= " AND t_table_schema = " . $cnx->quote($schema);
        $sql.= " AND f_table_name=".$cnx->quote($tableAlone);
        $res = $cnx->query($sql);
```

mathieubossaert: why the `2>1` test ?

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -424,8 +424,12 @@ private function getDataFields($datasource){
         // If postgresql, get real geometryType from geometry_columns (jelix prop gives 'geometry')
         if( $this->provider == 'postgres' and $this->geometryType == 'geometry' ){
           $cnx = jDb::getConnection($this->layerId);
-          $res = $cnx->query('SELECT type FROM geometry_columns WHERE f_table_name = '.$cnx->quote($this->tableName));
-          $res = $res->fetch();
+          $sql = "SELECT type FROM geometry_columns";
+          $sql.= " WHERE 2>1";
+          $sql.= " AND t_table_schema = " . $cnx->quote($schema);
```
rldhont: Can you change `t_table_schema` to `f_table_schema` ?

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -424,8 +424,12 @@ private function getDataFields($datasource){
...
+          $sql.= " AND f_table_name = " . $cnx->quote($tableAlone);
+          $res = $cnx->query($sql);
+	  $res = $res->fetch();
```
rldhont: Can you align the code ?

lizmap/modules/lizmap/controllers/edition.classic.php
mathieubossaert: sorry. Done.

lizmap/modules/lizmap/controllers/edition.classic.php
mathieubossaert: should be ok now.

rldhont:
Last request, can you do a rebase ?

rldhont:
Manually merged. Thanks !

Gustry:
Side question, why do you need to rebase and why don't you use the squash button on github?

PR #867, opened 2018-05-31
[Edition] Be more permissive for text and less for boolean

rldhont:
The way that lizmap was verifying the value for text and boolean was too
simple and strict. If the value is Null or empty, NULL is stored. This test
excluded to store '0' and 0.0.

The fix adds more specific tests.

### DIFF
--- a/lizmap/modules/lizmap/controllers/edition.classic.php
+++ b/lizmap/modules/lizmap/controllers/edition.classic.php
@@ -327,7 +327,7 @@ class editionCtrl extends jController {
       $tableAlone = $exp[1];
       $schema = $exp[0];
     }
-    
+
     // Set some private properties
     $this->schema = $schema;
     $this->table = $table;
@@ -961,8 +961,16 @@ class editionCtrl extends jController {
               $value = 'NULL';
             break;
           case 'text':
+            if ( is_null($value) or strlen((string)$value) == 0)
+              $value = 'NULL';
+            else
+              $value = $cnx->quote($value);
           case 'boolean':
-            if ( !$value or empty($value))
+            $strVal = strtolower($value);
+            if($strVal != 'true' &&  $strVal !== 't' && intval($value) != 1 &&
+               $strVal !== 'on' && $value !== true &&
+               $strVal != 'false' &&  $strVal !== 'f' && intval($value) != 0 &&
+               $strVal !== 'off' && $value !== false)
               $value = 'NULL';
             else
               $value = $cnx->quote($value);


rldhont:
@laurentj can your review this PR.
If you validate it. I will merge it in 3.0 and 3.1, and I manually merge it in 3.2.
Thanks

rldhont:
@mdouchin what do you thonk about these changes ?

laurentj:
```public function saveFormDataToDb($form)``` : it should be a protected/private method, not public

File: lizmap/modules/lizmap/controllers/edition.classic.php [hunk 1]
```diff
@@ -961,8 +961,15 @@ public function saveFormDataToDb($form){
               $value = 'NULL';
             break;
           case 'text':
+            if ( !$value or strlen(strtolower($value)) == 0)
```
laurentj: why strtolower ?

If $value='0', here 'null' will be stored. Are you sure you want this result?

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -961,8 +961,15 @@ public function saveFormDataToDb($form){
...
+              $value = 'NULL';
+            else
+              $value = $cnx->quote($value);
           case 'boolean':
-            if ( !$value or empty($value))
+            if(strtolower($value) != 'true' &&  $value !== 't' &&
```
laurentj: if $value == 'T', it won't pass the test. Probably you want to do strtolower before any tests...

lizmap/modules/lizmap/controllers/edition.classic.php
rldhont: if we have `'null'` it's because the user has wrote `null` in the HTML form isn't it ?

lizmap/modules/lizmap/controllers/edition.classic.php
rldhont: yes, I will add it for boolean.

rldhont:
> public function saveFormDataToDb($form) : it should be a protected/private method, not public

Yes but we done it in the futur 3.2.

lizmap/modules/lizmap/controllers/edition.classic.php
laurentj: if the user enter `0`, the result of !$value is true, so 'null' will be stored.

rldhont: so we have to use is_null ?

laurentj:
@rldhont ok for me

PR #869, opened 2018-06-01
[BUGFIX] disable filter on layer without geometry doesn't work

nboisteault:
A layer without geometry is not referenced in `lizMap.map` so `layer` variable is null and
 `config.layers[featureType]['request_params']['filter'] = null;`
`config.layers[featureType]['request_params']['exp_filter'] = null;`
is not executed.


rldhont:
thanks

Issue #902, opened 2018-06-23
Support ZM and Z values


Gustry:
### Is your feature request related to a problem? Please describe

When data are stored in Postgresql, lizmap is using the `geometry_columns` view to fetch informations about the geometric field:
https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/classes/qgisVectorLayer.class.php#L221

But if a layer contains a Z or ZM values in the geometry, then the `geometry_column` is not giving enough information.

For instance:
```sql
CREATE TABLE field_test
(
    SERIAL integer, geom geometry(Point,32620) NOT NULL,
    geom_z geometry(PointZ,32620) NOT NULL,
    geom_m geometry(PointM,32620) NOT NULL,
    geom_zm geometry(PointZM,32620) NOT NULL
);
SELECT f_table_name, f_geometry_column, coord_dimension, type from geometry_columns WHERE f_table_name = 'field_test';
```
is giving me:
```
"field_test"	"geom"	2	"POINT"
"field_test"	"geom_z"	3	"POINT"
"field_test"	"geom_m"	3	"POINTM"
"field_test"	"geom_zm"	4	"POINT"
```
As we can see, whenever it's a `Point`, or `PointZ` or `PointZM`, in the view it results in a `POINT`.
Lizmap is failing as soon as we try to save a new feature in the DB because one dimension is missing.

### Describe the solution you'd like

Lizmap should detect correctly the dimension of the field and apply `ST_Force_3DZ` `ST_Force_3DM` or `ST_Force_4D` in the same way as it's already done for `ST_Multi` in https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/classes/qgisVectorLayer.class.php#L362

### Describe alternatives you've considered

Using this SQL query is giving the correct geometry type:
```sql
SELECT a.attnum
      ,a.attname                            AS name
      ,format_type(a.atttypid, a.atttypmod) AS typ
      ,a.attnotnull                         AS notnull
      ,coalesce(p.indisprimary, FALSE)      AS primary_key
      ,f.adsrc                              AS default_val
      ,d.description                        AS col_comment
FROM   pg_attribute    a 
LEFT   JOIN pg_index   p ON p.indrelid = a.attrelid AND a.attnum = ANY(p.indkey)
LEFT   JOIN pg_description d ON d.objoid  = a.attrelid AND d.objsubid = a.attnum
[...]

PR #904, opened 2018-06-26
force geometry to Z and/or M when saving from lizmap

Gustry:
Very hard to switch back to PHP after many years in Python ...

Fix for #902 

Should I open this PR against 3.1, 3.2 or master? I think 3.2 is ok because it's still not release and the current implementation might be seen fragile.


Gustry:
After a discussion with @mdouchin this morning, indeed, it's not enough with this PR to support Z or M values because lizmap is missing the original Z or M values (especially when editing an existing feature).

But with this PR, at least the geometry has a Z or M value if the database needs it. (it will just set M or Z to 0 if needed).

rldhont:
@Gustry can you add some documentation about this point ?

Gustry:
Yes, adding it to my todolist.
Also a PR for lizmap plugin

PR #945, opened 2018-08-21
[BUGFIX] Add LDAP authenticated user to default groups even if not defined in LDAP

### LINKED ISSUE #926: ldapdao - add users to default "users" group, and do not remove them on group sync
When using ldapdao module, with the group synchronization, Lizmap remove the user from the default "users" group. This

This "users" group should be considered as the "admins" group, and not subject to synchronization. Without this feature, the ldap administrator should create a "Lizmap users" ldap group, and put ALL the users in his organization in it, which is a bit cumbersome.

We need this "users" default group, to be able to set rights to logged users.

rldhont:
Fixed #926 ldapdao - add users to default "users" group, and do not remove them on group sync

Need to be backported to ldapdao jelix module.


laurentj:
Ok for me

rldhont:
Thanks

Issue #980, opened 2018-10-05
PHP/Lizmap technical errors when list of layers in *.qgs.cfg does not correspond to list of layers into *.qgs


laurentj:
### What is the bug?

If the `.qgs.cfg` file, generated by the lizmap plugin in Qgis, has not been updated after a layer has been removed with Qqgis (so from the .qgs file), the removed layer is still shown into lizmap. And when we launch feature like edition on this layer, we have errors shown in Lizmap like 

```
Fatal error: Call to a member function getXmlLayer() on null in /lizmap_web_client/lizmap/modules/lizmap/controllers/edition.classic.php on line 141
```

### Steps to reproduce the issue

1. Create two layers in Qgis, and save the qgis project
2. Launch the lizmap plugin, select these layers, and save
3. Remove a layer with Qgis and save the qgis project
4. copy the qgs and ggs.cfg file into a lizmap project
5. in the editing panel, see both layers into the list of layers
6. select the removed layer, and click on "add"
7. an error like above appear 

### Expected behavior

In the editing panel, we should not see the removed layer. And if lizmap receive an id of an unexisting layer, it should trigger a user friendly error.

I suggest to add checks into lizmap : before using the list of layers indicating into the qgs.cfg file, lizmap should remove from this list, layers that are not in the list of layers defined into the qgs file.

### Environment

 - Lizmap version: 3.2
 - QGis version: 2.8.17
 - OS (windows, linux, macOS, android..): linux
 - Browser (Firefox, Chrome...): Firefox
 - Lizmap data stored into Sqlite (default)? Postgresql? Mysql?: Postgresql

rldhont:
@laurentj does this issue still exist ?
Has it been fixed by #982 ?

rldhont:
@nboisteault can you check this issue ? I'd like to know if it ahs been fixed. Thanks

nboisteault:
@rldhont It is fixed.
![Couche_non_editable](https://user-images.githubusercontent.com/2145040/55219534-72738c00-5205-11e9-9197-ad60a25a4880.png)

PR #982, opened 2018-10-05
Fix: don't show a technical error when the layer to edit does not exist

laurentj:
The error is more friendly: "the layer is not editable".

Refs #980


laurentj:
I'm not sure the error message is the good one...

PR #989, opened 2018-10-14
Support of proxy parameters for http requests to external services

### LINKED ISSUE #932: Search location doesn't work behind a webproxy
### What is the bug?

If lizmap is behind a web proxy for http requests, location searching is KO, with PHP or Curl.

### Expected behavior

image

### Screenshots and logs

image

Works with Curl by adding those lines in lizmap/controllers/osm.classic.php, line 46
    curl_setopt($curl_handle, CURLOPT_PROXY, '_myProxyHost:myProxyPort_');
    curl_setopt($curl_handle, CURLOPT_NOPROXY, '_mydomain.com,localhost_');

Probably same issues with functionnalities using getRemoteData() in lizmap/classes/lizmapProxy.class.php

### Environment

 - Lizmap version: 3.1.12
 - QGis version: 2.14 (N/A)
 - OS (windows, linux, macOS, android..): Linux
 - Browser (Firefox, Chrome...): N/A
 - Lizmap data stored into Sqlite (default)? Postgresql? Mysql? N/A

laurentj:
There are new possible parameters into lizmapConfig.ini.php, allowing to configure a proxy for requests  to external services. These parameters can be changed in the configuration form of Lizmap.

Some refactorizations have been made in order to call external services only from a single method, `lizmapProxy::getRemoteData()`, instead of having curl calls in many places in the code.

Fixes #932

### DIFF
--- a/lib/jelix/plugins/tpl/html/block.ifctrl_value.php
+++ b/lib/jelix/plugins/tpl/html/block.ifctrl_value.php
@@ -1,0 +1,44 @@
+<?php
+/**
+ * @package     jelix
+ * @subpackage  jtpl_plugin
+ * @author      Laurent Jouanneau
+ * @copyright   2018 Laurent Jouanneau
+ * @link        https://jelix.org
+ * @licence     GNU Lesser General Public Licence see LICENCE file or http://www.gnu.org/licenses/lgpl.html
+ */
+
+/**
+ * a special if block to test easily the current control value
+ * TO BE USED inside a {formcontrols} block
+ *
+ * {ifctrl_value 'name', 'expected-value'} some tpl {else} some other tpl {/ifctrl_value}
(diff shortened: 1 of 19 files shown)


laurentj:
Probably I have to make an improvement: as getRemoteData() is used for external services, but also for WMS adn Qgis server, if the proxy parameters are enabled, the proxy will be used for requests for WMS and Qgis. Except if the domain name of WMS and QGis server are set into requestProxyNotForDomain.

Is the parameter requestProxyNotForDomain enough? Or should I disable the use of the proxy for request to QGIS?

laurentj:
What about url of externalAccess in qgis layers?

rldhont:
For QGIS Server, the proxy information has to be added QGIS2.ini config file.

PR #1015, opened 2018-11-23
add translation for "connecter" button on sign in page

josemvm:
add translation for "connecter" button on sign in page


laurentj:
Thank you ! :-)

PR #1038, opened 2018-12-20
fix german typo

mbernasocchi:


laurentj:
Thank you. Fixed in Transifex : https://www.transifex.com/3liz-1/lizmap-locales/translate/#de/viewpot/81053776

If you want to fix some translations, please use Transifex ;-)
We don't accept anymore translations through Github. Thank you for your help!

PR #1047, opened 2019-01-09
fix popups that use QGIS maptip

mbernasocchi:
Currently Lizmap connected to a QGIS 3.x server always show popups in auto mode. This is a **regression** in lizmap because it didn't catchup with a change in QGIS GetFeatureInfo implementation.

QGIS 3.x fetches geometry and maptip in feature info only if they are requested by the client.
See [QGIS](https://github.com/qgis/QGIS/commit/5bd63b7e1789a72c2692d8d61fab267ec2258eb5
) for the commit and [QWC2](https://github.com/qgis/qwc2-demo-app/issues/55) for a discussion.

This patch restores the functionality of using QGIS' maptip as popup content


mbernasocchi:
Great!

PR #1050, opened 2019-01-10
PHP 5.6 not maintained => propose 7.2 as default

nboisteault:


nboisteault:
PR to recreate for lizmap 3.3.

laurentj:
I don't agree with this change, because :

- PHP 5.6 is not maintained anymore by the PHP core developers, but is still maintened by Debian mainteners until stretch will be dead
- Lizmap 3.2 is supposed to be work with PHP 5.6, so contributors should test for PHP 5.6. So it should be the default version.

For futur version, we could says that Lizmap will not be tested against PHP 5.6 anymore, and the official support will be 7.2.

PR #1066, opened 2019-01-30
Support of the multiauth module

laurentj:
I developed a new module,  [multiauth](https://github.com/jelix/multiauth-module), that will replace the ldapdao module, and which allow to authenticate against several login/password authentication providers, which can be ldap or a database for the moment (whereas ldapdao is only for ldap).

Some changes have been made into Jelix and into the jCommunity module in order to support this new module.

The proposed changes include also some changes into the lizmap configuration to ease the multiauth module installation. So then, to use the multiauth module:

1. install the source of the multiauth module into lizmap-modules
2. enable the module into the localconfig.ini.php `multiauth.access=1`, and disable the ldapdao module if already in use
3. configure a profile for ldap into profiles.ini.php : same parameters as ldapdao

I do not include the multiauth module sources, because it probably need more tests before distributing it. I will propose a patch for the master branch that will replace the ldapdao module by multiauth.


mdouchin:
Tested with 2 online test LDAP server:

```ini

; LDAP test server : https://www.zflexldapadministrator.com/index.php/blog/82-free-online-ldap#!kmt-start=10
[ldap:ldap_test]
hostname="www.zflexldap.com"
port=389
adminUserDn="cn=ro_admin,ou=sysadmins,dc=zflexsoftware,dc=com"
adminPassword="zflexpass"
searchUserBaseDN="dc=zflexsoftware,dc=com"
searchUserFilter="(uid=%%LOGIN%%)"
bindUserDN="uid=%%LOGIN%%,ou=users,ou=guests,dc=zflexsoftware,dc=com"
searchAttributes="uid:login,givenName:firstname,sn:lastname,mail:email,cn:,name:"
searchGroupFilter=""
searchGroupProperty="cn"
searchGroupBaseDN=""

; LDAP test server : http://www.forumsys.com/tutorials/integration-how-to/ldap/online-ldap-test-server/
[ldap:ldap_test_2]
hostname="ldap.forumsys.com"
port=389
adminUserDn="cn=read-only-admin,dc=example,dc=com"
adminPassword="password"
searchUserBaseDN="dc=example,dc=com"
searchUserFilter="(uid=%%LOGIN%%)"
bindUserDN="uid=%%LOGIN%%,dc=example,dc=com"
searchAttributes="uid:login,givenName:firstname,sn:lastname,mail:email,cn:,name:,dn:"
searchGroupFilter="(&(objectclass=groupOfUniqueNames)(uniqueMember=%%dn%%))"
searchGroupProperty="cn"
searchGroupBaseDN="dc=example,dc=com"
```
It works well for authentification, for 
* default lizmap jauth database
* LDAP ldap_test
* LDAP auth ldap_test_2

I have not yet been able to synchronize the groups "scientists" and "mathematicians" for the ldap_test_2 (forumsys) server. I will look deeper in the code.

ldapsearch works to retrieve information:

```bash
ldapsearch -h ldap.forumsys.com  -D "cn=read-only-admin,dc=example,dc=com" -b "dc=example,dc=com" -x -w "password"  -s sub '(&(objectclass=groupOfUniqueNames)(uniqueMember=uid=einstein,dc=example,dc=com))'
```

mdouchin:
Tested in another context, with ldapdao previously used. There is a change in the behaviour, compared to ldapdao : 

* in ldapdao, the user could connect with lowercase login or with UPPERCASE login (server = Microsoft AD). 
* Now with this new multiauth module, the users should respect the letter case. For example : I could connect with login MDOUCHIN, but not with mdouchin: the error was

```bash
 Erreur dans la requête (ERREUR:  la valeur d'une clé dupliquée rompt la contrainte unique « jlx_u$
DETAIL:  La clé « (usr_login)=(MDOUCHIN) » existe déjà.
```

It seems the getByLogin method in the auth dao has changed, and the module tries to INSERT again the already existing user.

NB: In the database, all the users login (already created by ldapdao module) where stored UPPERCASE.
NB : I can get around this bug with this kind of change in multiauth module, which obviously is very temporary

```diff
diff --git a/multiauth/plugins/auth/multiauth/multiauth.auth.php b/multiauth/plugins/auth/multiauth/multiauth.auth.php
index 3717971..4eb1ca0 100644
--- a/multiauth/plugins/auth/multiauth/multiauth.auth.php
+++ b/multiauth/plugins/auth/multiauth/multiauth.auth.php
@@ -213,6 +213,9 @@ class multiauthAuthDriver extends jAuthDriverBase implements jIAuthDriver2
         $dao = jDao::get($this->_params['dao'], $this->_params['profile']);
         $user = $dao->getByLogin($login);
         $createdUser = false;
+if(!$user){
+  $user = $dao->getByLogin(strtoupper($login));
+}
         if (!$user) {
             $createdUser = true;
             $user = $this->createUserObject($login, $password);
```

mdouchin:
Last thing I needed to do was to replace manually (SQL query) all the `usr_password` in jlx_user table

```sql
UPDATE jlx_user 
SET usr_password = '!!multiauth:ldap:ldap_cd30!!' 
WHERE usr_password = '!!ldapdao password!!'
```

laurentj:
>It seems the getByLogin method in the auth dao has changed, and the module tries to INSERT again the already existing user.

No, nothing has changed. Ldap code is 95% the same code as in ldapdao. The dao is the same (the lizmap one) so getByLogin() is the same.

The change is in the order of calling things. 

In multiauth, we're trying first to retrieve the database user with the login given by the user,  because we need the user record to know which provider we should use, and then we're trying to login with the ldap. 

In ldap dao, we try to login to the ldap FIRST, then we're search an existing user into the database. However, we do not the search with the login given by the user, but with the login stored into the user object. Unfortunately, (for multiauth), the login property is filled with the ldap attribute 'uid' (see into your searchAttributes). And I guess that the uid into this ldap is always in uppercase.

So, in ldapdao
- it searches into the ldap the user with the login given by the user
- it retrieves ldap attributes : the login becomes the uid from ldap attributes, which is in uppercase
- at the first authentication it stores the login (which is then in uppercase) into the database
- at next authentication, no problem to retrieve the user record, because the login = the uid given by ldap
Note: if there is not this `uid:login` mapping, you have the same issue as into ldap dao, because the search into the database is then made with the login given by the user. So if the user gives the login in lowercase at the first time, then in uppercase at the second time, the issue appear.

In multiauth:
- first authentication : it searchs the user into the database with the login given by the user, but not found.
- it searchs the user into the ldap. The uid:login mapping is made, then the login is in the uppercase. A record is created with the login in uppercase.
- at next authentication, if the login is not given in uppercase the user is not found in the database, but the ldap found him, and the login become uppercase because of the uid mapping. Because the user was not found in the database, it tries to insert a new record with the login in uppercase, but it fails, because it already exists :-/

Your patch with the `uppercase()` will work only in your case, because you have this `uid:login` mapping and because uid is in uppercase.

I must found an other solution.

>Last thing I needed to do was to replace manually (SQL query) all the `usr_password` in jlx_user table

I will made a fix, so it is not needed. And you SQL query may be wrong if some users are not in the ldap you indicate.

mdouchin:
Thanks for your detailed explanation and help

laurentj:
Issues are related to the mulitauth module itself, so I think it is safe to merge this PR

PR #1092, opened 2019-03-04
Fix edition when featureId contains multiple id

laurentj:
At some places, we should use the text version of featureId instead of the PHP version.
Else if featureId is an array, we have some PHP notices and bad values : "Array" instead of a string with integers separated by coma.

### DIFF
--- a/lizmap/modules/lizmap/controllers/edition.classic.php
+++ b/lizmap/modules/lizmap/controllers/edition.classic.php
@@ -23,10 +23,10 @@ class editionCtrl extends jController {
   /** @var string layer name (<layername> in QGIS project) */
   private $layerName = '';
 
-  /** @var string featureId parameter from the request */
+  /** @var string featureId parameter from the request, an integer or a string with coma separated integers */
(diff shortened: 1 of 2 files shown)


lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -168,7 +169,7 @@ private function getEditionParameters($save=false){
 
 
     // feature Id (optional, only for edition and save)
-    if(preg_match('#,#', $featureIdParam))
+    if (strpos($featureIdParam, ',') !== false)
       $featureId = preg_split('#,#', $featureIdParam);
```
rldhont: Il faudrait aussi ajouter:
```php
    if (strpos($featureIdParam, '@@') !== false)
      $featureId = preg_split('#@@#', $featureIdParam);
```

laurentj: ok, I added this change

rldhont:
thanks!

PR #1098, opened 2019-03-06
child objets in popup with QGIS server 3. Fix #1024

### LINKED ISSUE #1024: child objets in popup with QGIS server 3
### What is the bug?
with QGIS Server 3, childs elements are not displayed in the popup  of the lizmap client.
The response of the second GetFeatureInfo request (for the childs) is empty.

### Environment

 - Lizmap version: 3.2rc6
 - QGis version: 3.2
 - QGIS Server version: docker => 3liz/qgis-map-server:1.1
 - OS (windows, linux, macOS, android..): 
 - Browser (Firefox, Chrome...):
 - Lizmap data stored into Sqlite (default)? Postgresql? Mysql?

nboisteault:
I am not sure if it is ok to put EPSG:3857 or we should take base layer projection in case people use other projection than this.

### DIFF
--- a/lizmap/www/js/map.js
+++ b/lizmap/www/js/map.js
@@ -3644,6 +3644,7 @@ var lizMap = function() {
                             ,'STYLES': ''
                             ,'SERVICE': 'WMS'
                             ,'VERSION': '1.3.0'
+                            ,'CRS': 'EPSG:3857'
                             ,'REQUEST': 'GetFeatureInfo'
                             ,'EXCEPTIONS': 'application/vnd.ogc.se_inimage'
                             ,'FEATURE_COUNT': popupMaxFeatures


pcav:
IMHO better having other projections as well

rldhont:
@pcav has the fix have to be done in QGIS Server ?

nboisteault:
I think projection is not important in this case because we are just requesting for child objects using a filter. QGIS Server ask for CRS but it should be mandatory only when using I and J parameters.

lizmap/www/js/map.js
```diff
@@ -3644,6 +3644,7 @@ var lizMap = function() {
                             ,'STYLES': ''
                             ,'SERVICE': 'WMS'
                             ,'VERSION': '1.3.0'
+                            ,'CRS': 'EPSG:3857'
```
rldhont: you probably can use `rConfigLayer.crs` and if it is not defined `EPSG:4326`

pcav:
Thanks for clarifying. So my comment is not relevant.

nboisteault:
In GetFeatureInfo part of WMS specification, it is not clear if CRS is mandatory. It only says 'map request part' is mandatory saying : 

> The mandatory “map request part” entry in Table 9 represents a sequence of parameters from the GetMap
request that generated the original map. Two of the GetMap parameters are omitted because GetFeatureInfo
provides its own values: VERSION and REQUEST. The remainder of the GetMap request shall be embedded
contiguously in the GetFeatureInfo request.

And of course CRS is mandatory in GetMap request. So in QGIS Server should we apply specification or be 'smarter' and accept that CRS can be omitted in case I and J parameters are not set ?

rldhont:
CRS can be ommitted in case I and J parameters are not set

rldhont:
Replaced by #1099

PR #1104, opened 2019-03-09
Support of HTML wysiwyg editor in feature forms

laurentj:
When a feature attribute content HTML, it is better to use a wysiwyg HTML editor than a simple text editor. 

The proposed editor is CKEditor5, which is better and more modern than the one embedded into Jelix.


laurentj:
A screenshot to see how it is shown

lizmap-htmleditor

mdouchin:
Ok for me. This is a great addition indeed.

PR #1109, opened 2019-03-16
Show project switcher

mbernasocchi:
This allows always enabling the project switcher even when the onlyMaps setting is OFF.

if `show projects switcher` is set to `OFF`, Lizmap behaves as it alwasy did.
if `show projects switcher` is set to `ON` and `Maps Only`is set to `ON` Lizmap behaves as it alwasy did.
if `show projects switcher` is set to `ON` and `Maps Only`is set to `OFF` Lizmap has now a new dock item as seen below but with this icon:
image

image
Peek 2019-03-16 03-03


lizmap/www/css/map.css
```diff
@@ -127,7 +136,7 @@ CSS Sprite positions calculated via http://instantsprite.com/ or https://draeton
 #mapmenu .nav-list > li.map.active .icon,
 #mapmenu .nav-list > li.map a:hover .icon {
   background-position: -40px -60px;
-}
+}li.geololi.geolo
```
laurentj: I think you should revert this change...

laurentj:
First, we should not have this "home" icon for this feature (even for when "Only Maps" is "on", it does not make sens for me)

Second, if I understand well, you want to have always this  project switcher. In this case, why to have this flag? IMHO, we should rethink the project swicher and the "Only Maps" feature. The panel is horrible when there are more than 2 or three projects..

@rldhont your thoughts?

rldhont:
@laurentj the 3 ways to navigate between maps are requested by users. I think this feature is interesting.

@mbernasocchi I'd like the directory icon instead of home for the dock projects item.

We probably have to enhance the way projects are displayed in the dock item.

lizmap/www/css/map.css
mbernasocchi: ups

mbernasocchi:
> First, we should not have this "home" icon for this feature (even for when "Only Maps" is "on", it does not make sens for me)
> 
Agreed
> Second, if I understand well, you want to have always this project switcher. In this case, why to have this flag? 

No, if you set the flag to ON or if you set only maps, you get the project switcher.

> IMHO, we should rethink the project swicher and the "Only Maps" feature. The panel is horrible when there are more than 2 or three projects.

more than agreed, but with some jQuery (which I can share) loaded on uicreated , I made a really nice selector
selector

mbernasocchi:
> @mbernasocchi I'd like the directory icon instead of home for the dock projects item.
> 
the PR already includes the directory icon.
image
> We probably have to enhance the way projects are displayed in the dock item.

laurentj:
>No, if you set the flag to ON or if you set only maps, you get the project switcher.

Why having a flag ? Why not always having this project switcher ?

rldhont:
@laurentj because, in some maps, build as an Application you probably don't want to navigate through other maps.

Issue #1069, opened 2019-02-06
0 is like null


gcalibris:
In postgresql table I have the field temperatura with type integer and with NOT NULL.
In this field from QGIS program I can write the temperature value equal to 0, but from lizmap I can't write the value 0.

The log is: An error has been raised when saving form data edition to db : invalid query (ERROR:  null value in column "temperatura" violates not-null constraint.

software
qgi project from 2.18 qgis
lizmap web client 3.2.1
qgis-server 2.18.27+24xenial

Somebody can help for this issue?

rldhont:
@gcalibris can you test https://github.com/3liz/lizmap-web-client/pull/1111 or actual release_3_2 branch ?

rmarzocchi84:
I have the same problem using
- QGIS server 3.4.8 
- dev lizmap release (git pull from repository today) 

In my case I have a value map on QGIS for a not boolean field which cannot be a boolean.
The QGIS project seem OK
```


          <editWidget type="ValueMap">
            <config>
              <Option type="Map">
                <Option name="map" type="List">
                  <Option type="Map">
                    <Option value="0" name="No" type="QString"/>
                  </Option>
                  <Option type="Map">
                    <Option value="1" name="Si" type="QString"/>
                  </Option>
                </Option>
              </Option>
            </config>
          </editWidget>


 
```

nboisteault:
I can confirm bug on release_3_2.

nboisteault:
@rmarzocchi84 please open an issue if you have a similar behavior with lizmap 3.3rc2 and QGIS 3.4.x. Thanks.

Issue #958, opened 2018-09-07
Edition - Error when saving an uneditable empty boolean field


mdouchin:
### What is the bug?

When adding (creation) a new feature in a edition layer with a boolean field (with no default value set in database), Lizmap returns an error like `La saisie de "actif" est invalide`.
This field is not editable, so must not be clickable by the user

1. It seems the user can click it when he shoud not
2. the value of the field is set to 't' but should not be set. There is no default value in the databse. So the value should not be set to 't' nor 'f'
3. When saving again, without clicking or not the checkbox, the feature is saved.
It seems the read-only class is not set evenly the 1st and 2nd time.

Workaround : add a default value for this field in the database. It is cumbersome and not always wanted

nboisteault:
- Also the error message is not returned when value for checked/unchecked are 1/0 in QGIS.

mdouchin:
With @laurentj we decided to remove the disabled fields from the query (INSERT or UPDATE) sent to the PostgreSQL database. We think that if the editor has disabled the field (for example to allow to see it in the form, contrary to the hidden fields which are not visible), it should not be modified. 

For new features (INSERT query), the editor should add a DEFAULT VALUE if the field is set NOT NULL. Otherwise an error would be raised (which is not considered a Lizmap bug in this situation)

rldhont:
@mdouchin can you test https://github.com/3liz/lizmap-web-client/pull/1111 or actual release_3_2 branch ?

rldhont:
@mdouchin has been fixed ?
Can be closed ?

nboisteault:
@rldhont Not sure. I will do some tests later.

nboisteault:
Only 1. has been fixed.

nboisteault:
Please reopen if issue still occurs.

nboisteault:
@mdouchin Please reopen if issue still occurs.

Issue #572, opened 2017-04-27
field not null (edit widget photo/file name) - wrong error message if it is empty


josemvm:
if the photo field is empty, the message should be this:

![field](https://cloud.githubusercontent.com/assets/10053874/25481390/dba42232-2b44-11e7-9c9e-a1f983073b3e.jpg)

not this general:
![error](https://cloud.githubusercontent.com/assets/10053874/25481509/656deb1a-2b45-11e7-9958-8c7813738030.jpg)

thanks

josemvm:
some behavior for date widget

rldhont:
@josemvm can you test https://github.com/3liz/lizmap-web-client/pull/1111 or actual release_3_2 branch ?

josemvm:
hi @rldhont , the problem remains

laurentj:
I tried to fix the current code, but it is complicated to modify it to fix this bug, because of some limitation of jForms (the form component of Jelix). It is better to encapsulate all the logic of the upload field into a jForms plugin, to benefit of the feature of jForms, and so to avoid kind of bugs like this.

Hopefully, a plugin implementing the logic of the current code of upload, has been added to Jelix recently. We should use it and replace all "home made" code for upload. But I prefer to do this replacement into a non-stable version (3.3), to avoid regression in the 3.2 version.

laurentj:
fixed.

PR #1111, opened 2019-03-18
Fix qgisform null value

laurentj:
After the submit, qgisform convert some empty values '' to 0 instead of null. it causes issues when the value represents a foreign key. In this case, most of time 0 is not valid...


rldhont:
Does it fix some issues:
* 0 is like null #1069
* Edition - Error when saving with an empty boolean field #958
* field not null (edit widget photo/file name) - wrong error message if it is empty #572

laurentj:
@rldhont these issues are the same kind of bugs, but are not related directly to this one. 

However, I will check them.

PR #1115, opened 2019-03-19
[Bugfix] Print with QGIS Server 3.0 and Project 2.18

rldhont:
Layout maps now use a string UUID as "id", let's assume that the first map has id 0 and so on ...


lizmap/modules/lizmap/classes/lizmapProject.class.php
```diff
@@ -893,7 +893,13 @@ protected function readPrintCapabilities($qgsLoad, $cfg) {
                                     $ptMap['grid'] = 'True';
                                 }
                             }
-
+                            $services = lizmap::getServices();
+                            // In QGIS 3.*
+                            // Layout maps now use a string UUID as "id", let's assume that the first map
+                            // has id 0 and so on ...
+                            if( $services->intQgisServerVersion() >= 300 ) {
```
laurentj: I suggest to use [version_compare](http://php.net/manual/en/function.version-compare.php) instead of this custom function.

rldhont: I don't know this PHP function.

rldhont: fix

Issue #1065, opened 2019-01-29
Minidock popup with wrong dimensions


tethysgeco:
@pcav 

### What is the bug?
Sometimes (apparently with no logic) the popup (minidock type) has wrong size with respect to the content of compiled fields (both using Chrome and Mozilla Firefox - see pictures for examples). Do you have ever experienced something similar and have any idea of the reasons?  
![1](https://user-images.githubusercontent.com/23027624/51912961-cca6cb00-23d5-11e9-90cc-9773021c7def.JPG)
![2](https://user-images.githubusercontent.com/23027624/51912962-cd3f6180-23d5-11e9-86b9-1494fef6fcb9.JPG)
![immagine](https://user-images.githubusercontent.com/23027624/51969490-d3d0e600-2474-11e9-9d0b-5d69c5052187.png)

### Environment

 - Debian 6.3.0
 - Lizmap version: 3.2
 - QGis version: 2.18.25
 - Postgresql 9.6.10

rldhont:
@tethysgeco can you test PR #1117 ? thanks

rldhont:
@nboisteault ok, can you create a pull request ?

tethysgeco:
Sorry for the late reply.
We have just had time to test the improvement and we can confirm that minidock popup works properly.
Thank you very much for your work.

PR #1117, opened 2019-03-20
[Bugfix] Popups - always remained active and minidock dimensions

### LINKED ISSUE #1068: popup - used once, allways remains activated
when you do a click on the map the popup displays the contents of features and highlights them like you can see as below:

imagem

after you close the popup window and select some features using any selection method/tool (case selection tool, by circle):

imagem

the last/previous highlighted geometries by popup come back to the map:

imagem

lizmap master and previous
qgis server 2.14 e 2.18

thanks

rldhont:
Fixed #1068 popup - used once, allways remains activated

In the same time, popup in minidock dimension #1065 is calculated after content added.

### DIFF
--- a/lizmap/install/qgis/montpellier.qgs
+++ b/lizmap/install/qgis/montpellier.qgs
@@ -6375,7 +6375,7 @@ def my_form_open(dialog, layer, feature):
     <Macros>
       <pythonCode type="QString"></pythonCode>
     </Macros>
-    <WMSAddWktGeometry type="bool">false</WMSAddWktGeometry>
+    <WMSAddWktGeometry type="bool">true</WMSAddWktGeometry>
     <WCSUrl type="QString"></WCSUrl>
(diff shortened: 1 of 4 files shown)


josemvm:
hi @rldhont 
so, everything is OK, now, with popups!!!

however i think you can improve a little bit more:

when the user clicks on two or more very close features and chooses to edit one from the popup
imagem

the popup should be disabled and the feature one on edition should be marked as selected
imagem

thank you so much :-)

nboisteault:
@josemvm Do you change something in Montpellier QGIS project to display the geometries ?

nboisteault:
@josemvm forget my question, I had an issue on my side.

nboisteault:
@rldhont with lizmap_3_2 I can not display geometries when clicking on it but the selection tool does not select.
With your patch it is the contrary, I can display geometries when clicking on it but the select tool does select.

nboisteault:
@josemvm Can you open another issue for the enhancement you are suggesting please ?

PR #1120, opened 2019-03-20
Fix the type of the features property into layer objects of lizmap

laurentj:
It supposed to be an object, as the code in maps.js suggests, but some code in attributeTables.js create this property as an Array instead of an Object.

We fix the type as Object, as it is more efficient than an array, since feature id could be large numbers. 

We don't use a Map because of the support of IE10.

### DIFF
--- a/lizmap/www/js/atlas.js
+++ b/lizmap/www/js/atlas.js
@@ -95,11 +95,8 @@ var lizAtlas = function() {
             var s_field = lizAtlasConfig['sortField'];
             if ( !s_field )
                 s_field = pkey_field;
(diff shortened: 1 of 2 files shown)


laurentj:
I'm not propose this patch for 3.2, as I'm not sure it doesn't break some things or some custom modules.

rldhont:
@laurentj @mdouchin we will need to check our proposed examples scripts and https://github.com/3liz/lizmap-javascript-scripts

nboisteault:
Should this PR be made against release_3_2 and not master as suggested in @laurentj last comment ?

PR #1129, opened 2019-03-22
Remove exclamation mark from error messages. Fix #979

### LINKED ISSUE #979: Remove exclamation mark from error messages
Many error messages end with an exclamation mark.

I suggest to remove them, as it gives the impression of being assaulted. Sometimes, the error may come directly from the user himself, but many errors are triggered because of bad configuration, bad project files (that may no be created by the current user) etc...

Error messages should be more kindness :-)

nboisteault:


rldhont:
Has been manually merged.
I have also updated https://github.com/3liz/lizmap-locales and Transifex

PR #1141, opened 2019-03-27
Fix process of the alias parameter to qgisFormControl

laurentj:
qgisFormControl is expecting to receive a simpleXml object for the alias parameter, whereas it seems it receives a simple string.

Let's keep the process of the simpleXml object to not break custom modules, but check if it is a simple string to avoid warnings

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisFormControl.class.php
+++ b/lizmap/modules/lizmap/classes/qgisFormControl.class.php
@@ -166,7 +166,7 @@ class qgisFormControl
      *
      * @param string           $ref                name of the control
      * @param SimpleXMLElement $edittype           simplexml object corresponding to the QGIS edittype for this field
-     * @param object           $aliasXml           simplexml object corresponding to the QGIS alias for this field
+     * @param object|array|string $aliasXml        simplexml object corresponding to the QGIS alias for this field
      * @param object           $rendererCategories simplexml object corresponding to the QGIS categories of the renderer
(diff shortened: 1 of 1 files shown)


laurentj:
I want to propose an other patch on master, which remove this case of simpleXML object, unless you tell me that some custom module give this type of alias parameter... Are you ok for this other patch?

laurentj:
For more explanation:

qgisProject [read aliases and store them as simple strings](https://github.com/3liz/lizmap-web-client/blob/release_3_2/lizmap/modules/lizmap/classes/qgisProject.class.php#L660) :

```php
                    foreach ($fields as $field) {
                        $aliases[$field] = $field;
                        $alias = $xmlLayer->xpath("aliases/alias[@field='".$field."']");
                        if ($alias && count($alias) != 0) {
                            $alias = $alias[0];
                            $aliases[$field] = (string) $alias['name'];
                        }
                    }
```

This aliases array [is retrieving by qgisForm and qgisForm give the right alias to qgisFormControl](https://github.com/3liz/lizmap-web-client/blob/release_3_2/lizmap/modules/lizmap/classes/qgisForm.class.php#L111)

```php
            $alias = null;
            if ($aliases and array_key_exists($fieldName, $aliases)) {
                $alias = $aliases[$fieldName];
            }
            $formControl = new qgisFormControl($fieldName, $edittype, $alias, $categoriesXml, $prop);
```

So at this step,  $alias is a string (or did I miss something?).

Whereas qgisFormControl expects to have a simplexml or an object(?)...

PR #1142, opened 2019-03-27
Improve cache invalidation for qgis projects

laurentj:
If we add some properties into a deep object that is in the cache, like a new property into the layer structure, it will be more and more difficult to add tests on the check of the cache validation.

Using a version number is easier to invalidate the cache. We only had to increase this number, instead of adding more and more properties checks.


mdouchin:
Looks good to me !

PR #1143, opened 2019-03-27
Use Qgis default value to set default value on form controls

laurentj:
Default values other than raw strings or number, are ignored.

Parameters of qgisFormControl constructor are changed:
- new parameter $defaultValue
- $prop parameter is put before parameters that have a default value,
  else it is not a correct php syntax.

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisForm.class.php
+++ b/lizmap/modules/lizmap/classes/qgisForm.class.php
@@ -113,7 +113,9 @@ class qgisForm implements qgisFormControlsInterface
                 $alias = $aliases[$fieldName];
             }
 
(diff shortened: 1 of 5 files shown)


lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -167,6 +169,35 @@ public function __construct($layer, $form, $featureId, $loginFilteredOverride)
...
+            return null;
+        }
+        if (is_numeric($expression)) {
+            return $expression;
+        }
+        else if (preg_match("/^'.*'$/", $expression)) {
+            // it seems this is a simple string
+            $expression = trim($expression, "'");
+            // if there are some ' without a \, then probably we have a
+            // true expression, not a single string.
```
rldhont: An other characters for true expression is `"` which is used to decorate field name.

Issue #1132, opened 2019-03-22
Do not display atlas composition in print template list


rldhont:
### What is the bug?
Atlas composition are listed in print template selector.

### Steps to reproduce the issue

1. Go to https://demo.lizmap.com/lizmap_3_2/index.php/view/map/?repository=demo&project=cats
2. Click on `Print`
3. `Cat map` is displayed in Template selector

### Expected behavior
Atlas composition has not to be listed in print template selector, and if there is only atals composition, print tool has not to be available.

### Environment

 - Lizmap version: 3.2.2
 - QGis version: 2.18

PR #1147, opened 2019-03-28
[Bugfix] Remove atlas compositions from print template selector

rldhont:
Atlas compositions are available in the print template even if it has not been designed to be used as a simple print template

These changes remove atlas compositions from print template selector and update Montpellier project.
In Montpellier project, Atals configuration has been removed from Landscape A4 and District card composition has been added.


rldhont:
This PR will fix #1132 Do not display atlas composition in print template list

PR #1165, opened 2019-04-03
Fix attributte table - order by field type number Fix #1156

### LINKED ISSUE #1156: attribute table - order by field type number
hi

when i try to sort the table by gid, lizmap not parsing field number as number but as string, from postgreSQL tables

imagem

it's a regression...

lwc: all versions
qgis server 2.18 and 3.4

thanks

nboisteault:
With QGIS Server 3.4 DescribeFeatureType requests return 'int' and not 'integer' for integer type.

### DIFF
--- a/lizmap/www/js/attributeTable.js
+++ b/lizmap/www/js/attributeTable.js
@@ -1297,48 +1297,55 @@ var lizAttributeTable = function() {
                     // Check if we need to replace url or media by link
                     // Add function for any string cell
                     // First check if the col is number
-                    if (idx in cTypes && cTypes[idx] == 'integer')
-                        colConf['mRender'] = function( data, type, full, meta ){
-                            return parseInt(data);
-                        }
-                    else if (idx in cTypes && cTypes[idx] == 'long')
-                        colConf['mRender'] = function( data, type, full, meta ){
-                            return parseInt(data);
-                        }
-                    else if (idx in cTypes && cTypes[idx] == 'double')
-                        colConf['mRender'] = function( data, type, full, meta ){
-                            return parseFloat(data);
-                        }
-                    else
-                        colConf['mRender'] = function( data, type, full, meta ){
-                            // Translate field ( language translation OR code->label translation )
-                            var colMeta = meta.settings.aoColumns[meta.col];
-                            var colName = colMeta.mData
-                            var translation_dict = null;
-                            var tdata = data;
-                            if(data)
-                                tdata = lizMap.translateWfsFieldValues(aName, colName, data.toString(), translation_dict);
-                            if( tdata === null )
-                                tdata = data;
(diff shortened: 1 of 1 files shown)


rldhont:
It is the same in last QGIS 2.18 version

rldhont:
The Qgis server code to describe type is:
* https://github.com/qgis/QGIS/blob/release-2_18/src/server/qgswfsprojectparser.cpp#L451 for last QGIS Server 2.18.*
* https://github.com/qgis/QGIS/blob/release-3_4/src/server/services/wfs/qgswfsdescribefeaturetype.cpp#L281 for QGIS Server 3.4.* and upper

Numbers can be :
* int
* unsignedInt
* long
* unsignedLong
* integer
* decimal

lizmap/www/js/attributeTable.js
```diff
@@ -1297,7 +1297,7 @@ var lizAttributeTable = function() {
                     // Check if we need to replace url or media by link
                     // Add function for any string cell
                     // First check if the col is number
-                    if (idx in cTypes && cTypes[idx] == 'integer')
+                    if (idx in cTypes && (cTypes[idx] == 'integer' || cTypes[idx] == 'int'))
```
rldhont: Can you rebuild this by using switch. It will be easiest to cover all the numbers description.

nboisteault: Do we want 'int' and 'integer' types ? https://github.com/qgis/QGIS/blob/release-2_18/src/server/qgswfsprojectparser.cpp#L473

rldhont: Yes, the twice are possible and all other integer types.

lizmap/www/js/attributeTable.js
```diff
@@ -1297,48 +1297,55 @@ var lizAttributeTable = function() {
...
+                            case 'integer':
+                            case 'int':
+                            case 'unsignedInt':
+                            case 'long':
+                            case 'unsignedLong':
+                                colConf['mRender'] = function( data, type, full, meta ){
+                                    return parseInt(data);
+                                }
+                                break;
+                            // case 'double':
```
rldhont: Can you conserve double to avoid regression ?
Thanks

josemvm:
@rldhont and @nboisteault thanks a lot for fix it!
however, the hyperlink was lost to media columns (for some layers, and i don´t know why...)
instead of the alias field, the path appears: media/xxxxxxxxx_imagens/a_as_4477.jpg

josemvm:
@rldhont and @nboisteault i have found the problem after this commit: hyperlink was lost to media columns, for layers whose name contains blank spaces!

i can open a new issue if you prefer

rldhont:
@josemvm thanks to open a new issue

Issue #1173, opened 2019-04-04
Layer switcher's expander could sometimes does not work


rldhont:
### What is the bug?
In the layer switcher, the expander button is used to display or hide children layer or legend.

In some cases, it does not correctly work.

In this map, the _expander_ of the group **Cours d'eau** can hide but cannot display children.
https://eaurmc.lizmap.com/map/index.php/view/map/?repository=eaurmc&project=ProjetPdM2022_2027_RMC_Lizmap

In the same time, the _unfold all_ button works.

### Steps to reproduce the issue

1. Go to https://eaurmc.lizmap.com/map/index.php/view/map/?repository=eaurmc&project=ProjetPdM2022_2027_RMC_Lizmap
2. Click on the **Cours d'eau** expander to hide children
3. Click on the **Cours d'eau** expander to display children
4. **Cours d'eau** children does not display

### Expected behavior
The expander always display or hide children

### Environment

 - Lizmap version: 3.2.2
 - QGIS Server version: 2.18
 - QGIS Project version: 2.18
 - OS: All  
 - Browser: Tested on Firefox

nboisteault:
@rldhont Lizmap randomy returns "SERVICE NON DISPONIBLE!". Is this project located on our servers ?

nboisteault:
First catch : some layers (maybe all) are in hidden groups, if I disable this CSS part, behavior become normal.
`#switcher tr.liz-hidden {
    /* display: none; */
}`

rldhont:
Has been fixed by https://github.com/3liz/lizmap-web-client/pull/1174 thanks @nboisteault

PR #1174, opened 2019-04-05
Layerswitcher : Hide also children when parent group is hidden

nboisteault:


nboisteault:
Fix #1173 but some changes might be needed on the ProjetPdM2022_2027_RMC_Lizmap project.

Issue #1051, opened 2019-01-10
Bad display of top menu bar on mobile


laurentj:
### What is the bug?

With low width screen, the top menu bar display icons and item text, except in the map view mode.

### Steps to reproduce the issue

1. With Firefox > Developpement web > vue adpatative, select a little width for the page
2. go the project page or an administration page
3. see that item text are still shown, and overflow the menu bar.

![lizmap_mobile_menu_overflow](https://user-images.githubusercontent.com/336034/50962107-cd7cc900-14c9-11e9-856d-a2b639d34659.png)

### Expected behavior

We should see only icons, so it doesn't overflow the menu bar.

![lizmap_mobile_menu_ok](https://user-images.githubusercontent.com/336034/50962458-9529ba80-14ca-11e9-90ab-6adbfa563eaa.png)

### Environment

 - Lizmap version:  3.2
 - OS : kubuntu 
 - Browser : Firefox

nboisteault:
@laurentj do you think logged name should also be hidden ?

laurentj:
@nboisteault yes

laurentj:
we could open a new issue, to hide the search field on tiny screen. We could have instead a button in the menu, that open a dialog box showing the search field.

rldhont:
@laurentj agree with you.

nboisteault:
Working on search bar.

PR #1175, opened 2019-04-05
Better display of top menu bar on mobile

nboisteault:
Fix (partially ?) #1051


rldhont:
@thanks for this first fix

Issue #1044, opened 2019-01-08
Add a link to the dashboard at the header barre


laurentj:
### Is your feature request related to a problem? Please describe

To access to pages of the administration, the only way is to go into the user popup menu, click on "my account", then navigate into the sidebar menu.

It does not make sens for  a user to go to his account to access to some features not related to its account.

### Describe the solution you'd like

It would be better if there was an "admin dashboard" item at the top of the page. This item is a link to the admin dashboard page, which is a better entry point to administration features than the personal account of the user.

It could looked like that :

![lizmap_dashboard_item_admin](https://user-images.githubusercontent.com/336034/50828288-397ff580-1341-11e9-93e9-a6156b171101.png)
![lizmap_dashboard_item_map](https://user-images.githubusercontent.com/336034/50828289-397ff580-1341-11e9-919b-3b4a1f7201ba.png)

(we could change the name of the link if you prefer an other name)

We could add it into the 3.2 version, it's just a matter of few CSS and few html tags.
(I have a patch).

josemvm:
@laurentj i propose only the icon for the mobile devices mode...
thanks

laurentj:
Was done into 3.3.0

PR #1180, opened 2019-04-06
UI: add an item 'administration' on the top menu

laurentj:
Refs #1044


laurentj:
In the map page:
image

In an administration page:
image

lizmap/modules/lizmap/templates/user_menu.tpl
```diff
@@ -1,5 +1,9 @@
 {hook 'LizTopMenuHtmlItems'}
+
 {ifuserconnected}
+    <li class="dashboard-item"><a href="{jurl 'master_admin~default:index'}">
+            <span class="icon dashboard-icon"></span> <span class="text">{@view~default.header.menuitem.admin.label@}</span></a>
```
rldhont: Can you add phone-hidden class to the span text as connect text https://github.com/3liz/lizmap-web-client/pull/1180/files#diff-c6f3c6cf7ce683bf12c8bca40b3fab82R29
```
<span class="text hidden-phone">
```

nboisteault: Same comment as @rldhont. The rest is ok for me.

laurentj: added :-)

Issue #521, opened 2017-02-28
edition form - date widget display


josemvm:
hi,

the native language of my OS/win is PT, then the default language of my firefox web browser is also PT

for the native language (PT), LWC doesn't display correctly the date widget:
![pt](https://cloud.githubusercontent.com/assets/10053874/23413688/55b9a2a0-fdd1-11e6-885f-bc0a28f173a4.jpg)

but if i chose a different language for LWC, like lang=en_US or lang=fr_FR, the date widget appears correctly:
![en](https://cloud.githubusercontent.com/assets/10053874/23412568/d04810f0-fdcd-11e6-856d-613407de7d85.jpg) ![fr](https://cloud.githubusercontent.com/assets/10053874/23413225/ffff2994-fdcf-11e6-94ae-957f84c89346.jpg)

i don't know if it's a Lizmap issue or QGIS server issue...

qgis server and desktop 2.14.12
version 3.0.11 and master
firefox 51.01

thanks

PR #1182, opened 2019-04-08
[Bugfix] Add local JS for date picker form widget

rldhont:
Locale files for date picker widget was not added to the html map template, so date picker in edition form was not localized.

This PR completed the commit 1784a4e52114043ec9fb27af5fb6a622ab54a0e5 and the fix of
edition form - date widget display #521

### DIFF
--- a/lizmap/modules/view/controllers/lizMap.classic.php
+++ b/lizmap/modules/view/controllers/lizMap.classic.php
@@ -145,8 +145,21 @@ class lizMapCtrl extends jController
         $www = $confUrlEngine['jelixWWWPath'];
         $rep->addJSLink($www.'jquery/include/jquery.include.js');
         $rep->addJSLink($www.'js/jforms_jquery.js');
-        $rep->addJSLink($www.'js/jforms/datepickers/default/init.js');
-        $rep->addJSLink($www.'js/jforms/datepickers/default/ui.en.js');
+
+        // Add datepickers jForms js
+        $confDate = &jApp::config()->datepickers;
+        $rep->addJSLink($confDate['default']);
+        if (isset($confDate['default.js'])) {
+            $js = $confDate['default.js'];
+            foreach($js as $file) {
+                $file = str_replace('$lang', jLocale::getCurrentLang(), $file);
+                if (strpos($file, 'jquery.ui.datepicker-en.js') !== false) {
+                    continue;
+                }
+                $rep->addJSLink($file);
+            }
+        }
+        // Add other jForms js
         $rep->addJSLink($bp.'js/fileUpload/jquery.fileupload.js');
         $rep->addJSLink($bp.'js/bootstrapErrorDecoratorHtml.js');
 


rldhont:
@josemvm can you test it ?

josemvm:
@rldhont the calendar doesn´t appears, you can verify with montpellier project

lizmap/modules/view/controllers/lizMap.classic.php
```diff
@@ -145,8 +145,21 @@ public function index()
...
-        $rep->addJSLink($www.'js/jforms/datepickers/default/init.js');
-        $rep->addJSLink($www.'js/jforms/datepickers/default/ui.en.js');
+
+        // Add datepickers jForms js
+        $confDate = &jApp::config()->datepickers;
+        $rep->addJSLink($confDate['default']);
+        if (isset($confDate['default.js'])) {
+            $js = $confDate['default.js'];
+            foreach($js as $file) {
+                $file = str_replace('$lang', jLocale::getCurrentLang(), $file);
```
rldhont: @laurentj I have found an issue, If the lang is set to `pt_PT`, getCurrentLang() returns `en` and not `pt`. Do you knw where it comes from ?

rldhont: an error in my branches, no issue.

josemvm:
the widget only appears for `lang=en_US`

josemvm:
hi @rldhont the calendar is ok

but the date format by country is the same for France and Portugal dd-MM-yyyy

imagem

`lang=fr_FR`
default date format = **dd-MM-yyyy** - OK

imagem
`lang=pt_PT`
default date format = **MM-dd-yyyy** - WRONG

but the date format by country is the same for France and Portugal dd-MM-yyyy

thanks

mdouchin:
I think there is a misunderstanding here. @josemvm would like to have the order of the comboboxes depend on the locale. I am not sure it is really possible or easy to implement. Any idea @laurentj ?

rldhont:
@josemvm we need to check the Jquery local

rldhont:
@josemvm I have found why only french date-time widget have the right order

The issue comes from that the file `format.UTF-8.properties` is not localize:
* en_US https://github.com/3liz/lizmap-web-client/blob/master/lib/jelix/core-modules/jelix/locales/en_US/format.UTF-8.properties
* fr_FR https://github.com/3liz/lizmap-web-client/blob/master/lib/jelix/core-modules/jelix/locales/fr_FR/format.UTF-8.properties
* pt_PT https://github.com/3liz/lizmap-web-client/blob/master/lizmap/var/overloads/jelix/locales/pt_PT/format.UTF-8.properties
This file has not been pushed to https://github.com/3Liz/lizmap-locales and transifex because it is a technical file.

@josemvm @laurentj do I add this file to https://github.com/3Liz/lizmap-locales and transifex, or can we find a way to fix it from other project ?

rldhont:
@josemvm we will fix it by using https://www.php.net/intl

josemvm:
ok @rldhont 
thanks again

laurentj:
You can merge this PR. I created a new issue for this specific problem about the order of fields in the date picker: issue #1188

PR #1186, opened 2019-04-08
Set QGIS Server default version to > 3.0

nboisteault:


rldhont:
Can you check and class lizmapServices.class.php :
```
    // QGIS Server version
    public $qgisServerVersion = '3.0';
```

PR #1189, opened 2019-04-09
Vagrant - Propose increased software versions and fix PostgreSQL >= 10 user creation

mdouchin:


nboisteault:
I have tested bleeding edge and it worked for me. Thank you @mdouchin  !

PR #1195, opened 2019-04-11
Update login.form.tpl

rmarzocchi84:
Login form responsive

### DIFF
--- a/lib/jelix-modules/jauth/templates/login.form.tpl
+++ b/lib/jelix-modules/jauth/templates/login.form.tpl
@@ -6,7 +6,7 @@
 {if ! $isLogged}
 
 <form action="{formurl 'jauth~login:in'}" method="post" id="loginForm">
-      <fieldset>
+      <f<div class="container">
       <table>
        <tr>
            <th><label for="login">{@jauth~auth.login@}</label></th>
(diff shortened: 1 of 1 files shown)


lib/jelix-modules/jauth/templates/login.form.tpl
```diff
@@ -6,7 +6,7 @@
 {if ! $isLogged}
 
 <form action="{formurl 'jauth~login:in'}" method="post" id="loginForm">
-      <fieldset>
+      <f<div class="container">
```
rldhont: Can you remove `<f`

rldhont:
@rmarzocchi84 the file to modify is not the one you updated.
The right file is `lizmap/var/themes/default/jcommunity/login.tpl`
Can you do the update ?

rldhont:
I have tested the changes to `lizmap/var/themes/default/jcommunity/login.tpl` and it is not the right way.

rldhont:
@rmarzocchi84 I have found how to fix responsive login page.

rmarzocchi84:
Sorry for my not correct suggestion. Unfortunately I never used this version of bootstrap (I have experience only with v.3 and v.4)

rldhont:
@rmarzocchi84 thanks to point the issue

PR #1250, opened 2019-05-23
Hide the "data" button when there is no displayed attribute tables

laurentj:
The data panel and its corresponding button should not be shown when
there is no attribute table set with the "hideLayer=False" into the
map configuration.


rldhont:
Thanks

PR #1260, opened 2019-06-06
[Feature] Cached GetCapabilities

rldhont:
The GetCapabilities responses are not yet stored in cached.
To get faster lizmap GetCapabilities request, the result has been stored in cahce with session_id and the user login.


laurentj:
ok for me :-)

PR #1302, opened 2019-07-26
Checkbox always tick. Fix #1261

### LINKED ISSUE #1261: Checkbox always unticked when updating data
Hi,
I set a boolean field in a table. When the users tick the checkbox, it is shown as True in the data tool (so far so good).
But when a user wants to edit the data with True value, the checkbox from the form is not checked anymore, and stays as False when the edited row is saved.
I though it was because of the default value set in Postgresql and QGIS (set as False), but I observed the same behaviour when I dropped this default constraint (in both Postgres and QGIS).
I would like the value of this boolean field (true for instance) stays as true when editing that row.
Any idea please?
Thanks

nboisteault:


rldhont:
Good catch ! No needs to update the method `setFormDataFromDefault` ?

PR #1307, opened 2019-08-06
Add LizmapZoomElement

nboisteault:


assets/src/modules/LizmapMap.js
```diff
@@ -38,11 +40,29 @@ export default class LizmapMap {
...
     }
 
-    get zoom () {
+    get zoom() {
         return this._zoom;
     }
 
-    get baseLayerGroup () {
+    set zoom(zoom) {
+        this._zoom = zoom;
```
laurentj: To avoid potential infinite loop, you should add  `if(this._zoom == zoom) { return;}` before the assignation.

assets/src/modules/LizmapMap.js
```diff
@@ -38,11 +40,29 @@ export default class LizmapMap {
```
laurentj: No initial set of _zoom from the configuration?

assets/src/components/LizmapZoomElement.js
```diff
@@ -0,0 +1,48 @@
...
+                    z-index: 1;
+                }
+                button{
+                    display:block;
+                    width: 30px;
+                    height: 30px;
+                }
+            </style>`;
+
+            const zoomin = document.createElement('button');
```
laurentj: bad indentation on this line and next lines

assets/src/modules/LizmapMap.js
nboisteault: I have some improvement to make about configuration on the whole code. I'll make it in another commit.

PR #1309, opened 2019-08-07
Set min and max zoom based on configuration. Create and add zoom slider component

nboisteault:


laurentj:
The code is good, but I think we could optimize it.

To avoid too many events in the long terme, and since the setting of minZoom and maxZoom are always set together, i think we should have only one event to propagate the change of min/maxZoom.

So I suggest :

- on LizmapMap: to replace properties setter of minZoom maxZoom by a setMinMaxZoom() method
- to replace events map-min-zoom-set and map-max-zoom-set by a single event max-min-mas-zoom-set

What do you think about it?

laurentj:
Ok for me

PR #1311, opened 2019-08-07
Translation availability - pt_BR

alexanderfschley:
Hello. Initially I want to congratulate on the lizmap project. I have finished translating the Lizmap Web Client into Brazilian Portuguese on Transifex to expand its use in Brazil and would like to know what is required for the translation to be available in the new versions? Because I did the translation some time ago, some versions have been released after that, but the translation is not yet available.

Thank you very much in advance.


nboisteault:
@alexanderfschley thank you very much for your work. We'll integrate your translations soon.

laurentj:
Sorry, we don't accept pull request for translations in this repository (see CONTRIBUTING.md file). You should work only with Transifex. We then pull your translations from transifex into the https://github.com/3liz/lizmap-locales/ and update lizmap-web-client with our tools.

However I found that we has an issue in the use of the transifex client, to retrieve translations, and we didn't get all latest translations. I fixed that and translations are now up to date in lizmap-web-client

PR #1341, opened 2019-10-02
[Regression] Display values from Array field in lizmap edit form.

### LINKED ISSUE #1340: [Form] Display values from Array field in lizmap edit form
### What is the bug?

When I want edit layer in Lizmap I have a problem with array type fields. In form this fields are represented by checkboxes. When I  am on edit form, the fields values are not checked.

1. Need an array field in database: integer[]
2. In QGIS put the field in relational value which references another layer in properties ->attribute form
3. make the layer editable on Lizmap
4. In Lizmap select a feature and click edit layer
5. see none of the checkboxes are checked

### Expected behavior

The value get from database is {102,104,107,108}, jelix does not recognized it as a list of value.

### Screenshots and logs

Select feature
select

Field with relationnal value not empty
field

Edit feature: none are checked
check

### Environment

 - Lizmap version: 3.3
 - QGIS Server version: 3.4
 - QGIS Project version: 3.4
 - OS (Windows, Linux, MacOS, Android…): Linux
 - Browser (Firefox, Chrome…): Chrome and Firefox
 - Lizmap admin information (users, logs…) stored into Sqlite (default)? PostgreSQL? MySQL?

nboisteault:
Fix #1340

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisForm.class.php
+++ b/lizmap/modules/lizmap/classes/qgisForm.class.php
@@ -425,15 +425,11 @@ class qgisForm implements qgisFormControlsInterface
                 $form->getControl($ref)->setDataFromDao($value, 'boolean');
             }
             // ValueRelation can be an array (i.e. {1,2,3})
-            if ($this->formControls[$ref]->isValueRelation()) {
-                if ($value[0] == '{') {
-                    $arrayValue = explode(',', trim($value, '{}'));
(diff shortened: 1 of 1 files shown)


mdouchin:
ok for me, even if IMHO the elseif are error prone if the conditions includes specific things. Here it is ok because the last else if not far and will take care of the "normal" valuerelation with simple values (not array).

Issue #1301, opened 2019-07-26
Split Features ?


Mavialle:
Hello,

In Lizmap, is it possible to split a feature in two part and create two objects from one ? As does the split features tool in QGIS ?

Thank you for your help and advices.

rldhont:
Not yet

laurentj:
The feature has been implemented.

geographit:
Hi ,

I have tried integrating the files from #1336 and #1351 in Lizmap 3.3 .
I can see the Split functionality when editing, however when I split a feature, one part gets deleted.

For example if I split a polygon into 2 pieces I can see in Postgresql log that the biggest part of the split feature gets updated while the other one is an INSERT statement with the same ID which gets rejected. Hence the smallest piece become "deleted".

```
LOG:  statement:  INSERT INTO "webapp"."00_test_polygon" ("id", "geom" ) VALUES (1, ST_GeomFromText('POLYGON((12942892.827410802 -3766037.86237783,12943128.0256651 -3767249.13338755,12942577.357415842 -3767435.3888248047,12942498.842978436 -3765921.8049481474,12942892.827410802 -3766037.86237783))', 3857) )  RETURNING "id";
ERROR:  duplicate key value violates unique constraint "00_test_polygon_pkey"
DETAIL:  Key (id)=(1) already exists.
STATEMENT:   INSERT INTO "webapp"."00_test_polygon" ("id", "geom" ) VALUES (1, ST_GeomFromText('POLYGON((12942892.827410802 -3766037.86237783,12943128.0256651 -3767249.13338755,12942577.357415842 -3767435.3888248047,12942498.842978436 -3765921.8049481474,12942892.827410802 -3766037.86237783))', 3857) )  RETURNING "id";
```

Is there anything I might have missed?

Thanks, this is a great feature!

dtk78:
Hi laurentj ,

On which version that this feature is implemented ?
It is a great feature to be able to split one object into 2 pieces
I got an error while doing split even in the last version 3.4.0pre
```
2020-05-05 20:31:33.146 UTC [1472] ERROR:  duplicate key value violates unique constraint "00_test_polygon_pkey"
2020-05-05 20:31:33.146 UTC [1472] DETAIL:  Key (id)=(11) already exists.
[...]

Issue #687, opened 2017-10-13
delete photos from features in case of related tables


josemvm:
hi

i noticed that for related tables the LWC doesn't delete the photo files (upload folder) for both tables (main and related), when the user deletes features from main table, 

if the user delete a related feature (related table), there's no problem, the photo file is deleted

thanks

rldhont:
@josemvm you'd like to do cascading deletion, isn't it ?

josemvm:
@rldhont just to remind you

josemvm:
i am talking about it because this problem is really embarrassing. in the case of deleting a feature, the user must search for the photo and manually delete it

thanks again

josemvm:
@nboisteault please can you do something to solve this old bug?

PR #1351, opened 2019-10-11
Improve the edition of parent feature

laurentj:
When we launch the edition of a child feature during the edition of a feature, the return to the edition of the parent is not well done in the case when the split tool was used (issue #1301).

To fix this issue, parts of the code of the editor has been refactorized, and will allow to improve it easier.


josemvm:
hi @laurentj 
can this pull also be a solution for this issue https://github.com/3liz/lizmap-web-client/issues/687 ?

laurentj:
@josemvm no, i don't think it fixes this issue.

PR #1353, opened 2019-10-14
Add a way to filter features in updateAtlasData()

nboisteault:


nboisteault:
@rldhont @mdouchin  I finally dislike the fact another request is sent when filtering.
@mdouchin suggested to use `lizMap.getLayerFeature()`  but it had to be used feature by feature, we could modify it to accept an array but I prefer to not modify this part of the code.
I found another way to develop the filter on atlas.
I would like to put this PR in draft mode but it is not possible so I close it.

Issue #1347, opened 2019-10-10
Custom HTML content at top of projects page


nboisteault:
### Is your feature request related to a problem? Please describe
We need a way to let admin create some HTML content at top of projects page.

### Describe the solution you'd like
- [x] A link in the admin menu to a configuration page (Contenu de la page d'accueil / Landing page content)
- [x] The configuration page has a wysiwyg widget to let the user build the HTML content.
- [x] HTML content is saved in `lizmapConfig.ini.php` in the `[landing_page_content]` section

- [x] HTML content is shown at top of projects page

nboisteault:
@mdouchin @laurentj le contenu du wysiwyg, c'est stockable dans un fichier ini ou ça devrait plutôt aller en base sqlite ou postgres ?

mdouchin:
IMHO, ini file is better

laurentj:
No, ini file is bad to store HTML content. You should store it else where, in a file for example. Or in a template ? 
The problem of the database is we don't have any existing table to store this html part, and we won't create a table just for this item.

nboisteault:
@laurentj If I store HTML in a file, is it ok to put it in `lizmap/var/config/` ?

PR #1358, opened 2019-10-22
Custom html projects page

nboisteault:
Fix https://github.com/3liz/lizmap-web-client/issues/1347

### DIFF
--- a/lizmap/modules/admin/controllers/landing_page_content.classic.php
+++ b/lizmap/modules/admin/controllers/landing_page_content.classic.php
@@ -1,0 +1,107 @@
+<?php
+/**
+ * Lizmap administration : landing page content.
+ *
+ * @author    3liz
+ * @copyright 2016 3liz
+ *
+ * @see      http://3liz.com
+ *
+ * @license Mozilla Public License : http://www.mozilla.org/MPL/
+ */
+class landing_page_contentCtrl extends jController
+{
+    // Configure access via jacl2 rights management
+    public $pluginParams = array(
+        '*' => array('jacl2.right' => 'lizmap.admin.access'),
+    );
+
+    /**
+     * Display a wysiwyg editor.
+     */
+    public function index()
+    {
+        $rep = $this->getResponse('html');
+
+        // Create the form
+        $form = jForms::create('admin~landing_page_content');
+
+        // Get HTML content
+        $HTMLContent = jFile::read(jApp::varPath('lizmap-theme-config/landing_page_content.html'));
+        if($HTMLContent){
+            $form->setData( "HTMLContent", $HTMLContent );
+        }
+
+        $tpl = new jTpl();
+
+        $tpl->assign('form', $form);
+        $rep->body->assign('MAIN', $tpl->fetch('landing_page_content'));
+        $rep->body->assign('selectedMenuItem', 'lizmap_landing_page_content');
+
+        return $rep;
+    }
+
+    /**
+     * Save wysiwyg editor content in ini file.
+     */
+    public function save()
+    {
+        $form = jForms::get('admin~landing_page_content');
+
+        // token
+        $token = $this->param('__JFORMS_TOKEN__');
+        if (!$token) {
+            // redirection vers la page d'erreur
+            $rep = $this->getResponse('redirect');
+            $rep->action = 'landing_page_content:index';
+
+            return $rep;
+        }
+
+        // If the form is not defined, redirection
+        if (!$form) {
+            $rep = $this->getResponse('redirect');
+            $rep->action = 'landing_page_content:index';
+
+            return $rep;
+        }
+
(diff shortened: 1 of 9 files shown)


rldhont:
If it is a new feature it has to be merged in master ?

rldhont:
@laurentj can we create a branch release_3_4 ?

lizmap/modules/admin/controllers/landing_page_content.classic.php
```diff
@@ -0,0 +1,107 @@
...
+     * Display a wysiwyg editor.
+     */
+    public function index()
+    {
+        $rep = $this->getResponse('html');
+
+        // Get HTML content from ini file
+        $lizmapConfigIniFile = jApp::configPath('lizmapConfig.ini.php');
+        $iniModifier = new jIniFileModifier ($lizmapConfigIniFile);
+
```
laurentj: no, you must not use jIniFileModifier to read an ini file. It takes too much resources.

And you should not read lizmapConfig.ini.php directly, use the class LizmapService

lizmap/modules/admin/controllers/landing_page_content.classic.php
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        // Remove line breaks, spaces between > and < then transform with htmlentities to get a single string
+        $HTMLContent = htmlentities(preg_replace('/>\s+</','><',str_replace(array("\r", "\n"), '', $form->getData('HTMLContent'))));
+
+        // Save the data
+        $lizmapConfigIniFile = jApp::configPath('lizmapConfig.ini.php');
+        $iniModifier = new jIniFileModifier ($lizmapConfigIniFile);
+
+        $iniModifier->setValue("html", $HTMLContent, "landing_page_content");
```
laurentj: I'm not sure lizmapConfig.ini.php is the right place to store some HTML content...

lizmap/www/js/filter.js
```diff
@@ -538,7 +538,7 @@ var lizLayerFilterTool = function() {
                         var cval = clist[i];
                         filter+= sep + '"' + field + '"' + " " + lk + " '%" + cval + "%' ";
                         // if postgresql use ILIKE instead for WMS filtered requests
-                        sep = ' OR ';
+                        sep = ' AND ';
```
laurentj: Is this change really part of the new feature?

.gitlab-ci.yml
```diff
@@ -69,6 +69,7 @@ deploy_stable:
   script:
     - make deploy_download_stable
     - make saas_release
+    - make trigger_ci
```
laurentj: Is this change really part of the new feature? Please rebase your branch

nboisteault:
> @laurentj can we create a branch release_3_4 ?

@laurentj Can we ?

lizmap/www/js/filter.js
nboisteault: It is not. I changed branch release_3_3 to master.

nboisteault:
@laurentj  @mdouchin got an issue : 
`Impossible d'écrire le fichier /var/www/LIZMAP/release_3_3/lizmap/modules/view/templates/landing_page_content.tpl, assurez-vous que le répertoire /var/www/LIZMAP/release_3_3/lizmap/modules/view/templates autorise l'écriture	/var/www/LIZMAP/release_3_3/lib/jelix/utils/jFile.class.php`

He suggests to write template in `/lizmap/var/lizmap-theme-config/`. Is it ok for you ?

mdouchin:
I propose to use a file `landing_page_content.tpl` in `/lizmap/var/lizmap-theme-config/` and use virtual templates 

```php
$templatePath = jApp::varPath('lizmap-theme-config/landing_page_content.tpl');
$templateContent = jFile::read($templatePath);
```
Then use virtual template like in https://docs.jelix.org/fr/manuel-1.6/classes-utiles/japp#autres-methodes-de-chemins

nboisteault:
@mdouchin There is no 'source' option in CKEditor 5 for [some good reasons](https://stackoverflow.com/a/47775840/2000654). Do we agree with this answer or I try another wysiwyg widget ?

mdouchin:
Another option would be to let the user choose one of this solution

* to create and edit HTML in the CKEditor -> for normal users
* to upload an HTML file already containing the source code, which will let power users do what they want with no editor limitation

You cannot use both at the same time

nboisteault:
@mdouchin where do power user save his HTML file in second case ?

nboisteault:
@mdouchin @laurentj I have rebased commits in a single one to have a more clear history. User can use wysiwyg or edit html file created to add an iframe for instance. In this case he should not use wysiwyg after as he will lose iframe. It has to be documented to warn of this.

nboisteault:
@rldhont @laurentj  I think you can merge

PR #1382, opened 2019-11-11
Simple filter by tags/keywords

nboisteault:


nboisteault:
@mdouchin 
> * Add a tooltip on the search input to tell the user "Please start typing to filter your projects. You can use the left button to toggle the search mode, among **maps** keywords or title"

"maps" not  "projects" ?

nboisteault:
> * I think the search input should be centered, not on the right side of the header.

I think it is better to let it at right because it overlaps title on small width screen.

image

mdouchin:
> "maps" not "projects" ?
> Also the sentence is too long for a tooltip. "Type to filter your projects" would be enough IMO.

The end user has sometime no idea of what a project is, or even that QGIS is used in the background to create maps. For the end users, map is the correct word IMHO

> I think it is better to let it at right because it overlaps title on small width screen

Could we put it right centered for small screens (via @media queries in the media.css file or via JS) and put it centered in normal screens ?

nboisteault:
@mdouchin I forgot to handle accent. Working on it.

ygorigor:
@nboisteault when there is no projects in list and no user logged in the header looks like this 
image

nboisteault:
@ygorigor Fixed by https://github.com/3liz/lizmap-web-client/commit/2b8a4e336ec042a555d1a7396d471b3ba178953e
Thank you early tester ;)

PR #1393, opened 2019-11-18
Ckeditor image upload

### LINKED ISSUE #1379: Admin CKEditor for landing page - Allow embedding images

At present, the CKEditor Wysiwig tool does not allow to insert images. This is a key feature for users.

CKEditor 5 provides a package to do so:
https://ckeditor.com/docs/ckeditor5/latest/features/image.html

We should allow the user to upload an image, then place it and resize it if needed.

nboisteault:
Fix #1379


laurentj:
Sorry, i'm not agree with this PR and with the inclusion of 2154 files into our repository.

CkFinder includes a vendor directory, generated probably with Composer. We should then install CkFinder with Composer and not include the source code. And it seems it includes many packages we don't need (like packages about AWS  and other connectors...)

Don't commit change into your actual branch. Drop it completely else our history will contain all of those files.

lizmap/modules/admin/controllers/landing_page_content.classic.php
```diff
@@ -23,6 +23,10 @@ public function index()
     {
         $rep = $this->getResponse('html');
 
+        // Authorize ckfinder image upload
+        session_start();
```
laurentj: session_start() is already executed by jelix. Don't include it here

laurentj:
With all those new files, I cannot review what you changed into lizmap...

PR #1395, opened 2019-11-19
New commands to migrate lizmap data from a sqlite database to pgsql/mysql

laurentj:
New commands allow to migrate all users/acl tables from sqlite to pgsql, as well as to migrate log table.

```
php lizmap/scripts/script.php lizmap~database:migratelog
php lizmap/scripts/script.php lizmap~database:migrateusers
```

`migrateusers` is experimental. 
`migratelog` has been well tested, and will be needed on our infrastructure.

### DIFF
--- a/lizmap/modules/lizmap/controllers/config.cmdline.php
+++ b/lizmap/modules/lizmap/controllers/config.cmdline.php
@@ -1,0 +1,120 @@
+<?php
+/**
+ * @copyright 2019 3liz
+ *
+ * @see      http://3liz.com
+ *
+ * @license   MPL-2.0
+ */
+
+
+class configCtrl extends jControllerCmdLine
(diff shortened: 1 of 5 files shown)


mdouchin:
Seems all good by reading code !

rldhont:
log and geobookmark have no `%%PREFIX%%` in the pgsql sql install files.

Do we have to update them for fresh install ?

laurentj:
@rldhont we could add it, but it is for an other ticket/PR. `%%PREFIX%%` is replaced by a prefix indicated into the db profile. It can be useful if the user want to use the same database for several lizmap. But I'm not sure this is a needed feature in our case.

rldhont:
`migratelog` could be used on avignon

PR #1400, opened 2019-11-22
[Bugfix] Quote schema for search_path in dbParams for GetFeature

rldhont:
The Layer schema has to be quote in the search_path parameter of the db connection to avoid problem with `-`, upper character, etc


rldhont:
@laurentj I don't know if search_path can have `"`

rldhont:
@mdouchin The bug is critical, some data cannot be used in lizmap if the search path is not correct.

mdouchin:
in PostgreSQL, you can have double-quotes around schemas 

```sql
SET search_path TO "MYSCHEMA",public;
```
wil work

PR #1401, opened 2019-11-22
[Bugfix] Support parsing GeoJSON of geometry with Z and/or M dimension

rldhont:
OpenLayers GeoJSON format has an attribute to read 3-4D points : `ignoreExtraDims`

The commit adds `ignoreExtraDims: true` to all `new OpenLayers.Format.GeoJSON`


rldhont:
link to #1399

PR #1410, opened 2019-12-01
[Bugfix] Add HTTP code in service controller responses

rldhont:
Quick fix in HTTP code use

### DIFF
--- a/lizmap/modules/lizmap/controllers/service.classic.php
+++ b/lizmap/modules/lizmap/controllers/service.classic.php
@@ -1384,7 +1384,7 @@ class serviceCtrl extends jController
         // Get remote data
         list($data, $mime, $code) = lizmapProxy::getRemoteData($querystring);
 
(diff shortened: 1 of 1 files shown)


rldhont:
No, the issue is with QGIS 3 that returns right codes and not only 200.

lizmap/modules/lizmap/controllers/service.classic.php
```diff
@@ -1379,7 +1384,7 @@ public function DescribeFeatureType()
         // Get remote data
         list($data, $mime, $code) = lizmapProxy::getRemoteData($querystring);
 
-        if ($returnJson) {
+        if ($code/100 < 4 && $returnJson) {
```
dmarteau: Please simplify the expression: 
```suggestion
        if ($code < 400 && $returnJson) {
```

Or maybe I'm missing something here....

PR #1411, opened 2019-12-02
Add X-Lizmap-User and X-Lizmap-User-Groups HTTP headers

rldhont:


rldhont:
It to use it in QGIS Server plugins to do access control.

dmarteau:
@nboisteault more precisely it is actually used in py-qgis-wps [policy](https://github.com/3liz/pyqgisservercontrib-lizmap-access-policy) extension and there is plan for managing fine grained accesses in py-qgis-server.

PR #1419, opened 2019-12-06
Fix error in qgisVectorLayer when geometrytype contain parenthesis

laurentj:
I have many errors like this in some logs of a lizmap instance:

```
preg_match(): Compilation failed: unmatched parentheses at offset 5 
modules/lizmap/classes/qgisVectorLayer.class.php   453
```

It seems the geometryType may contains some parenthesis, and since these characters are reserved into a regexp, they should be quoted.

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisVectorLayer.class.php
+++ b/lizmap/modules/lizmap/classes/qgisVectorLayer.class.php
@@ -450,8 +450,8 @@ class qgisVectorLayer extends qgisMapLayer
         // test type
         $rs = $cnx->query('SELECT GeometryType('.$nvalue.') as geomtype');
         $rs = $rs->fetch();
-        if (!preg_match('/'.$dbFieldsInfo->geometryType.'/', strtolower($rs->geomtype))) {
-            if (preg_match('/'.str_replace('multi', '', $dbFieldsInfo->geometryType).'/', strtolower($rs->geomtype))) {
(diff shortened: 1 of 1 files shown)


laurentj:
However, I guess that the geometryType property is malformed. 

At this line, we have this code:

```
//A
                if ($this->provider == 'postgres' and $dbInfo->geometryType == 'geometry') {
                    $tablename = '"'.$dtParams->schema.'"."'.$dtParams->tablename.'"';
                    $sql = 'SELECT format_type(atttypid,atttypmod) AS type';
                    $sql .= ' FROM pg_attribute';
                    $sql .= ' WHERE attname = '.$cnx->quote($dbInfo->geometryColumn);
                    $sql .= ' AND attrelid = '.$cnx->quote($tablename).'::regclass';
                    $res = $cnx->query($sql);
                    $res = $res->fetch();
                    if ($res && strpos($res->type, '(') !== false) {
                        // It returns something like "geometry(PointZ,32620)".
//B
                        $dbInfo->geometryType = explode(',', explode('(', strtolower($res->type))[1])[0];
                    }
                }
```

Maybe the test A is false, because geometryType is not equal to `geometry` and may content some parenthesis.

Or maybe at line B, `$res->type` may equal to an expression having only one value between parenthesis, e.g. `geometry(foo)`, and then geometryType will be `foo)`.

@mdouchin @rldhont what do you think about it?

rldhont:
@laurentj I think you done right
@mdouchin can you confirm ?

rldhont:
Thanks @laurentj

PR #1420, opened 2019-12-06
Add div to ease override of geobookmark text

nboisteault:
It will be easier to override texts with theme for example. I need it for Forland.


rldhont:
@nboisteault merge as you want

PR #1421, opened 2019-12-09
Add missing dollars for some variables

nboisteault:
Last proj4php version should be installed with composer instead. All those mistakes were visible thanks to PHP intelephense in VS Code.


rldhont:
@nboisteault merge as you want

PR #1423, opened 2019-12-11
QGIS themes in Lizmap Web Client

nboisteault:


nboisteault:
@rldhont @mdouchin ce PR crée un notice [ici](https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/classes/lizmapProject.class.php#L204) dans les logs. Une idée ?

`notice	2019-12-11 11:29:38	[8]	Undefined index: themes`

rldhont:
@nboisteault I think it's when the themes has not already saved.

Probably needs to test `array_key_exists` before `$this->{$prop} = $data[$prop];`

nboisteault:
@rldhont what surprises me is that there is no notice for others prop

nboisteault:
@rldhont @mdouchin I need to fix [this issue](https://github.com/3liz/lizmap-web-client/issues/1425) first.

nboisteault:
@rldhont @mdouchin Main functionalities are developed so I think PR can be merged if no bugs are found. I'll develop later the fact that a theme is unselected when the treeview state does not correspond.

Gustry:
Nice feature!

Do you plan to update the documentation too about this new feature? (at least a ticket in the documentation so we don't forget) Thanks.

nboisteault:
@Gustry Thanks ! https://github.com/3liz/lizmap-documentation/issues/125

Issue #1380, opened 2019-11-08
Jelix cache error


ygorigor:
Hello.
My setup has a lot of errors regarding cache, ex:

`2019-11-08 12:03:24	::1	error	2019-11-08 12:03:24	::1	[901]	 error key C:\webserver\projects\acoperire/acoperire.qgs, use only alphanumeric characters and the character '_' and maximum length 255	C:\webserver\lizmap\lib\jelix\utils\jCache.class.php	444
	/index.php/view/
array (   'module' => 'view',   'action' => 'default:index', )

	0	jCache::_checkKey()	C:\webserver\lizmap\lib\jelix\utils\jCache.class.php : 109
	1	jCache::get()	C:\webserver\lizmap\lizmap\modules\lizmap\classes\lizmapProject.class.php : 173
	2	lizmapProject->__construct()	C:\webserver\lizmap\lizmap\modules\lizmap\classes\lizmap.class.php : 298
	3	lizmap::getProject()	C:\webserver\lizmap\lizmap\modules\lizmap\classes\lizmapRepository.class.php : 150
	4	lizmapRepository->getProjects()	C:\webserver\lizmap\lizmap\modules\view\zones\main_view.zone.php : 47
	5	main_viewZone->_prepareTpl()	C:\webserver\lizmap\lib\jelix\utils\jZone.class.php : 239
	6	jZone->_createContent()	C:\webserver\lizmap\lib\jelix\utils\jZone.class.php : 206
	7	jZone->getContent()	C:\webserver\lizmap\lib\jelix\utils\jZone.class.php : 294
	8	jZone::_callZone()	C:\webserver\lizmap\lib\jelix\utils\jZone.class.php : 95
	9	jZone::get()	C:\webserver\lizmap\lib\jelix\tpl\jTpl.class.php : 125
	10	jTpl->assignZone()	C:\webserver\lizmap\lizmap\modules\view\controllers\default.classic.php : 108
	11	defaultCtrl->index()	C:\webserver\lizmap\lib\jelix\core\jCoordinator.class.php : 225
	12	jCoordinator->process()	C:\webserver\lizmap\lizmap\www\index.php : 23`

And this error persists in all projects I created, even in demo projects that are builtin.
Is this a known issue?
Does anyone have this issue?

OS: Windows10
QGIS desktop version: 3.4.13-ltr
QGIS Server version: 3.4.13-ltr
Lizmap version: 3.4.0pre.190808

laurentj:
[...]

PR #1436, opened 2019-12-19
Fix project cache with windows

laurentj:
Cache backends don't support '\' in a key. When using a file path as a key under windows, we must replace this character by '/'

Refs #1380


rldhont:
@laurentj thanks

PR #1450, opened 2020-01-02
New geolocation

nboisteault:
I have to improve Lizmap geolocation for a client but I don't want to use old code so I'm creating a geolocation web component which communicates to a Lizmap main object which is a proxy to Lizmap 3. I think we could do that for other parts of LWC in a release_3_5 branch. This way we could remove functionalities old code one by one and create new one the Lizmap 4 way.
We will then make the big jump (but smaller) to Lizmap 4. Lizmap main class will be modified by removing call to Lizmap 3 and adding its own properties for states.


nboisteault:
@rldhont @laurentj You can now review. @laurentj We'll need node in CI to build JS.

lizmap/www/js/edition.js
```diff
@@ -776,10 +764,13 @@ OpenLayers.Geometry.pointOnSegment = function(point, segment) {
                     // Get SRID and transform geometry
                     var srid = $('#edition-point-coord-crs').val();
                     vertex.transform( editionLayer['ol'].projection,'EPSG:'+srid );
-                    if ( !$('#edition-point-coord-x').attr('disabled') )
+                    if (srid === "4326"){
+                        $('#edition-point-coord-x').val(vertex.x.toFixed(6));
+                        $('#edition-point-coord-y').val(vertex.y.toFixed(6));
+                    }else{
                         $('#edition-point-coord-x').val(vertex.x);
```
rldhont:
```suggestion
                         $('#edition-point-coord-x').val(vertex.x.toFixed(3));
```

lizmap/www/js/edition.js
```diff
@@ -776,10 +764,13 @@ OpenLayers.Geometry.pointOnSegment = function(point, segment) {
...
-                    if ( !$('#edition-point-coord-y').attr('disabled') )
                         $('#edition-point-coord-y').val(vertex.y);
```
rldhont:
```suggestion
                         $('#edition-point-coord-y').val(vertex.y.toFixed(3));
```

lizmap/www/js/edition.js
```diff
@@ -835,12 +826,11 @@ OpenLayers.Geometry.pointOnSegment = function(point, segment) {
...
                     $('#edition-point-coord-y').attr('disabled','disabled');
-                    if ( lizMap.controls.geolocation.layer.features.length != 0 ) {
-                        var geometryType = editionLayer.geometryType;
-                        var vertex = lizMap.controls.geolocation.layer.features[0].geometry;
-                        var px = editCtrls[geometryType].handler.layer.getViewPortPxFromLonLat({lon:vertex.x,lat:vertex.y});
-                        editCtrls[geometryType].handler.modifyFeature(px);
-                    }
+
+                    var geometryType = editionLayer.geometryType;
+                    var [lon, lat] = lizMap.mainLizmap.geolocation.getPositionInCRS(editionLayer['ol'].projection);
```
rldhont: Are you sure that the geolocation return position ?
Are you sure that the geolocation has started ?

lizmap/www/js/edition.js
nboisteault: Are 3 decimals enough for any other cases ?

rldhont: Can we check if the projection is geographic ?

lizmap/www/js/edition.js
nboisteault: https://github.com/3liz/lizmap-web-client/pull/1450/commits/fd58c8ca7d5dc286e6a2d4d62e6e0c2d647d5af5

rldhont:
@nboisteault thanks

nboisteault:
@rldhont @laurentj it is now finished I think.

rldhont:
You can merge for me.

PR #1470, opened 2020-01-20
Display current/total segment length and angle in edition mode

nboisteault:
Angle measurement coming...


nboisteault:
@rldhont @mdouchin ready for review !

PR #1487, opened 2020-01-31
Multiatlas : handle 'layers' key. Fix popup display.

nboisteault:


Gustry:
I'm not the PHP expert but it looks good to me.

Can we also hide the atlas selection combobox when there is a single atlas? I think we noticed this yesterday.

nboisteault:
> I'm not the PHP expert but it looks good to me.
> 
> Can we also hide the atlas selection combobox when there is a single atlas? I think we noticed this yesterday.

@Gustry Done !

PR #1488, opened 2020-01-31
Editor: new event lizmapeditionfeatureinit to customize edition layers

laurentj:
It allows plugins to customize the style of the edition layers. Default colors of styles of features shown in the edit layer or split edit layer, may not be suitable, or maybe the same as colors of features displayed in the map. So it is difficult to use editing tools.


rldhont:
@laurentj can be merged ?

rldhont:
@laurentj can you update it ?

PR #1491, opened 2020-02-03
Set as maximal date the max date of the features

signedav:
I'm quite unsure about it. But as far as I understand the user should be able to select the last possible date in the date picker.


nboisteault:
Done in https://github.com/3liz/lizmap-web-client/pull/1497

PR #1494, opened 2020-02-04
Patch filter

nboisteault:
This fixes a bug @pcav wrote about in lizmap mailing list the 31/01/2020.

### DIFF
--- a/lizmap/www/js/filter.js
+++ b/lizmap/www/js/filter.js
@@ -349,7 +349,7 @@ var lizLayerFilterTool = function() {
                 var autocompleteData = [];
                 for(var a in result){
                     var feat = result[a];
-                    if( feat['v'] === null || !feat['v'] || feat['v'].trim() == '' )
+                    if (feat['v'] === null || !feat['v'] || (typeof feat['v'] === 'string' && feat['v'].trim() === '') )
                         continue;
                     autocompleteData.push(feat['v']);
                 }
--- a/lizmap/www/js/filter.js
+++ b/lizmap/www/js/filter.js
@@ -359,6 +359,7 @@ var lizLayerFilterTool = function() {
                     delay: 200,
                     minLength: 2,
                     select: function( event, ui ) {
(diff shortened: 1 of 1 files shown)


rldhont:
This issue does not exist in 3.2 ?

nboisteault:
Filter has been introduced in 3.3 isn't it ?

nboisteault:
@pcav can you test this PR on your project ? Thanks

rldhont:
> Filter has been introduced in 3.3 isn't it ?

Yes, you're right

pcav:
Will do ASAP, thanks!
What is the best way to test?

nboisteault:
With git : 
`git checkout -b nboisteault-patch-filter release_3_3`
`git pull https://github.com/nboisteault/lizmap-web-client.git patch-filter`

pcav:
The filter by country name seems fixed, the numeric filters still do not show up.
Is replacing the filters.js file enough for testing?
Thanks.

nboisteault:
As I just changed `filter.js` in this commit, yes it is enough. Numeric filters have to be fixed server side.

pcav:
Oh, I see. So what about disabling them for others in the plugin @Gustry , or at least give a warning?

rldhont:
@pcav if lizmap has 2 differents issues, it will be great to have 2 description in github/lizmap-web-client/issues and to PR to fix them.

pcav:
OK, will open a ticket for the numeric filters. Sorry I cannot be timely, now on intensive training

Gustry:
> Oh, I see. So what about disabling them for others in the plugin @Gustry , or at least give a warning?

Sure, it's usually what the plugin does.

According to the documentation:
https://docs.lizmap.com/next/en/publish/lizmap_plugin/form_filtering.html

> It works only for database layers: PostgreSQL (recommended), Spatialite and GeoPackage

So it should work with your GPKG too according to the doc

PR #1496, opened 2020-02-05
[BUGFIX] Hide header with parameter h=0 in landing page & use it for maps url

mdouchin:


rldhont:
It will be great to have CSS classes associated to these parameters.

PR #1536, opened 2020-02-20
[Bugfix] Read only form control cannot be required

rldhont:
Some changes in QGIS 3 force required option even if the field is not editable.


rldhont:
@drillinP can you test and valdiate ?

drillinP:
I tested and it freezes when submitting the form, so in PHP

PR #1590, opened 2020-04-09
Install external modules with Composer

laurentj:
Install source code of jCommunity and ldapdao with Composer, instead of copying their source code into our repository.


mdouchin:
Ok for me. Would the `make build` command do the magic for us ?

nboisteault:
@laurentj I followed instructions [here](https://github.com/3liz/lizmap-web-client/tree/master/lizmap/my-packages) till `composer install` but LWC did not start. Did I misunderstood ?

laurentj:
@mdouchin yes of course, this is already the case since weeks ;-)

@nboisteault sorry, but telling "LWC did not start" don't help me. And this is not related to this PR. What is the problem exactly?

nboisteault:
@laurentj I'd like to try in docker. Can you tell me how I can launch composer in docker machine ?

PR #1592, opened 2020-04-10
switch from datavizTheme to theme

Gustry:
I know, it's less readable like this for now.
But as soon as we will move to Lizmap 4, the setting will move to the dataviz section:

```
    "dataviz": {
        "layers": []
        "config": [
              "theme":"light"
         ]
    },
```

So just to be in the move to Lizmap 4 ;-)


laurentj:
I'm not sure you should only rename the property without keeping compatibility with "old" configuration files.

Gustry:
@laurentj This property has been added recently in LWC 3.4 and only last week in QGIS plugin. So there isn't old configuration. Maybe only a few projects @mdouchin used for dev.

laurentj:
@Gustry ok, that's fine then :)

PR #1597, opened 2020-04-17
Check QGIS Server URL and get QGIS Server version information

rldhont:
With a simple QGIS project, Lizmap can test QGIS Server URL and get release name and version number.

### DIFF
--- a/lizmap/modules/admin/controllers/config.classic.php
+++ b/lizmap/modules/admin/controllers/config.classic.php
@@ -65,6 +65,10 @@ class configCtrl extends jController
         $xmlLoad = simplexml_load_file($xmlPath);
         $version = (string) $xmlLoad->info->version;
 
+        // Get QGIS Server info
+        $qgisServer = jClasses::getService('lizmap~qgisServer');
+        $qgis_server_info = $qgisServer->info();
+
         // Get the data
         $services = lizmap::getServices();
 
@@ -77,6 +81,7 @@ class configCtrl extends jController
         $tpl->assign('showSystem', !lizmap::getServices()->hideSensitiveProperties());
         $tpl->assign('servicesForm', $form);
         $tpl->assign('version', $version);
+        $tpl->assign('qgis_server_info', $qgis_server_info);
         $rep->body->assign('MAIN', $tpl->fetch('config'));
         $rep->body->assign('selectedMenuItem', 'lizmap_configuration');
 
@@ -203,6 +208,17 @@ class configCtrl extends jController
         $ok = true;
         if (!$form->check()) {
             $ok = false;
+        }
+
+        // Check the WMS Server URL
+        $wmsServerURL = $form->getData('wmsServerURL');
+        jClasses::inc('lizmap~qgisServer');
+        if (!qgisServer::checkUrl($wmsServerURL)) {
+            $ok = false;
+            $form->setErrorOn(
+                'wmsServerURL',
+                jLocale::get('admin~admin.form.admin_services.message.wmsServerURL.wrong')
+            );
         }
 
         // Check the cacheRootDirectory : must be writable
--- a/lizmap/modules/admin/locales/en_US/admin.UTF-8.properties
+++ b/lizmap/modules/admin/locales/en_US/admin.UTF-8.properties
@@ -7,7 +7,10 @@ menu.lizmap.logs.label=Lizmap Logs
 menu.lizmap.logs.view.label=View logs
 
 generic.h2=Generic
-generic.version.number.label=Version number
+generic.version.number.label=Lizmap version number
+generic.qgis_server.info.label=QGIS Server version information
+generic.qgis_server.info.value=QGIS %s, version %s
+generic.qgis_server.info.error=Error with QGIS Server URL
 
 configuration.h1=Lizmap configuration
 configuration.repository.label=Repository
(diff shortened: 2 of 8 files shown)


rldhont:
@Gustry We can get QGIS Server version information with simple virtual field and a WFS GetFeature Request.

rldhont:
@laurentj any suggestion about the place to put the QGIS Project to get QGIS Server version information.

Gustry:
Question, it's already in the Lizmap server plugin
SERVICE=LIZMAP
REQUEST=GetServerSettings
```json
{
  "qgis": {
    "version": "3.13.0",
    "name": "Master"
  },
  "services": [
    "WMS",
    "WFS",
    "WCS",
    "WMTS",
    "ATLAS",
    "EXPRESSION",
    "LIZMAP"
  ],
  "lizmap": {
    "name": "Lizmap",
    "version": "3.3.99-beta"
  }
}
```
So it's jsut in the meantime?

Small documentation https://github.com/3liz/lizmap-plugin/#lizmap-server-api I made it quickly.

Gustry:
Nice idea!

rldhont:
@Gustry it is a way to get it without plugin.

laurentj:
@rldhont I'm not sure I understand all the code about this "fake" project. But if I well understood, you assume that Qgis server has an access to this qgis_info.qgs file, right?

If this is a case, I think it won't work in many case :

- when the `rootRepository` is set in the configuration
- when the QGIS server is installed in an other server/container, and only the directory  of real qgis projects is shared between servers/containers.

Anyway, I don't like this way to get informations about qgis server. I would prefer to enhance Qgis server or to create a plugin for Qgis Server. This could be the first feature to implement for the Qgis plugin we talked about weeks ago.

lizmap/modules/admin/controllers/config.classic.php
```diff
@@ -205,6 +210,17 @@ public function saveServices()
             $ok = false;
         }
 
+        // Check the WMS Server URL
+        $wmsServerURL = $form->getData('wmsServerURL');
+        jClasses::inc('lizmap~qgisServer');
```
laurentj, line 215: we should declare qgisServer as an autoloaded class into module.xml, and avoid to use jClasses.

rldhont:
> @rldhont I'm not sure I understand all the code about this "fake" project. But if I well understood, you assume that Qgis server has an access to this qgis_info.qgs file, right?
> 
> If this is a case, I think it won't work in many case :
> 
>     * when the `rootRepository` is set in the configuration
> 
>     * when the QGIS server is installed in an other server/container, and only the directory  of real qgis projects is shared between servers/containers.
> 
> 
> Anyway, I don't like this way to get informations about qgis server. I would prefer to enhance Qgis server or to create a plugin for Qgis Server. This could be the first feature to implement the Qgis plugin we talked about weeks ago.

The `rootRepository` is not a problem, because the project is not a lizmap one just a QGIS one.

The idea is to get QGIS Server information without plugins. The plugin code has been already done [1](https://github.com/3liz/lizmap-plugin/pull/196) but all Lizmap users do not install plugins and we have some difficulties to know QGIS Server version.

It is not a fake project but it has to be accessible for QGIS Server.

lizmap/modules/admin/controllers/config.classic.php
rldhont: `qgisServer` is also used with `jClasses::getService` https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/classes/lizmapProject.class.php#L684. can `qgisServer` be an autoloaded classes ?

laurentj:
>It is not a fake project but it has to be accessible for QGIS Server.

So it won't work at all on our servers

mdouchin:
We could probably easily use this as a alternate solution when no information could have been fetched from the server lizmap plugin. Check QGIS Server version in this order

* attempt to get information from lizmap plugin (server side).
* if lizmap plugin not installed, try to fetch it from this QGIS project
* if this QGIS project is not accessible, trust the configuration `lizmapConfig.ini.php`

rldhont:
> > It is not a fake project but it has to be accessible for QGIS Server.
> 
> So it won't work at all on our servers

Do you have an idea to provide this project ?

It will be necessary to have a QGIS project to get QGIS Server version information by a plugin and to get plugins informations.

mdouchin:
What about puttin this project in the QGIS Server VM/container/etc., with the environment `QGIS_PROJECT_FILE` variable to set this project path. If I understand correctly, QGIS Server will first look for some environment variable to get the project, which is overriden when using the MAP parameter ? If the QGIS project lies in the same server as where QGIS Server is installed, no issue then ?
See https://docs.qgis.org/3.10/en/docs/user_manual/working_with_ogc/server/config.html#qgis-project-file

rldhont:
@mdouchin With or without `QGIS_PROJECT_FILE`, it will be great to have a way to install this QGIS project and its data somewhere QGIS Server can access it.

@laurentj what is your recommendation about installing the QGIS project somewhere QGIS Server can access it but outside the FTP directory ?

nboisteault:
@rldhont  @laurentj @mdouchin Do I tag this PR to 3.5 ?

rldhont:
Done @nboisteault

Gustry:
Can we revive this pull request and find a solution?
It's very needed to have this information somehow.

I wish we can have a warning message bar for admins saying that they might have issues. 🙏

QGIS 3.16 changed the way to save the "Do not expose in WMS/WFS". QGIS Server < 3.16 is absolutely not able to read this configuration.

nboisteault:
@rldhont should we tag this PR for release 3.6?

Gustry:
IMHO, this PR is a dead end since https://github.com/3liz/lizmap-plugin/pull/335

Should we close it ?

Issue #1422, opened 2019-12-09
Update proj4php and install it with composer


nboisteault:

laurentj:
depends of #1056

laurentj:
@nboisteault please, do a pull request on the upstream project, against their proj4php5.2 branch, with fixes you made in commit 6ae3098bc046e70ce0c6ca8d0b8a7bb3b10e030f

As we still support php 5.6 for Lizmap 3.4, we should install the proj4php5.2 branch.

nboisteault:
@laurentj https://github.com/proj4php/proj4php/pull/95

julien2512:
merged

laurentj:
Thank you @nboisteault and @julien2512 .

There is still a difference with our version : our version has a defs EPSG3857 that is not into the proj4php5.2 branch. @rldhont  added probably it when it added sources into lizmap for the first time. I will propose a patch.

laurentj:
proj4php/proj4php#97

PR #1598, opened 2020-04-19
Replace inline sources of Proj4Php by its Composer package

laurentj:
Refs #1422


laurentj:
Probably we could remove the proj4php module entirely, if no external modules are using `jClasses::create('proj4php~proj4php')`. 

@rldhont what do you think about removing the proj4php module ? Should we wait for Lizmap 3.5?

rldhont:
Don't wait for Lizmap Web Client 3.5. I am happy if you can install proj4php with composer.

laurentj:
I will remove the module later

PR #1600, opened 2020-04-20
Add an hidden table to popup with children

nboisteault:


nboisteault:
@mdouchin finally I've added a table in backend. I'm going to add an HTML button to toggle visibility now.

nboisteault:
compact_explode_table

josemvm:
@nboisteault thanks for this new feature!

did you define some criteria for children ordering in popup?

i think it would be a good idea to have a sorting criterion, for example the primary key or another field to be defined in the plugin... and not a random order

mdouchin:
thanks @nboisteault Ordering could be achieved perhaps with no pain by "datatabling" the table, so that people can order by any columns

```javascript
$('#myTable').DataTable();
```
See: https://datatables.net/examples/basic_init/zero_configuration.html

rldhont:
You can merge

PR #1611, opened 2020-04-22
move close button in right dock

josemvm:
move close button in right dock and align marker and headers between docks

imagem


nboisteault:
@josemvm thank you !

PR #1629, opened 2020-05-07
Fix bug in Chrome when scrolling in map

### LINKED ISSUE #1619: mouse wheel issue in Chrome when lizmap is in an iframe
### What is the bug?

In Chrome and in some version of Firefox, when using the mouse wheel or the trackpad to zoom in/out on the map, the whole page scroll at the same time.

- Chrome MacOS : bug
- Chrome win10 : bug
- Chrome linux + trackpad : bug
- Chrome linux + molette souris : bug
- Firefox linux + trackpad : OK
- Firefox linux + molette souris : OK
- Firefox Mac + trackpad : bug
- Firefox Mac + molette souris : OK
- Firefox win10 : OK
- Safari MacOS : OK
- Edge win10 : OK

1. use an iframe into a web page, using the `/index.php/view/embed/` url to display lizmap
2. put the mouse over the map, and use the mouse wheel or the trackpad to zoom
3. the map zoom in or zoom out, but the page also scrolls.

### Expected behavior

The page should not scroll.

### Environment

 - Lizmap version: 3.3
 - OS : Windows, Linux, MacOS 
 - Browser : Chrome

nboisteault:
Message in console was : [Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/6662647093133312
Fix #1619


nboisteault:
@laurentj Yes I realized that after I commit.

PR #1651, opened 2020-05-22
Add env variable for customising php

NyakudyaA:
* Add a couple of more env variable in lizmap-entrypoint
* Update README


dmarteau:
Looks good to me. 

Note that at some point, it is not desirable to add all PHP-FPM configuration variables in the entrypoint and one should consider mounting a  customized `www.conf` as a volume file.

PR #1662, opened 2020-05-30
Update jQuery to last version

### LINKED ISSUE #1511: Please upgrade jQuery version
The jQuery version used by Lizmap is obsolete, unsupported, and may represent a non negligible security risk. Standard penetration tests reports this as a grave vulnerability.
A guide is available for the migration:
https://jquery.com/upgrade-guide/3.0/
together with plugins that should help smooth it out.
I suggest upgrading it ASAP. I'm available for support.
Thanks a lot.

nboisteault:
@rldhont @laurentj @mdouchin I'm testing an update of jQuery to last version with help of jquery-migrate. Until now it doesn't break everything. jquery-migrate gives us warnings to find deprecated code.
Fix #1511


nboisteault:
@rldhont @mdouchin @laurentj : I'm looking to update jquery-ui. It seems we only need those widgets : https://jqueryui.com/download/#!version=1.12.1&components=111111011111101000111111110000000000000000000000
Can you confirm ?

rldhont:
@nboisteault I think so

nboisteault:
Jquery-ui is dying. [You can't use their builder for 10 months now](https://github.com/jquery/download.jqueryui.com/issues/359). There is a mirror [here](https://jqueryui.templersmc.net/download) but that the signal we'll have to drop it from LWC. No big deal as we don't have many jquery-ui widgets.

nboisteault:
I won't update jQuery-ui as I prefer to spend time later to remove this dependency.
There are many warnings linked to deprecated code when I use jquery-migrate to display those but they'll disappear when we will refactor code for LWC 4 and also update to Bootstrap 5.
I can't find any other bug so I think we can merge in master.

nboisteault:
> I cannot approve that PR yet, as jQuery is used also by Jelix (jForms and other components), and so we should study the impact of this new version on Jelix.

Can you list tests that should be made ?

laurentj:
I will review Jelix JS code to see if there are incompatibilities

laurentj:
Reviewed Jelix parts : it is ok to use jQuery 3.5. On the jelix 1.7 branch, I already migrated from jQuery 1.10 to 3.3 and it was ok.

The issue here is to use jQueryUI. Upgrading to 1.12 should be a first step. I will do it as I made a custom build. I don't remember yet what was included but  I should have the recipe somewhere on my disk.

Anyway, we need jQueryUI because some components are criticals:

- the datepicker, which is used by jForms (and so by Lizmap in edition forms). I never found a datepicker that is as configurable and usable as the jqueryui's one (this was the primary reason to use jQueryUI in fact)
- the autocomplete component, also used by jForms, Lizmap and in some custom modules
- the dialog component. I don't remember if we are using it in Lizmap, but I'm using it in some custom modules for Lizmap

If we want to remove this dependency, we should find some alternate good components.

I propose you to open a new ticket about the replacement of jQueryUI.

nboisteault:
https://github.com/3liz/lizmap-web-client/issues/1678

PR #1663, opened 2020-06-01
Fix: Wrong UPDATE query sent to PostgreSQL, data loss

### LINKED ISSUE #1647: Wrong UPDATE query sent to PostgreSQL, data loss
### What is the bug?

I have just noticed the following, source of big concerns.

1) I have a PostGIS polygon table/layer which has quite a few columns but not many records (I don't think this is important for the problem)

2) The layer is added to a minimal project (just this layer). The project is set up to be served with QGIS Server and the Lizmap configuration file is generated. The layer is configured to be editable (insert, update, delete) in Lizmap.

3) If the layer has **no** D&D form defined (so using the "auto generate" QGIS option) there are no issues when updating attributes within LMWC.

4) The issue arise when using a D&D form: in this specific test case in the form it was added just the "gid" (PK) field, another field that happens to have been defined as int8, and another field with datatype varchar. All the other attributes of this layer are left out of the form,

5) A feature is edited in LMWC with this form, and the value of this int8 column is changed. Edits are saved.

6) Inspecting the table I can see how the value of the int8 column was updated correctly, and the value of the varchar column was left unchanged (as expected), but for my great surprise all the values of the other columns (that are not part of the form) were completely **wiped out**.

7) I then inspected the UPDATE query sent by LMWC to PostgreSQL and as a fact all the wiped out fields were set as NULL

The query looks like:

```
UPDATE "schemaname"."table" SET "geom"=ST_GeomFromText('MULTIPOLYGON(((-5463.366843911836 -7237.887256608387,-5452.475539068239
 -7228.778165316916,-5427.128502341457 -7211.649113131462,-5412.27672300934...)))', 3763), "id_parceir"=NULL, "parcela"=NULL, "tipo_benef"=NULL, "nome_benef"='First Last', "nif_benef"=NULL, "telf_benef"=NULL, "mail_benef"=NULL, "parceiro"=NULL, "regiao_celpa"=NULL, "concelho"=NULL, "freguesia"=NULL, "rot_atual"=NULL, "ano_inicio"=NULL, "mes_inicio"=NULL, "densidade"=NULL, "area_ha"=NULL,
"certif_fl"=NULL, "grupo_cert"=NULL, "codig_cert"=NULL, "sist_cert"=NULL, "status_cvi_el"=NULL, "tipo_veg_el"=NULL, "grau_cobert_el"=NULL, "altura_veg_el"=NULL, "metodo_cvi_el"=NULL, "status_cvi_l"=NULL, "tipo_veg_l"=NULL, "g
rau_cobert_l"=NULL, "altura_veg_l"=NULL, "metodo_cvi_l"=NULL, "psf_cvi"=NULL, "data_cvi"=NULL, "controlo_cvi"=NULL, "observ_cvi"=NULL, "status_svaras"=NULL, "data_svaras"=NULL, "controlo_svaras"=NULL, "observ_svaras"=NULL, "s
tatus_nutricional"=NULL, "metodo_adubacao"=NULL, "tipo_adubo"=NULL, "dose_g_pl"=NULL, "nr_sacos"=NULL, "qtdd_adubo"=NULL, "dose_kg_ha"=NULL, "ano_adubacao"=NULL, "psf_aduba"=NULL, "data_adubacao"=20201006, "controlo_adubacao"
=NULL, "observ_adubacao"=NULL, "estado_candidatura"=NULL, "estado_informacao"=NULL, "estado_processo"=NULL, "registo_foto"=NULL, "data_registo"=NULL, "ultima_atualizacao"=NULL, "cod_programa"=NULL, "obs_gerais"=NULL, "grupo_webgis"=NULL, "obsv_hv"=NULL WHERE "gid" = 3907  RETURNING "gid";
```

As you can see all fields where set to NULL but "data_adubacao" that is the field that has been edited and "nome_benef" that is the other field in the form. All values of fields not in the form were wiped out.

Additional notes:

1) if the fields that are not in the form are **not** published as WMS/WFS in the layer properties in QGIS then there is no data loss, but this may not be a workaround for all cases

2) another workaround is to put all fields in the D&D form, which again may not be an option for all cases.

3) it is unclear why LMWC sends the UPDATE command for all fields including the ones that were not changed at all (included the geometry). For instance this is the query sent directly by QGIS when chancing just the value of 1 field:

`UPDATE "schemaname"."table SET "data_adubacao"=20201006 WHERE "gid"=3907`

### Environment

 - Lizmap version: 3.3.6
 - QGIS Server version: 3.10.5
 - QGIS Project version: 3.10.5
 - OS (Windows, Linux, MacOS, Android…): Linux (server), Windows/Linux (client)
 - Browser (Firefox, Chrome…): Firefox and Chrome, others not tested
 - Backend: PostgreSQL 10

rldhont:
Fixed #1647 

Check field has a jFormsControl defined in form.

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisForm.class.php
+++ b/lizmap/modules/lizmap/classes/qgisForm.class.php
@@ -537,9 +537,13 @@ class qgisForm
         $values = array();
         // Loop though the fields and filter the form posted values
         foreach ($fields as $ref) {
-            if ($form->getControl($ref) instanceof jFormsControlUpload) {
+            $jCtrl = $form->getControl($ref);
+            // Field not in form
+            if ($jCtrl === null) {
+                continue;
+            }
+            if ($jCtrl instanceof jFormsControlUpload) {
                 $values[$ref] = $this->processUploadedFile($form, $ref, $cnx);
-
                 continue;
             }
 


nboisteault:
@rldhont In my test `$form->getControl($ref);` does not return `null` for fields not in form. So the bug still occurs.

rldhont:
superseeded by https://github.com/3liz/lizmap-web-client/pull/1667

PR #1673, opened 2020-06-05
Display OL6 scalebar instead of OL2 one.

nboisteault:


nboisteault:
@rldhont @mdouchin The OL6 scalebar widget is at bottom left on the screenshot : 
- Do you want some change on its look ?

- Do I put OL6 scalebar where OL2 scalebar was or somewhere else ?

- Scale values between OL2 and OL6 are different. I tried `geodesic:true` for the OL2 one but values are even more different then. This parameter does not exist in OL6. Have you some ideas why there is this difference?

image

mdouchin:
I think we should choose between:

* keep the old position, and replace OL2 scalebar+scale with the new one.
* put them outside the "overview" thing, at the bottom left of the map. The problem is the size of the left dock which changes and is not stable.

The scale value difference could be linked to the DPI we have set to 96 instead of default 72 for OL2, to have a correct "scale visibility" per layer (between QGIS and Lizmap)

nboisteault:
About the scale value difference, in next OL6 version it will possible to set DPI value with [this PR](https://github.com/openlayers/openlayers/pull/10537).

nboisteault:
The scalebar width can be very wide (size is dynamic) so I let the scale text on top and made #overview-bar taller.
Here is what it looks like : 
image

josemvm:
much bigger...

nboisteault:
@josemvm any suggestion?

josemvm:
>@josemvm any suggestion?

reduce by 50% if possible...

nboisteault:
> > @josemvm any suggestion?
> 
> reduce by 50% if possible...

PR are welcomed...

josemvm:
;-)

PR #1689, opened 2020-06-17
Move wms request part 2

rldhont:
## Description

Move WMS request logic from controllers to dedicated class `lizmapWmsRequest` part 2. The first part https://github.com/3liz/lizmap-web-client/pull/1682

Needed for features layers access control.


rldhont:
@laurentj it is ready for you ?

PR #1698, opened 2020-06-25
Fix lizmapWFSRequest parameters access

rldhont:
## Description

The acces to parameters in `lizmapWFSRequest` is not based on param function.


rldhont:
Thanks @nboisteault

PR #1704, opened 2020-06-30
Tests - add travis config

Gustry:
## Description

Quick try to setup Travis
I would prefer GitHub actions, but it was hanging and I didn't want to spend time setting this up

As @mdouchin is writting some PHP tests next to me ;-)

CC @laurentj FYI


laurentj:
r+

Issue #1722, opened 2020-07-17
Editing and PostgreSQL environment vars/.pgpass file


shermp:
### Description

I am attempting to get layer editing working for Lizmap WITHOUT having the Postgres credentials stored in the qgs file. I've tried setting PGUSER/PGPASSWORD environment variables. I've tried setting PGPASSFILE. I've tried using ~/.pgpass.

Looking at the code, a service would work... except that lizmap assumes the service will also contain the DB name, so never sets it, making a "generic" service unusable.

For everything I've attempted, I've gotten a very generic `[402]	error during the connection pgdb	/www/lib/jelix/plugins/db/pgsql/pgsql.dbconnection.php	153` error from the log, which tells me very little.

I'm running a docker stack with PostGIS, py-qgis-server, lizmap 3.3.x, nginx. I've played with permissions. I've tried creating a user (for the pgpass file).

My last resort will be to hack `lizmap/modules/lizmap/classes/qgisVectorLayer.class.php` to read some env variables. But I really don't want to do this if there is any other way.

Thank you for your time.

### Environment

 - Lizmap version: 3liz/lizmap-web-client:3.3
 - QGIS Server FULL version: 3liz/qgis-map-server:3.10
 - QGIS Desktop FULL version: N/A
 - OS: Ubuntu 18.04 with Docker 
 - Browser: N/A
 - Lizmap admin information (users, logs…) stored into PostgreSQL

mdouchin:
The way Lizmap parses the PostgreSQL layer datasource was fragile in versions <= 3.3.X. Service worked well with older QGIS ( 2.18 & 3.4). Not fully tested yet with 3.10.

In current master, and future 3.4.X, we have improved the datasource parsing, and added unit tests to cover some of the datasources. We did not dare to backport to 3.3.x branches, but you should be able to test the changes in your 3.3.x installation (do not try in production environment first !)

See: https://github.com/3liz/lizmap-web-client/pull/1706/files

A test should be added to cover the service both with and without database name. Could you provide here the full `<datasource>` of your layer (with service, no dbname and obviously no password) ?

shermp:
Thanks for the quick reply! We're setting up a brand new system, so nothing to break in production yet 😄 Means I can play with different versions etc.

Just upgraded to the currently available 3.4 docker image (alpha 2 it appears to be). Editing still doesn't work with the current setup.

Here's the datasource for my test edit layer:
```
[...]

PR #1723, opened 2020-07-17
QGIS Vector Layer - Datasource: dbname can be given when service is used

mdouchin:
The dbname parameters is not mandatory in PostgreSQL service file.
QGIS vector datasource stores the dbname: we pass it to connection parameters
even if service is given, when dbname is not empty

Related ticket : #1722


mdouchin:
@laurentj this PR modifies the Jelix core lib.

PR #1740, opened 2020-07-29
Upgrade jQuery File Upload to version 10.31.0

rldhont:
Fix a regression in lizmap file upload

https://github.com/blueimp/jQuery-File-Upload


rldhont:
The jQuery File Upload version in lizmap is v9.28.0 integrated in april 2019.

PR #1747, opened 2020-08-05
adding lizmapServices tests

alagroy-42:
Adding tests for the lizmapServices class

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmap.class.php
+++ b/lizmap/modules/lizmap/classes/lizmap.class.php
@@ -37,6 +37,10 @@ class lizmap
     // log items
     protected static $logItems = array();
 
+    // lizmapServices counter
+    protected $lizmapServices = 0;
+    protected $lizmapServicesInstance = null;
+
     /**
      * this is a static class, so private constructor.
      */
@@ -47,10 +51,36 @@ class lizmap
     /**
      * @return lizmapServices
      */
-    public static function getServices()
+    public static function	getServices()
     {
-        return jClasses::getService('lizmap~lizmapServices');
-    }
+        if (isset($lizmapServices) && !$lizmapServices)
+        {
+                $lizmapConfigTab = parse_ini_file(jApp::configPath('lizmapConfig.ini.php'), true);
+                $globalConfig = jApp::config();
+				$ldapEnabled = jApp::isModuleEnabled('ldapdao');
+				$varPath = jApp::varPath();
+                $lizmapServicesInstance = new lizmapServices($lizmapConfigPath, $globalConfig, $ldapEnabled, $varPath);
+                $lizmapServices = 1;
+        }
+        return $lizmapServicesInstance;
+        //return jClasses::getService('lizmap~lizmapServices');
+
+	}
+	
+	public static function	saveServices()
+	{
+		$ini = new jIniFileModifier(jApp::configPath('lizmapConfig.ini.php'));
+        $liveIni = new jIniFileModifier(jApp::configPath('liveconfig.ini.php'));
+
+		$services = self::getServices();
+		$services->saveIntoIni($ini, $liveIni);
+
+		$modified = $ini->isModified() || $liveIni->isModified();
+        $ini->save();
+        $liveIni->save();
+
+        return $modified;
+	}
 
     /**
      * @return lizmapTheme
--- a/lizmap/modules/lizmap/classes/lizmap.class.php
+++ b/lizmap/modules/lizmap/classes/lizmap.class.php
@@ -38,8 +38,7 @@ class lizmap
     protected static $logItems = array();
 
     // lizmapServices counter
-    protected $lizmapServices = 0;
-    protected $lizmapServicesInstance;
+    protected static $lizmapServicesInstance = null;
 
     /**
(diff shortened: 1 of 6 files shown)


tests/units/classes/lizmapServicesTest.php
```diff
@@ -0,0 +1,291 @@
...
+
+	/**
+	 * @dataProvider getModifyGlobalData
+	 */
+
+	public function	testModifyGlobal($globalConfig, $newConfig, $changedProperty, $changedValue, $expectedReturnValue)	{
+
+		$testLizmapServices = new LizmapServices(array(), (object)$globalConfig, false, '');
+		$this->assertEquals($expectedReturnValue, $testLizmapServices->modify($newConfig));
+		if (isset($changedProperty))	{
```
laurentj: `if ($changedProperty)` is better (or `if ($changedProperty !== null)`.

lizmap/modules/lizmap/classes/lizmapServices.class.php
```diff
@@ -339,42 +346,27 @@ public function update($data)
...
+		
         foreach ($this->properties as $prop) {
             if ($dontSaveSensitiveProps && isset($hiddenProps[$prop])) {
                 continue;
             }
             if (isset($this->globalConfigProperties[$prop])) {
                 list($key, $section) = $this->globalConfigProperties[$prop];
                 $liveIni->setValue($key, $this->{$prop}, $section);
-            } elseif ($this->{$prop} != '') {
+            } elseif (isset($this->{$prop}) && $this->{$prop} != '') {
```
laurentj, line 365: without this `isset`, are there some issues? because it seems to me this is useless

tests/units/classes/lizmapServicesTest.php
```diff
@@ -0,0 +1,291 @@
...
+		);
+	}
+
+	/**
+	 * @dataProvider getSaveIntoIniData
+	 */
+
+	public function testSaveIntoIni($dataModification, $expectedIniValues, $expectedLiveIniValues, $section_name, $hide)
+	{
+		$iniPath = __DIR__.'/../tmp/local.ini.php';
```
laurentj: Creation of `tmp/` is missing from your PR. So the file_put_contents function will fail. Add an empty file into it (`.empty` for example), so it could be added into the repository.

tests/units/classes/lizmapServicesTest.php
```diff
@@ -0,0 +1,291 @@
...
+		$ini1_1 = array(
+			'appName' => 'Lizmap'
+		);
+		$ini2 = array();
+		$liveIni = array(
+			'webmasterEmail' => 'test.test@test.com',
+			'webmasterName' => 'Adrien'
+		);
+		return array(
+			array('$testLizmapServices->appName = "Lizmap"; $testLizmapServices->adminContactEmail = "test.test@test.com";', $ini1, null, 'services', false),
```
laurentj: I don't like this very much. Using `eval` is considered as an evil function ;-). You should set an array here, instead.

laurentj:
You forgot also to change the call of the `save()` method, which is called by the `update` method. Delete this `update` method, find where it is called, and replace it by `lizmap::saveServices()`

lizmap/modules/lizmap/classes/lizmapServices.class.php
alagroy-42: There is a 'cacheExpiration' in the properties tab that is not a class property so there is an 'Undefined property' PHP Error.

laurentj: So this is a missing name. we should add it.

Gustry:
@laurentj @alagroy-42 I have setup Travis a few weeks ago. But feel free of course to update it or skip it. I will have no idea to know if it's a false positive or if it's a real fail. We were aware that this Travis setup is quite limited for now, it was just in the short term to have something running. Thanks for tackling these tests !

laurentj:
@Gustry it was a test with an hard-coded path into it, which is not good. @alagroy-42 has been fixed it.

lizmap/modules/lizmap/classes/lizmap.class.php
```diff
@@ -49,7 +52,30 @@ private function __construct()
...
+    {
+        $ini = new jIniFileModifier(jApp::configPath('lizmapConfig.ini.php'));
+        $liveIni = new jIniFileModifier(jApp::configPath('liveconfig.ini.php'));
+
+        $services = self::getServices();
+        $services->saveIntoIni($ini, $liveIni);
+
+        $modified = $ini->isModified() || $liveIni->isModified();
+        $ini->save();
+        $liveIni->save();
```
rldhont, line 76:
```suggestion
        $modified = $ini->isModified() || $liveIni->isModified();
        if ($modified) {
            $ini->save();
            $liveIni->save();
        }
```
Is `save` necessary even if ini files are not modified ?

laurentj: no, not necessary

rldhont: My suggestion is not necessary, I have read that the old code save even if ini files are not modified

PR #1749, opened 2020-08-06
Don't sort values with ValueMap widget to keep order defined in QGIS

nboisteault:
* Ticket : support # 954

### DIFF
--- a/CHANGELOG-3.2.md
+++ b/CHANGELOG-3.2.md
@@ -4,9 +4,9 @@ Changelog 3.2
 Version 3.2.16 (next)
 ---------------------
 
-- update some libraries to fix some bugs :
+- Update some libraries to fix some bugs :
     - Upgrade jQuery File Upload to version 10.31.0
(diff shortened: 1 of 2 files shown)


nboisteault:
@mdouchin @rldhont do you see some other widgets that should not be sorted by default ?

rldhont:
In value relation but it is already manage by config.

Gustry:
Just for information, I have noticed value map has sometimes a different data structure, either a dict or a list of dict. We encountered the issue in February with @rldhont with the Calvados.

I wrote unittest in Lizmap plugin:
* list of dict https://github.com/3liz/lizmap-plugin/blob/master/lizmap/test/test_tooltip.py#L158
* dict only https://github.com/3liz/lizmap-plugin/blob/master/lizmap/test/test_tooltip.py#L178

Sorry if I'm off topic, just to raise one issue I got about sorting a ValueMap.

PR #1753, opened 2020-08-10
Lizmap log config

alagroy-42:
Adding tests for lizmapLogConfig class + remodeling some things in the class.
Moved the the getLogItem and getLogItemList methods from the lizmap class to lizmapLogConfig

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmap.class.php
+++ b/lizmap/modules/lizmap/classes/lizmap.class.php
@@ -29,9 +29,6 @@ class lizmap
      */
     protected static $repositoryInstances = array();
 
-    // log items
-    protected static $logItems = array();
-
     // lizmapServices instance
     protected static $lizmapServicesInstance = null;
 
@@ -326,31 +323,10 @@ class lizmap
     public static function getLogConfig()
     {
         if (!self::$lizmapLogConfigInstance) {
-            $readConfigPath = parse_ini_file(jApp::varPath().self::lizmapLogConfig, true);
+            $readConfigPath = parse_ini_file(jApp::varPath().self::$lizmapLogConfig, true);
             self::$lizmapLogConfigInstance = new lizmapLogConfig($readConfigPath);
         }
         return self::$lizmapLogConfigInstance;
-    }
-
-    /**
-     * Get a list of log items names.
-     *
-     * @return string[] list of names
-     */
-    public static function getLogItemList()
-    {
-        // read the lizmap log configuration file
-        $readConfigPath = parse_ini_file(jApp::varPath().self::$lizmapLogConfig, true);
-        $logItemList = array();
-        foreach ($readConfigPath as $section => $data) {
-            $match = preg_match('#(^item:)#', $section, $matches);
-            if (isset($matches[0])) {
-                $logItemList[] = str_replace($matches[0], '', $section);
-            }
-        }
-        self::$logItems = $logItemList;
-
-        return self::$logItems;
     }
 
     /**
@@ -381,7 +357,8 @@ class lizmap
         return new lizmapLogItem($key);
     }
 
-    /* Returns time spent in milliseconds from beginning of request
+    /**
+     * Returns time spent in milliseconds from beginning of request
      * @param string $label Name of the action to lo
      */
     public static function logMetric($label, $start = 'index')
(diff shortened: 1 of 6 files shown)


lizmap/modules/lizmap/classes/lizmapLogItem.class.php
```diff
@@ -54,18 +54,21 @@ class lizmapLogItem
         'email',
     );
 
-    public function __construct($key)
+    public function __construct($key, $readConfigPath)
```
rldhont, line 57: Can you add the method documentation ?

lizmap/modules/lizmap/classes/lizmapLogItem.class.php
```diff
@@ -54,18 +54,21 @@ class lizmapLogItem
...
 
         // Check if this item exists in the ini file
         if (array_key_exists($section, $readConfigPath)) {
             // Set each property
             foreach (self::$properties as $property) {
-                $this->data[$property] = $readConfigPath[$section][$property];
+                if (isset($readConfigPath[$section][$property])) {
+                    $this->data[$property] = $readConfigPath[$section][$property];
+                }
+                // not sure about those lines, are every properties essential ? if yes uncomment
```
rldhont, line 68: every properties are not essential. `getData` can return `null` if the property is not defined in `$this->data`.

lizmap/modules/lizmap/classes/lizmapLogConfig.class.php
```diff
@@ -56,6 +89,9 @@ public function getProperties()
     public function modify($data)
     {
         $modified = false;
+        if (!$data) {
+            return false;
```
rldhont, line 93:
```suggestion
            return $modified;
```

lizmap/modules/lizmap/classes/lizmap.class.php
```diff
@@ -322,28 +322,11 @@ public static function getProject($key)
...
     {
-        return jClasses::getService('lizmap~lizmapLogConfig');
-    }
-
-    /**
-     * Get a list of log items names.
-     *
-     * @return string[] list of names
-     */
-    public static function getLogItemList()
```
laurentj, line 333: We should not remove this method, to avoid issues with existing external modules, and to avoid an issue in `getLogItem()`.
Mark it as deprecated, and change its code to use new methods

lizmap/modules/lizmap/classes/lizmap.class.php
```diff
@@ -374,7 +357,8 @@ public static function getLogItem($key)
         return new lizmapLogItem($key);
```
laurentj, line 357: you should change getLogItem to use new methods

laurentj: mark it as deprecated

lizmap/modules/lizmap/classes/lizmapLogItem.class.php
```diff
@@ -54,18 +54,26 @@ class lizmapLogItem
...
 
-    public function __construct($key)
+    /**
+     * Construct the object, you should use the lizmapLogConfig::getLogItem() method
+     * which will call this constructor
+     *
+     * @param string $key the name of the item
+     * @param array $readConfigPath the array containing the fields of lizmapLogConfig.ini.php
+     */
+    public function __construct($key, $readConfigPath)
```
laurentj, line 64: $readConfigPath should be only the content of the section corresponding to the item. It would avoid to check if the section exists etc, since it has already be done into getLogItem().

lizmap/modules/lizmap/classes/lizmapLogConfig.class.php
```diff
@@ -72,11 +108,11 @@ public function modify($data)
      *
      * @param array $data array containing the data of the general options
      */
-    public function update($data)
+    public function update($data, $ini)
```
laurentj, line 111: this change of this method annoy me, as it could break some modules if this method is used (it seems it is not used into Lizmap directly).

I propose to set this `$ini` parameter as optional : `update($data, $ini=null)`. Same for `save($ini=null)`. And in the `save()` method, if `$ini` is null, the jIniFileModifier should be instancied.

PR #1755, opened 2020-08-12
adding documentation generation environment

alagroy-42:
Adding a documentation generation environment working with phpDocumentor.
It should automatically generate web pages with the documentation and host it on GitlabPages


rldhont (approved):

Gustry:
Side question, why is-it closed?

Gustry:
Thanks for reopening.
Your stage is missing a `tags` statement, it won't be executed by gitlab AFAIK.
What do you do with generated files in `docs/` ? Will it be published automatically in Gitlab pages?
Maybe you should do this PR in Gitlab to make it run by the CI?

Let's wait for @laurentj

rldhont:
Makefile en conflit

rldhont:
@alagroy-42 conflict to resolve

mdouchin (approved):
LGTM thanks

nboisteault:
@rldhont @mdouchin what do we do with this PR?

Gustry:
IMHO, we should have the HTML generation.
So we should : 
* rebase, fix conflict
* check that the given `make doc` works and generate some HTML files

The gitlab-ci file will fail for now because there isn't any `tag` statement. Then later, we can decide where to publish the HTML, on docs.3liz.org temporarily ?, docs.lizmap.com ?

Gustry:
I think it would be nice to have, but I'm not a PHP dev so I won't be able to do something here.

PR #1757, opened 2020-08-13
adding JelixInfos class

alagroy-42:
Adding the JelixInfos class that allows us not to use Jelix directly in class methods.
This class will also allows us to alter the Jelix methods in the tests so we can create scenarios that can happens in some situations without having to directly touch the tested class or Jelix utilities.

This class is currently only populated with the utilities I need to test and modify the lizmapProject class but I will populate it with other utilities when I will need them.


laurentj:
I merge but I'm going to add comment and change some method signatures

PR #1758, opened 2020-08-14
Lizmap projects no conflicts

alagroy-42:
Removing Jelix and lizmap context + correcting/optimizing code.


laurentj:
The lizmapAppContext interface is missing.

Method names of lizmapAppContext should not be related to jelix. for example:

- `jAcl2CheckResult()` may be `aclCheck()`
- `jAcl2DbUserGroups()` may be `aclUserGroups()`
- `jAuthIsConnected()` may be `isConnectedUser()`
- `createJForm` may be `createForm`
- Etc.

`jAppInfos`, `jCacheHandler`, `jDaoHandler`: i would prefer a method for each static method of the object we call.

I don't understand why you created these methods:  `useDbConnection`, `tplAssign`, `tplFetch`.

alagroy-42:
The interface is in lizmapAppContext.interface.php.

I will rename the methods.

Tpl functions are indeed useless.
Are we assuming that all the Db queries are correct and that they cannot fail. If yes, I will remove useDbConnection also.

PR #1759, opened 2020-08-17
fixing error 500

alagroy-42:
There was an error 500 on admin.php/admin/maps because Jelix doesn't allow :: in Tpl.

### DIFF
--- a/lizmap/modules/admin/templates/maps.tpl
+++ b/lizmap/modules/admin/templates/maps.tpl
@@ -21,7 +21,7 @@
                 <dd>
                     <table class="table">
                         {assign $section = 'repository:'.$repo->getKey()}
-                        {assign $properties = lizmapRepository::getProperties()}
+                        {assign $properties = $repo->getRepoProperties()}
                         {foreach $properties as $prop}
                             <tr>
                                 {if $prop == 'path' && $rootRepositories != ''}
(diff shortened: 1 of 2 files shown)


lizmap/modules/admin/templates/maps.tpl
```diff
@@ -21,7 +21,7 @@
                 <dd>
                     <table class="table">
                         {assign $section = 'repository:'.$repo->getKey()}
-                        {assign $properties = lizmapRepository::getProperties()}
+                        {assign $properties = $repo->getRepoProperties()}
```
nboisteault, line 24: Why not using `$repo->getProperties()` and create `$repo->getRepoProperties()`

alagroy-42: Because getProperties is a static method that can be used out of lizmapRepository so we don't have to create a lizmapRepository instance just to access its list of properties which is always the same.

nboisteault: @laurentj is it normal static methods can't be called in templates ?

laurentj: Yes we can call static methods in theory. But here, since we've got the repository object, we can call directly the method on the object.

PR #1761, opened 2020-08-17
Move and rework the new AppContextInterface interface

laurentj:
More comments. Some methods have been renamed, in order to be less dependent to Jelix.

### DIFF
--- a/lizmap/modules/lizmap/lib/App/JelixContext.php
+++ b/lizmap/modules/lizmap/lib/App/JelixContext.php
@@ -30,7 +30,7 @@ class JelixContext implements AppContextInterface
      */
     public function aclCheck($role, $resource = null)
     {
-        return \jAcl2::check($role);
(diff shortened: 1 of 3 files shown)


lizmap/modules/lizmap/lib/App/JelixContext.php
```diff
@@ -0,0 +1,183 @@
...
+    /**
+     * says if the current user has the given right. Call jAcl2::check().
+     *
+     * @param string $right the key of the right to check
+     * @param string $resource the id of a resource if any
+     * @return bool The result of jAcl2::check()
+     */
+    public function aclCheck($role, $resource = null)
+    {
+        return \jAcl2::check($role);
```
alagroy-42, line 33:
```suggestion
        return \jAcl2::check($role, $resource);
```

We sould send the resource to jAcl2::check()

PR #1768, opened 2020-08-20
Lizmap\Project\Repository

alagroy-42:
Moving lizmapRepository to namespace Lizmap\Project\ + marking lizmapRepository as deprecated.
The lizmapRepository class is still there to avoid breaking the API but it's now storing and using a Lizmap\Project\Repository.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ services:
 branches:
   only:
     - master
+    - refactor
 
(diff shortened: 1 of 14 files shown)


Gustry:
There is a conflict on this PR.

rldhont:
Conflict to resolve

PR #1731, opened 2020-07-24
Use QGIS expression in Lizmap edition

rldhont:
## Description

* Use QGIS Expression in Lizmap editor
* Linked to lizmap plugin [implementing lizmap server side](https://github.com/3liz/lizmap-plugin/pull/196)
* Funded by Conseil Départemental des Pyrénées-Atlantiques

## Steps

- [x] Evaluate QGIS default value expression
- [x] Drill down values
- [x] Constraints
- [x] Fields group display

PR #1775, opened 2020-08-24
[Feature] Filter by user on edition only

rldhont:
New feature, restrict filter by user on edition only, based on lizmap plugin config https://github.com/3liz/lizmap-plugin/pull/265

Funded by Conseil Départmental du Calvados


rldhont:
#1731 has to be updated when this will be merged

mdouchin (approved):

PR #1777, opened 2020-08-25
Filter - Encode parameters values for text & unique values

### LINKED ISSUE #1517: form filtering - fields with values containning the + character
the form filter does not show geometries for fields that contain `+` values, like `REN+RAN`
only the counter works in these cases

thanks for this very useful tool!

mdouchin:
## Description

* Fixes #1517


mdouchin:
Closing as duplicate of #1777 and it is better to fix it in release_3_3 then merge it in master

PR #1787, opened 2020-08-28
Digitizing + redlining. Handle multiple geometries

nboisteault:
Sponsored by [PNR Ballon des Vosges](https://www.parc-ballons-vosges.fr/).

Handle multiple geometries when digitizing and redlining


nboisteault:
@mdouchin @Gustry It should be good now!

mdouchin (approved):
LGTM !

PR #1791, opened 2020-08-28
Optimization : cache translate query

nboisteault:
translate query is slow because it reads translations line by line.
It can be cached taking care of future invalidation with version and lang as URL parameters

### DIFF
--- a/lizmap/modules/view/controllers/lizMap.classic.php
+++ b/lizmap/modules/view/controllers/lizMap.classic.php
@@ -128,7 +128,10 @@ class lizMapCtrl extends jController
 
         // the html response
         $rep = $this->getResponse('htmlmap');
-        $rep->addJSLink(jUrl::get('view~translate:index'));
+        // Get Lizmap version from project.xml
+        $xmlLoad = simplexml_load_file(jApp::appPath('project.xml'));
+        $version = (string) $xmlLoad->info->version;
+        $rep->addJSLink((jUrl::get('view~translate:index')).'?v='.$version.'&lang='.jApp::config()->locale);
 
         $this->repositoryKey = $lrep->getKey();
         $this->projectKey = $lproj->getKey();
--- a/lizmap/modules/view/controllers/translate.classic.php
+++ b/lizmap/modules/view/controllers/translate.classic.php
@@ -23,6 +23,7 @@ class translateCtrl extends jController
         $rep = $this->getResponse('binary');
         $rep->mimeType = 'text/javascript';
         $rep->doDownload = false;
+        $rep->setExpires('+1 year');
 
         $lang = $this->param('lang');
 


lizmap/modules/view/controllers/lizMap.classic.php
```diff
@@ -128,7 +128,10 @@ public function index()
 
         // the html response
         $rep = $this->getResponse('htmlmap');
-        $rep->addJSLink(jUrl::get('view~translate:index'));
+        // Get Lizmap version from project.xml
+        $xmlLoad = simplexml_load_file(jApp::appPath('project.xml'));
```
laurentj, line 132: I don't like it. opening an xml file is slooooww. we should find an other value....

lizmap/modules/view/controllers/translate.classic.php
```diff
@@ -23,6 +23,7 @@ public function index()
         $rep = $this->getResponse('binary');
         $rep->mimeType = 'text/javascript';
         $rep->doDownload = false;
+        $rep->setExpires('+1 year');
```
laurentj, line 26: on year is too much and useless. Few days max. Few hours is good imho.

laurentj (changes requested):

lizmap/modules/view/controllers/lizMap.classic.php
nboisteault: I don't know where else to find this value. It might be slow but it is slow only the first time.

lizmap/modules/view/controllers/translate.classic.php
nboisteault: We invalidate cache when needed so we can put a very long value. It is a [best practice](https://web.dev/uses-long-cache-ttl/) : 

> When possible, cache immutable static assets for a long time, such as a year or longer.

nboisteault:
@rldhont @mdouchin @dmarteau Your opinion?

rldhont:
@nboisteault why not used jCache ?

nboisteault:
> @nboisteault why not used jCache ?

Yes we could cache the translation server side with jCache but in my PR I avoid a request to the server as the translation is cached client side. We know this ressource will change only when version changes.

lizmap/modules/view/controllers/translate.classic.php
laurentj: >We invalidate cache

how?

nboisteault: As we change LWC version parameter in url when we update LWC, the new version of translation file is loaded.
```php
$version = (string) $xmlLoad->info->version;
$rep->addJSLink((jUrl::get('view~translate:index')).'?v='.$version.'&lang='.jLocale::getCurrentLang().'_'.jLocale::getCurrentCountry());`

nboisteault:
@rldhont @mdouchin @dmarteau @laurentj We need to conclude on this PR. I think it is an improvement and don't see problem, this might improved later if needed.

laurentj:
@nboisteault no problem to conclude, after you will fix the little change I asked about `&lang=`

laurentj (changes requested):

lizmap/modules/view/controllers/lizMap.classic.php
laurentj: ok

PR #1800, opened 2020-09-04
Lizmap\Project\Repository

alagroy-42:
Moving lizmapRepository to namespace Project.


alagroy-42:
Tests are failing because the composer.json is not well configured but it is on all the others Project classes tests PR so I'm not modifying it to avoid conflicts but it will be fine once everything is merged

laurentj:
Why having `$services` and `$context` on `getProject()`? It seems to me useless since we already give these parameters to the constructor.

laurentj (approved):

PR #1802, opened 2020-09-04
Remove wps module from docker image

dmarteau:
The [WPS module](https://github.com/3liz/lizmap-wps-web-client-module) is still a work in progresse and should not be delivered with docker official lizmap image  since it may be cause of instability.

This PR remove installation of the wps module in the docker image. 

The wps configuration is kept atm since this is a conditional setup and not activated if there is no LIZMAP_WPS_URL defined.


laurentj (approved):

Gustry:
Il ne faut pas oublier de supprimer la branche et encore mieux, d'utiliser un fork pour les branches temporaires.

PR #1810, opened 2020-09-09
fixing custom dates bug

alagroy-42:
fixing an issue reported by CD14. They couldn't insert their dates in the format they wanted.

[link to the support issue](http://support.3liz.com/issues/916/)

Issue #1076, opened 2019-02-18
date/time widget - customized format


josemvm:
hi

the LWC dowsn´t respect the custom widget display definet in qgis

![imagem](https://user-images.githubusercontent.com/10053874/52953283-f3c23e00-337e-11e9-9a23-678b6b5b86ee.png)

MM-dd-yyyy
![imagem](https://user-images.githubusercontent.com/10053874/52952967-11db6e80-337e-11e9-8ee4-2eb83a775509.png)

https://github.com/3liz/lizmap-web-client/issues/521
https://github.com/3liz/lizmap-web-client/commit/67ec2bfc59a5c4afe9dd2a1267f9e76961441221

thanks

nboisteault:
@rldhont is it a bug or an enhancement ? Did it work in past ?

rldhont:
The data is provided by WMS or WFS and I think QGIS SERVER 2.* does not respect it. I'm not sure.

I have fix it in QGIS Server 3.

rldhont:
@nboisteault can you check WMS GetFeatureInfo response and WFS GetFeature response for date fields.

nboisteault:
@rldhont With 2.18 and 3.4.5 date fields come with the same format. Custom format doesn't change anything. But when custom format is set we have to change the way we display data not the QGIS Server response I think.

rldhont:
@josemvm LWC does not support date/time custom widget display

rldhont:
For the Popup, GetFeatureInfo, the custom widget display has to be supported in QGIS Server.

For the edition, it's Lizmap Web Client enhancement but the default date-time widget is base on multi selector and not a simple line editor.

PR #1811, opened 2020-09-09
fixing custom dates bug

alagroy-42:
#1810 but on release3_3

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisForm.class.php
+++ b/lizmap/modules/lizmap/classes/qgisForm.class.php
@@ -462,7 +462,7 @@ class qgisForm implements qgisFormControlsInterface
     /**
      * Converts the datetime to the format specified in the qgis Project.
      *
-     * @param string $value The datetime to convert
+     * @param string $value    The datetime to convert
      * @param object $edittype The format of the date
      */
     public function convertDateTimeToFormat($value, $edittype)
@@ -470,12 +470,46 @@ class qgisForm implements qgisFormControlsInterface
         if (!property_exists($edittype, 'options') || !property_exists($edittype->options, 'field_format')) {
             return $value;
         }
-        $dateTab = explode('-', $value);
-        $value = $edittype->options->field_format;
-        $value = str_replace('yyyy', $dateTab[0], $value);
-        $value = str_replace('MM', $dateTab[1], $value);
+        $dateFormat = $edittype->options->field_format;
+        // conversion from QGIS to PHP format
+        $format = array_reverse(array(
+            'd' => 'j',
+            'dd' => 'd',
+            'ddd' => 'D',
+            'dddd' => 'l',
+            'M' => 'n',
+            'MM' => 'm',
+            'MMM' => 'M',
+            'MMMM' => 'F',
+            'yy' => 'y',
+            'yyyy' => 'Y',
+            'H' => 'G',
+            'HH' => 'H',
+            'h' => 'G',
+            'hh' => 'H',
+            'AP' => 'A',
+            'ap' => 'a',
+            'm' => 'i',
+            'mm' => 'i',
+            'ss' => 's',
+            't' => 'T',
+        ));
+        $format12h = array('a', 'ap', 'A', 'AP');
+        foreach ($format12h as $am) {
+            if (strstr($dateFormat, $am)) {
+                $format['h'] = 'g';
+                $format['hh'] = 'h';
 
-        return str_replace('dd', $dateTab[2], $value);
+                break;
+            }
+        }
+        foreach ($format as $qgis => $php) {
+            $dateFormat = str_replace($qgis, $php, $dateFormat);
+        }
+
+        $date = new DateTime($value);
+
+        return $date->format($dateFormat);
     }
 
     /**
@@ -585,7 +619,9 @@ class qgisForm implements qgisFormControlsInterface
                 continue;
             }
 
-            if ($this->formControls[$ref]->fieldEditType == 'DateTime') {
+            $convertDate = array('date', 'time', 'datetime');
+
+            if (in_array(strtolower($this->formControls[$ref]->fieldEditType), $convertDate)) {
                 $value = $this->convertDateTimeToFormat($value, $this->formControls[$ref]->getEditType());
             }
 
--- a/lizmap/modules/lizmap/classes/qgisForm.class.php
+++ b/lizmap/modules/lizmap/classes/qgisForm.class.php
@@ -452,6 +452,10 @@ class qgisForm implements qgisFormControlsInterface
                 }
                 $form->setData($ref.'_hidden', $value);
             } else {
+                if (in_array(strtolower($this->formControls[$ref]->fieldEditType), array('date', 'time', 'datetime'))) {
+                    $date = new DateTime($value);
+                    $value = $date->format('Y-m-d H:i:s');
+                }
                 $form->setData($ref, $value);
             }
         }
--- a/lizmap/modules/lizmap/classes/qgisForm.class.php
+++ b/lizmap/modules/lizmap/classes/qgisForm.class.php
@@ -453,7 +453,13 @@ class qgisForm implements qgisFormControlsInterface
                 $form->setData($ref.'_hidden', $value);
             } else {
                 if (in_array(strtolower($this->formControls[$ref]->fieldEditType), array('date', 'time', 'datetime'))) {
-                    $date = new DateTime($value);
+                    $edittype = $this->formControls[$ref]->getEditType();
+                    if (!$edittype || !property_exists($edittype, 'options')
+                        || !property_exists($edittype->options, 'field_format')) {
+                        break ;
+                    }
+                    $format = $this->convertQgisFormatToPHP($edittype->options->field_format);
+                    $date = DateTime::createFromFormat($format, $value);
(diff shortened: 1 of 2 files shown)


nboisteault:
@alagroy-42 I have tested and saving custom date is ok. Now I'd like to see the date being filled in the date picker when editing the feature. Thanks.

lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -459,6 +459,25 @@ public function setFormDataFromFields($feature)
...
+     *
+     * @param string $value The datetime to convert
+     * @param object $edittype The format of the date
+     */
+    public function convertDateTimeToFormat($value, $edittype)
+    {
+        if (!property_exists($edittype, 'options') || !property_exists($edittype->options, 'field_format')) {
+            return $value;
+        }
+        $dateTab = explode('-', $value);
```
nboisteault, line 473: You have the code handling this specific date format (yyyyMMdd) but I'd like it to be more generic and handle other formats like in this screenshot showing QGIS customizing date part.

image

nboisteault (changes requested):

Gustry:
@alagroy-42 Side note, can you also rename the commit with the actual bugfix ? You can skip the name of the customer I think. Thanks

josemvm:
https://github.com/3liz/lizmap-web-client/issues/1076

lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -459,6 +459,25 @@ public function setFormDataFromFields($feature)
...
+        $value = $edittype->options->field_format;
+        $value = str_replace('yyyy', $dateTab[0], $value);
+        $value = str_replace('MM', $dateTab[1], $value);
+
+        return str_replace('dd', $dateTab[2], $value);
```
rldhont, line 478:
```suggestion
        $date = new DateTime($value);
        return $date->format($edittype->options->field_format);
```
I think this is the easiest way to do. https://www.php.net/manual/fr/datetime.construct.php

rldhont (changes requested):
Is it possible to add a test for the method `convertDateTimeToFormat` ?

rldhont (approved):

lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -470,12 +470,44 @@ public function convertDateTimeToFormat($value, $edittype)
         if (!property_exists($edittype, 'options') || !property_exists($edittype->options, 'field_format')) {
             return $value;
         }
-        $dateTab = explode('-', $value);
-        $value = $edittype->options->field_format;
-        $value = str_replace('yyyy', $dateTab[0], $value);
-        $value = str_replace('MM', $dateTab[1], $value);
+        $dateFormat = $edittype->options->field_format;
+        // convertion from QGIS to PHP format
```
nboisteault, line 474: convertion => conversion

nboisteault:
@alagroy-42 
Tested **OK** : 

- yyyyMMdd
- ddMMyyyy
- yyyy.MM.dd
- ddd MMMM d yy
- yyyy-MM-dd HH:mm

Tested **NOK** : 
- h:m:s ap
In this case, we only get the datepicker without timepicker. '12:00:00 am' is saved though.

rldhont:
> Tested **NOK** :
> 
>     * `h:m:s ap`
>       In this case, we only get the datepicker without timepicker. '12:00:00 am' is saved though.
@nboisteault time control is not supported in Lizmap-web-client 3.3, so I think that the fix is ended.

lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -452,13 +452,70 @@ public function setFormDataFromFields($feature)
                 }
                 $form->setData($ref.'_hidden', $value);
             } else {
+                if (in_array(strtolower($this->formControls[$ref]->fieldEditType), array('date', 'time', 'datetime'))) {
+                    $date = new DateTime($value);
```
rldhont, line 456: $value could be in the `$edittype->options->field_format`

rldhont (changes requested):

lizmap/modules/lizmap/classes/qgisForm.class.php
alagroy-42: Here, `$value` is the date/time already stored in the DB that we want to put in the datepicker when we edit the feature and to do so, we need to convert it into the format accepted by Jelix.
[date](https://github.com/jelix/jelix/blob/8f313e8aa19f7a72bad7733a4f9354f5916b9ccb/lib/jelix/plugins/formwidget/date_html/date_html.formwidget.php#L79)
[datetimepicker](https://github.com/jelix/jelix/blob/8f313e8aa19f7a72bad7733a4f9354f5916b9ccb/lib/jelix/plugins/formwidget/datetime_html/datetime_html.formwidget.php#L89)

lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -452,13 +452,188 @@ public function setFormDataFromFields($feature)
...
+                if (in_array(strtolower($this->formControls[$ref]->fieldEditType), array('date', 'time', 'datetime'))) {
+                    $edittype = $this->formControls[$ref]->getEditType();
+                    if (!$edittype || !property_exists($edittype, 'options')
+                        || !property_exists($edittype->options, 'field_format')) {
+                        break ;
+                    }
+                    $format = $this->convertQgisFormatToPHP($edittype->options->field_format);
+                    $date = DateTime::createFromFormat($format, $value);
+                    $value = $date->format('Y-m-d H:i:s');
+                }
```
rldhont, line 464:
```suggestion
                if (in_array(strtolower($this->formControls[$ref]->fieldEditType), array('date', 'time', 'datetime'))) {
                    $edittype = $this->formControls[$ref]->getEditType();
                    if ($edittype && property_exists($edittype, 'options')
                        && property_exists($edittype->options, 'field_format')) {
                        $format = $this->convertQgisFormatToPHP($edittype->options->field_format);
                        $date = DateTime::createFromFormat($format, $value);
                        $value = $date->format('Y-m-d H:i:s');
                    }
                }
```
I was not sure that you can use a `break` here. @laurentj what do you think about it ?

lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -452,13 +452,188 @@ public function setFormDataFromFields($feature)
...
                 $form->setData($ref, $value);
             }
         }
 
         return $form;
     }
 
+    public function check($feature = null)
```
rldhont, line 472: Remove this method that comes from master and is not used in 3.3

rldhont (changes requested):

lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -452,13 +452,85 @@ public function setFormDataFromFields($feature)
...
+
+        return $dateFormat;
+    }
+
+    /**
+     * Converts the datetime to the format specified in the qgis Project.
+     *
+     * @param string $value       The datetime to convert
+     * @param object $edittype    The format of the date
+     * @param mixed  $fieldFormat
```
rldhont, line 523:
```suggestion
     * @param string $value           The datetime to convert
     * @param string $fieldFormat The format of the date
     * @return string
```
Update the docstring

lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -452,13 +452,85 @@ public function setFormDataFromFields($feature)
...
+            'AP' => 'A',
+            'ap' => 'a',
+            'm' => 'i',
+            'mm' => 'i',
+            'ss' => 's',
+            't' => 'T',
+        ));
+    }
+
+    public function convertQgisFormatToPHP($fieldFormat)
```
rldhont, line 497: Can you add a docstring ?

rldhont (changes requested):

lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -452,13 +452,85 @@ public function setFormDataFromFields($feature)
                 }
                 $form->setData($ref.'_hidden', $value);
             } else {
+                if (in_array(strtolower($this->formControls[$ref]->fieldEditType), array('date', 'time', 'datetime'))) {
+                    $edittype = $this->formControls[$ref]->getEditType();
+                    if ($edittype && property_exists($edittype, 'options')
+                            && property_exists($edittype->options, 'field_format')) {
+                        $format = $this->convertQgisFormatToPHP($edittype->options->field_format);
+                        $date = DateTime::createFromFormat($format, $value);
+                        $value = $date->format('Y-m-d H:i:s');
```
nboisteault, line 461: I have an error `Call to a member function format() on bool` when `$value` is `null`. Please fix.

nboisteault (changes requested):

rldhont (approved):

nboisteault (approved):
Thanks!

josemvm:
before:
imagem


after:
imagem -- must be --> 2020-10-02

@alagroy-42 thanks for your great work on LWC

PR #1814, opened 2020-09-14
Project tests

alagroy-42:
Tests for the Project class


lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -349,8 +349,9 @@ protected function readProject($key, Repository $rep)
 
     public function getQgisPath()
     {
-        if (!$this->file) {
-            $this->file = realpath($this->repository->getPath()).'/'.$this->key.'.qgs';
+        $path = $this->repository->getPath();
+        if (!$this->file && $path != '' && $path != false) {
+            $this->file = realpath($path).'/'.$this->key.'.qgs';
```
laurentj, line 354: the `realpath()` should be done into the Repository class, so the code calling `getPath()` don't have to do it

lizmap/modules/lizmap/lib/Project/ProjectCache.php
```diff
@@ -104,6 +104,7 @@ public function retrieveProjectData()
      */
     public function storeProjectData($data)
     {
+        file_put_contents(__DIR__.'/cache.log', json_encode($data, JSON_PRETTY_PRINT));
```
laurentj, line 107: debug stuff? if yes, it should be removed

tests/units/classes/Project/qgisProjectForTests.php
```diff
@@ -4,8 +4,11 @@
 
 class qgisProjectForTests extends QgisProject
 {
-    public function __construct()
+    public function __construct($original = false, $data = array())
     {
+        if ($original) {
+            parent::__construct(__FILE__, new lizmapServices(null, null, false, '', ''), $data);
```
laurentj, line 10: why `__FILE__` ? the expected parameter is the name of a qgis file, not the filename of the class. Use a fake name like 'fake.qgs', it is less disturbing.

Instead of having this `$original` parameter, just have `$data = null`, and if `$data` is not null, call the parent constructor

laurentj (changes requested):

laurentj (approved):

PR #1821, opened 2020-09-16
Map menu UI more mobile friendly (hamburger menu)

nboisteault:


nboisteault:
@mdouchin We close dock and subdock in mobile context when zooming or changing opacity in sub-dock :  

- https://github.com/3liz/lizmap-web-client/blob/master/lizmap/www/assets/js/switcher-layers-actions.js#L406
- https://github.com/3liz/lizmap-web-client/blob/master/lizmap/www/assets/js/switcher-layers-actions.js#L481

I find this behavior annoying. I think we could maybe close sub-dock but not dock and even better close nothing imo. Thoughts?

nboisteault:
@Gustry feel free to test this new UI ;)

mdouchin:
Thanks for your work. A quick question : have you tested to run the lizMap.addDock() public method and check if the new item is correctly added in the mobile dock new interface ?

nboisteault:
@mdouchin yes atlas is correctly added in Montpellier project.

rldhont (approved):

PR #1816, opened 2020-09-15
Feature : element to go to given coordinates

nboisteault:
Funded by [DEAL Mayotte](http://www.mayotte.gouv.fr/Services-de-l-Etat/Direction-de-l-Environnement-de-l-Amenagement-et-du-Logement-DEAL).

Custom element to let user zoom or center to given latitude/longitude.

Fix #1232 
Fix #268 

go-to-coordinates

PR #1827, opened 2020-09-21
Feature : component to display mouse position or center to given coordinates

### LINKED ISSUE #268: zoom to coordinates in the search box
i supose that's possible to search with Google (API) by geographic coordinate system... but not (yet)...

### LINKED ISSUE #1232: Search by coordinates with OpenStreetMap
Hi everyone,

is it possible to set the search by address with the "internal" service of OpenStreetMap and not Nominatim?
I need to set the search by coordinates.

Thank you all.
Giorgio

nboisteault:
Funded by [DEAL Mayotte](http://www.mayotte.gouv.fr/Services-de-l-Etat/Direction-de-l-Environnement-de-l-Amenagement-et-du-Logement-DEAL).

Custom element displaying mouse position or let user center to given latitude/longitude in different projections. (Like in QGIS)

Fix #1232 
Fix #268


rldhont (approved):

nboisteault:
@mdouchin : how to handle 'degree minute' unit with edition as it add some symbols? I would say edition is enabled only with other 'simple' units.

image

mdouchin:
I agree: no need to let the user enter `43°34'67''` -> the user will only be able to use simple real numbers

nboisteault:
@rldhont @josemvm : concerning [this comment](https://github.com/3liz/lizmap-web-client/pull/1816#issuecomment-694826430), this button is only available if address search is active. We then need something else.

josemvm:
@nboisteault is it possible for you (or something like this)? 

imagem

thanks for your work!

nboisteault:
@mdouchin @rldhont Actually when mouse is out of the map we display a text instead of last coordinates. Should i keep this behavior? I'd prefer to keep that coordinates as it can be handy for users to write their own coordinates based on those.

nboisteault:
@mdouchin @rldhont It is ready. Please test exhaustively with different projections.

rldhont (approved):

rldhont:
new approval

mdouchin (approved):

josemvm:
@nboisteault 
>component to display mouse position

does this mean that the user can also read coordinates on different projection systems?

PR #1838, opened 2020-09-24
CFG - Make group_visibility an array instead of a string

Gustry:
## Description

* Ticket : #https://github.com/3liz/lizmap-plugin/issues/292
* CFG - Make group_visibility an array instead of a string
* Linked to https://github.com/3liz/lizmap-plugin/pull/293


rldhont (approved):
Good to me

PR #1840, opened 2020-09-28
QgisForm* namespace + Jelix uncontextualization

alagroy-42:
Moving qgisForm * classes to namespace `Lizmap\Form` + changing some Jelix methods in appContext methods.


lizmap/modules/lizmap/lib/Form/QgisForm.php
```diff
@@ -410,15 +419,15 @@ public function getAttributesEditorForm()
     }
 
     /**
-     * @return jFormsPlugin[]
+     * @return \jFormsPlugin[]
```
laurentj, line 422: jFormsPlugin does not exist

laurentj (approved):

Issue #1468, opened 2020-01-20
Form Filter: do not display the form or input if not well configure


rldhont:
### What is the bug?
If the form filter is not well configured, for example field and min_date have no values, the form and input is display.

Linked to https://github.com/3liz/lizmap-plugin/issues/194

### Steps to reproduce the issue

1. Go to lizmap-plugin
2. Add a form filter with field empty
3. Update on the server
4. Got to the web page
5. Open Form Filter
6. See the not well configured form

### Expected behavior
Do not display the input or the form if theyr are not well configured

### Environment

 - Lizmap version: 3.3.*
 - QGIS Server version: 3.4.*
 - QGIS Project version: 3.4.*
 - OS (Windows, Linux, MacOS, Android…): All
 - Browser (Firefox, Chrome…): All

rldhont:
Check the `title` when getting the `field_item` 
* https://github.com/3liz/lizmap-web-client/blob/release_3_3/lizmap/www/js/filter.js#L132 mandatory
* https://github.com/3liz/lizmap-web-client/blob/release_3_3/lizmap/www/js/filter.js#L140
* https://github.com/3liz/lizmap-web-client/blob/release_3_3/lizmap/www/js/filter.js#L660 mandatory
* https://github.com/3liz/lizmap-web-client/blob/release_3_3/lizmap/www/js/filter.js#L761 mandatory
* https://github.com/3liz/lizmap-web-client/blob/release_3_3/lizmap/www/js/filter.js#L795
* https://github.com/3liz/lizmap-web-client/blob/release_3_3/lizmap/www/js/filter.js#L874

PR #1854, opened 2020-10-06
Don't display filter when no data

nboisteault:
## Description
This PR should handle filter UI when there is no data
* Ticket : #1468
*


mdouchin:
I will test asap

Issue #1717, opened 2020-07-09
Improve user admin list


arno974:
It is not directly a feature in Lizmap, but rather in Jelix (Admin module). In its current configuration, the user selection form has only a limited interest. Indeed, it requires to know the exact user pseudo. It would be relevant to filter the list instead of having to find the exact user.

![image](https://user-images.githubusercontent.com/425243/87002828-801b0e00-c1cb-11ea-9938-f742e7db549c.png)

It would certainly be relevant to propose an improvement of this one is :

- that the input of the form (j_user_login) is an auto-incrementing list and that when you type the user's name, it proposes a list of choices.
- or that the table listing the users be filtered according to the parameter entered (ideally an ILIKE with a % at the beginning and end of the pseudo to have a large match)

PR #1865, opened 2020-10-08
adding autocompletion for jAuthdb

alagroy-42:
## Description

*  Ticket : #1717 
*  There is now an autocompletion on the input form to view user infos of the user management page.

### DIFF
--- a/lib/jelix-admin-modules/jauthdb_admin/controllers/default.classic.php
+++ b/lib/jelix-admin-modules/jauthdb_admin/controllers/default.classic.php
@@ -1,82 +1,89 @@
 <?php
 /**
-* @package   admin
-* @subpackage jauthdb_admin
-* @author    Laurent Jouanneau
-* @copyright 2009-2019 Laurent Jouanneau
-* @link      http://jelix.org
-* @license   http://www.gnu.org/licenses/old-licenses/gpl-2.0.html GNU Public Licence
-*/
+ * @author    Laurent Jouanneau
+ * @copyright 2009-2019 Laurent Jouanneau
+ *
+ * @see      http://jelix.org
+ *
+ * @license   http://www.gnu.org/licenses/old-licenses/gpl-2.0.html GNU Public Licence
+ */
 
 /**
- * controller to manage all users
+ * controller to manage all users.
  */
-class defaultCtrl extends jController {
-
+class defaultCtrl extends jController
+{
     public $sensitiveParameters = array('password', 'password_confirm', 'pwd', 'pwd_confirm');
 
-    public $pluginParams=array(
-        'index'        =>array('jacl2.right'=>'auth.users.list'),
-        'view'         =>array('jacl2.right'=>'auth.users.view'),
-        'precreate'    =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.create')),
-        'create'       =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.create')),
-        'savecreate'   =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.create')),
-        'preupdate'    =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.modify')),
-        'editupdate'   =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.modify')),
-        'saveupdate'   =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.modify')),
-        'deleteconfirm'=>array('jacl2.rights.and'=>array('auth.users.view','auth.users.delete')),
-        'delete'       =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.delete')),
+    public $pluginParams = array(
+        'index'         => array('jacl2.right' => 'auth.users.list'),
+        'view'          => array('jacl2.right' => 'auth.users.view'),
+        'precreate'     => array('jacl2.rights.and' => array('auth.users.view', 'auth.users.create')),
+        'create'        => array('jacl2.rights.and' => array('auth.users.view', 'auth.users.create')),
+        'savecreate'    => array('jacl2.rights.and' => array('auth.users.view', 'auth.users.create')),
+        'preupdate'     => array('jacl2.rights.and' => array('auth.users.view', 'auth.users.modify')),
+        'editupdate'    => array('jacl2.rights.and' => array('auth.users.view', 'auth.users.modify')),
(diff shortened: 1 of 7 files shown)


rldhont (approved):

lib/jelix-admin-modules/jauthdb_admin/controllers/default.classic.php
```diff
@@ -1,82 +1,89 @@
...
-        'precreate'    =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.create')),
-        'create'       =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.create')),
-        'savecreate'   =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.create')),
-        'preupdate'    =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.modify')),
-        'editupdate'   =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.modify')),
-        'saveupdate'   =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.modify')),
-        'deleteconfirm'=>array('jacl2.rights.and'=>array('auth.users.view','auth.users.delete')),
-        'delete'       =>array('jacl2.rights.and'=>array('auth.users.view','auth.users.delete')),
+    public $pluginParams = array(
+        'index' => array('jacl2.right' => 'auth.users.list'),
```
laurentj, line 19: php-cs-fixer is not good. it is less readable than the previous code. Is there a configuation parameter we can set to avoid this change?

lib/jelix-admin-modules/jauthdb_admin/controllers/default.classic.php
```diff
@@ -413,26 +448,44 @@ function delete(){
...
             $rep->action = 'default:index';
         }
+
         return $rep;
     }
 
-}
+    public function autocomplete()
+    {
+        $rep = $this->getResponse('text');
```
laurentj, line 474: There is a JSON response object, it's better because it encodes content for you, and return the right mime type. so, indicate 'json'. See jResponseJson to know about its API.

lib/jelix-admin-modules/jauthdb_admin/controllers/default.classic.php
```diff
@@ -413,26 +448,44 @@ function delete(){
...
+
+        $dao = jDao::get($this->dao, $this->dbProfile);
+
+        $cond = jDao::createConditions();
+        $cond->addItemOrder('login', 'asc');
+        $list = $dao->findBy($cond);
+        $users = array();
+        foreach ($list as $prop) {
+            if (strstr($prop->login, $this->param('term')) || $this->param('term') === '') {
```
laurentj, line 483: if the term is empty or if its length is lower than 2 characters for exemple, you should return an empty JSON array, else it could returns hundred logins, this is not what we want.

Chek the content of term before using the dao

lib/jelix-admin-modules/jauthdb_admin/templates/crud_list.tpl
```diff
@@ -1,11 +1,12 @@
+{meta_html js $j_jelixwww.'js/authdb_admin.js'}
 <h1>{@jauthdb_admin~crud.title.list@}</h1>
 
 {if $canview}
 <form action="{formurl 'jauthdb_admin~default:view'}" method="get">
     <div>
         <label for="search-login">{@jauthdb_admin~crud.title.view@}</label>
         <input id="search-login" name="j_user_login">
-        <button type="submit">{@jauthdb_admin~crud.link.view.record@}</button>
+        <button type="submit" onsubmit="test();">{@jauthdb_admin~crud.link.view.record@}</button>
```
laurentj, line 9: what is this `test()`? ;-)

lib/jelix-www/js/authdb_admin.js
```diff
@@ -0,0 +1,6 @@
+$("document").ready( function () {
+    $("#search-login").autocomplete({
+        source: window.location.href+'/default/autocomplete',
```
laurentj, line 3: no, you must not set url your self in JS code, because this url could change, depending on how the application is configured.

You could store the url into the template, with the `{jurl}` template function, into a data attribute of the search login field for example, and retrieving its value here to give it to autocomplete()

lib/jelix-www/js/authdb_admin.js
```diff
@@ -0,0 +1,6 @@
...
+        minLength: 0
```
laurentj, line 4: no, not 0. we don't want to retrieve all login. Let's say 2 characters?

lizmap/www/assets/jelix/js/authdb_admin.js
```diff
@@ -0,0 +1,6 @@
+$("document").ready( function () {
+    $("#search-login").autocomplete({
+        source: window.location.href+'/default/autocomplete',
```
laurentj, line 3: same comments at the copy of this file

laurentj (changes requested):

lib/jelix-admin-modules/jauthdb_admin/controllers/default.classic.php
alagroy-42: There is but we should do a whole PR to fix it in every file otherwise the lint will fail everytime.

alagroy-42:
The lint issue is due to the modification of `.php_cs.dist`, if you approve the modification, I'll do a PR with the linting of all the files.

lib/jelix-admin-modules/jauthdb_admin/templates/crud_list.tpl
```diff
@@ -5,8 +5,8 @@
 <form action="{formurl 'jauthdb_admin~default:view'}" method="get">
     <div>
         <label for="search-login">{@jauthdb_admin~crud.title.view@}</label>
-        <input id="search-login" name="j_user_login">
-        <button type="submit" onsubmit="test();">{@jauthdb_admin~crud.link.view.record@}</button>
+        <input id="search-login" name="j_user_login" link_autocomplete="{jurl 'jauthdb_admin~default:autocomplete'}">
```
laurentj, line 8: You should use a 'data' like attribute, like `data-link-autocomplete`. See [HTML doc about them](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*) to read the corresponding JS API to retrieve `data-*` attributes (or the jquery API). Custom attributes that are not `data-*` attributes are not allowed by the HTML5 specification.

laurentj (changes requested):

laurentj (approved):

PR #1868, opened 2020-10-12
JSON Generation tests

alagroy-42:
Tests for `QgisProject` JSON generation methods for forms.
Correcting information extraction from qgis files with `edittypes` tags.

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/QgisProject.php
+++ b/lizmap/modules/lizmap/lib/Project/QgisProject.php
@@ -1475,7 +1475,7 @@ class QgisProject
         return $tab;
     }
 
-    protected function getEditType($layerXml, $edittypes)
+    protected function getEditType($edittypes)
     {
         $editTab = array();
 
@@ -1587,7 +1587,7 @@ class QgisProject
         // <maplayer><map-layer-style-manager><map-layer-style><qgis/></map-layer-style></map-layer-style-manager></maplayer>
         $edittypes = $layerXml->xpath('.//edittypes');
         if ($edittypes && count($edittypes)) {
-            $props = $this->getEditType($layerXml, $edittypes[0]);
+            $props = $this->getEditType($edittypes[0]);
         } elseif ($layerXml->fieldConfiguration && count($layerXml->fieldConfiguration)) {
             $props = $this->getFieldConfiguration($layerXml);
         } else {
--- a/lizmap/modules/lizmap/lib/Project/QgisProject.php
+++ b/lizmap/modules/lizmap/lib/Project/QgisProject.php
@@ -1435,7 +1435,7 @@ class QgisProject
 
                 // Option with string list of values
                 } elseif ((string) $option->attributes()->type === 'StringList') {
-                    $fieldEditOptions[(string) $option->attributes()->name] = ${$this}->getValuesFromOptions($option, false);
+                    $fieldEditOptions[(string) $option->attributes()->name] = $this->getValuesFromOptions($option, false);
                 // Simple option
                 } else {
                     $fieldEditOptions[(string) $option->attributes()->name] = (string) $option->attributes()->value;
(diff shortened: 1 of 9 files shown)


lizmap/modules/lizmap/lib/Project/QgisProject.php
```diff
@@ -1464,25 +1464,43 @@ protected function getFieldConfiguration($layerXml)
...
-                $editTab[$attributes->name]['widgetv2configAttr'] = $edittype->widgetv2config->attributes();
-                $editTab[$attributes->name]['fieldEditType'] = (string) $edittype->attributes()->widgetv2type;
+            $name = (string) $attributes->name;
+            $type = (string) $attributes->widgetv2type;
+            $editTab[$name] = array();
+            // foreach ($edittype as $key => $value) {
+            //     $editTab[$name]['edittype'][$key] = (object)$this->getTabFromAttributes($edittype->$key->attributes());
+            // }
+            $editTab[$name]['edittype'] = (object) $this->getTabFromAttributes($edittype->attributes());
+            $editTab[$name]['edittype']->widgetv2config = (object) $this->getTabFromAttributes($edittype->widgetv2config->attributes());
```
laurentj, line 1493: why duplicate this content? Since you put the same value into widgetv2configAttr below, you should remove this line

lizmap/modules/lizmap/lib/Project/QgisProject.php
```diff
@@ -1464,25 +1464,43 @@ protected function getFieldConfiguration($layerXml)
...
+            $tab[(string) $key] = (string) $value;
+        }
+
+        return $tab;
+    }
+
     protected function getEditType($layerXml)
     {
-        $edittypes = $layerXml->edittypes;
+        $edittypes = $layerXml->xpath('.//edittypes');
```
laurentj, line 1480: add a comment:

```
// <edittypes> can be a child of <maplayer>, but can be also elsewhere inside a <maplayer><map-layer-style-manager><map-layer-style><qgis/></map-layer-style></map-layer-style-manager></maplayer>
```

lizmap/modules/lizmap/lib/Project/QgisProject.php
```diff
@@ -1577,6 +1595,14 @@ public function readFormControls($layerXml, $layerId, $proj)
         } else {
             $data = null;
         }
+        $edittypes = $layerXml->xpath('.//edittypes');
```
laurentj, line 1598: you did already these if statements few line below.

laurentj (changes requested):

lizmap/modules/lizmap/lib/Project/QgisProject.php
laurentj: `getEditType()` is called after already retrieving the edittype elements in `readFormControls()`. So, to avoid to call twice `xpath` for the same thing, `getEditType` should be called with the `<edittypes>` element, no the `<maplayer>` element.

laurentj: so, add the comment into `readFormControls()` instead.

lizmap/modules/lizmap/lib/Project/QgisProject.php
alagroy-42: It's because they are two different properties in qgisFormControl and it's easier to loop on the Json object with each property being a different element.

lizmap/modules/lizmap/lib/Project/QgisProject.php
```diff
@@ -1464,25 +1464,39 @@ protected function getFieldConfiguration($layerXml)
...
+        $tab = array();
+
+        foreach ($attributes as $key => $value) {
+            $tab[(string) $key] = (string) $value;
+        }
+
+        return $tab;
+    }
+
+    protected function getEditType($layerXml, $edittypes)
```
laurentj, line 1478: $layerXml parameter is useless

laurentj (changes requested):
last change to make, and it will be ok

laurentj:
It seems your last change breaks unit tests. Please fix tests.

PR #1869, opened 2020-10-13
[Feature] Filter by user on edition only - use in edition

rldhont:
## Description

Last part of the new feature, restrict filter by user on edition only https://github.com/3liz/lizmap-web-client/pull/1775, based on lizmap plugin config 3liz/lizmap-plugin#265

Using the filter by login in edition.

Funded by Conseil Départmental du Calvados

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisExpressionUtils.class.php
+++ b/lizmap/modules/lizmap/classes/qgisExpressionUtils.class.php
@@ -79,38 +79,72 @@ class qgisExpressionUtils
         return preg_match('/\\B@current_geometry\\b/', $exp) === 1;
     }
 
-    public static function updateExpressionByUser($layer, $expression)
+    /**
+     * Returns the expression used to filter by login the layer.
+     *
+     * @param qgisVectorLayer $layer   A QGIS vector layer
+     * @param bool            $edition It's for editon
+     *
+     * @return string the expression to filter by login the layer
+     */
+    public static function getExpressionByUser($layer, $edition = false)
     {
         $project = $layer->getProject();
         $repository = $project->getRepository();
         // No filter data by login rights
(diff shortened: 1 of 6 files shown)


nboisteault:
@rldhont can you give a scenario to test this feature?

rldhont:
> @rldhont can you give a scenario to test this feature?

* Create QGIS project with a PostGIS layer
* Config the project for QGIS server
* Published the PostGIS layer in WFS
* In lizmap plugin, define the PostGIS Layer as editable
* In lizmap plugin, define a filter by login for the PostGIS Layer with the option only for edition https://github.com/3liz/lizmap-plugin/pull/265
* Published the project to lizmap
* Try to edit features with different users in different groups

lizmap/modules/lizmap/classes/qgisExpressionUtils.class.php
```diff
@@ -89,28 +97,54 @@ public static function updateExpressionByUser($layer, $expression)
         }
```
laurentj, line 97: on the line above, it returns $expression, which does not exist anymore

laurentj (changes requested):

nboisteault:
@rldhont is it ready for tests?

nboisteault:
@rldhont I can't edit this feature while I'm connected as admin user.
Also edition and delete buttons are visible even if user doesn't have rights to do so. I think, this might handled client side.
image

rldhont:
@nboisteault do you have some logs ?

lizmap/modules/lizmap/classes/qgisExpressionUtils.class.php
rldhont: fixed

nboisteault (approved):
I have tested. Behavior is ok.

PR #1876, opened 2020-10-19
[Bugfix] Use form check in Ajax Save new feature

rldhont:
In the Ajax saveNewFeature method, the form check was not used to return before saveToDb.

### DIFF
--- a/lizmap/modules/lizmap/controllers/edition.classic.php
+++ b/lizmap/modules/lizmap/controllers/edition.classic.php
@@ -973,19 +973,21 @@ class editionCtrl extends jController
 
         // Check the form data and redirect if needed
         $check = $form->check();
-        $modifyGeometry = $this->layer->getEditionCapabilities()->capabilities->modifyGeometry;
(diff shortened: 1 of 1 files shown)


rldhont:
To simplify the method `saveNewFeature` and avoid duplicated codes, we can use `$qgisForm->check()` instead of `$form->check()`, isn't it @laurentj ?

nboisteault:
@rldhont is it just to be reviewed or also tested? If it needs tests, what is the scenario?

rldhont:
I am waiting for @laurentj comments. I would like to know if I can do more changes.

@laurentj has develop this method, he can get you a scenario to test it.

laurentj (approved):

laurentj:
> To simplify the method `saveNewFeature` and avoid duplicated codes, we can use `$qgisForm->check()` instead of `$form->check()`, isn't it @laurentj ?

Probably. I don't know. but we have to be care about content we returned. For example `qgisForm::check()` stores error messages into the form, se we would have to retrieve them in order to return them as json content, as the actuel code.

laurentj:
at least, the use of `$qgisForm->check()` should be an other PR

rldhont:
@laurentj thanks for review, I'll do a new PR.

PR #1878, opened 2020-10-19
[Bugfix] Edition: On insert failed keep the geometry

rldhont:
## Description

If the form is reopned because of a failed server side, the geometry was destroyed.

Funded by Conseil Département des Pyrénées Atlantiques


nboisteault (approved):
Tested changing https://github.com/3liz/lizmap-web-client/blob/release_3_3/lizmap/modules/lizmap/controllers/edition.classic.php#L611 with `$check = false`

PR #1883, opened 2020-10-21
[Feature] Get editable features based on login filter

rldhont:
This is the last feature to use **Filter by user on edition only** https://github.com/3liz/lizmap-plugin/pull/265 after https://github.com/3liz/lizmap-web-client/pull/1775 and https://github.com/3liz/lizmap-web-client/pull/1869

The idea is to get the editable features to provide edition buttons to the right features.

Funded by Conseil Départmental du Calvados https://www.calvados.fr


nboisteault (approved):

rldhont (commented):
I approve

alagroy-42 (approved):

PR #1892, opened 2020-10-27
[tests] Fix postgresql access by using a pg service file

laurentj:
Credentials to access to postgresql from the host machine
are different from credentials (hostname + port) to access
to postgresql from a container. So we must use a
postgresql service in Qgis project.


tests/README.md
```diff
@@ -63,6 +54,51 @@ Available commands:
...
+accessing from one of the containers, like qgis or lizmap.
+
+You must use the postgresql service `lizmapdb` as access parameter in your qgis projects. 
+Create the file `~/.pg_service.conf` and put these parameters in it:
+
+```
+[lizmapdb]
+host=localhost
+port=8132
+database=lizmap
```
alagroy-42, line 80:
```suggestion
```
QGIS won't connect to pg if there is the database field.

tests/docker-conf/pg_service.conf
```diff
@@ -0,0 +1,7 @@
+[lizmapdb]
+host=pgsql
+port=5432
+database=lizmap
```
alagroy-42, line 4: Same as above

alagroy-42 (changes requested):
Even with these changes, connection doesn't work when initializing a form in LWC.  `pg_connect` fails in `lib/jelix/plugin/db/pgsql/pgsql.dbconnection.php`, but only when it takes a service as a param, when i set the full data by hand, there is no issue to connect and I get the form with the correct fields.

laurentj:
Thank you for your review. I fixed the pg service file, and updated the README.md file.

alagroy-42 (changes requested):
It still doesn't work in LWC, the project loads and LWC gets the db field correctly but it doesn't display the existing features and I get a DB connection error when I want to add a feature to the layer.

PR #1893, opened 2020-10-28
Fix: Track fields changes for creation

rldhont:
Add a method to reset form controls data before tracking form controls changes for creation feature.

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisForm.class.php
+++ b/lizmap/modules/lizmap/classes/qgisForm.class.php
@@ -384,7 +384,7 @@ class qgisForm implements qgisFormControlsInterface
     /**
      * Reset the form controls data to Null.
      *
-     * @return object the Jelix jForm object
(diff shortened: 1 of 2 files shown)


lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -381,6 +381,25 @@ public function getForm()
         return $this->form;
     }
 
+    /**
+     * Reset the form controls data to Null.
+     *
+     * @return object the Jelix jForm object
```
laurentj, line 387: no, this is not a "jForm" object. `object` is too vague. You should put the real class name, `jFormsBase`

rldhont: Done

rldhont:
@nboisteault can you test it ?

nboisteault (approved):
Tested OK!

rldhont:
Thanks @nboisteault

PR #1894, opened 2020-10-28
Edition, manage children: Do not activate combobox or autocomplete on disabled select

rldhont:
* Do not activate combobox or autocomplete on disabled select
* Fix widget relation value for tram_stop_id in tram_stop_works


nboisteault (approved):
Tested. OK !

PR #1895, opened 2020-10-28
Fix edition manage children value

rldhont:
Fixing an issue introduced with https://github.com/3liz/lizmap-web-client/commit/8053456875252733598a5c5387487fb20cd3421c

It is also a manually merge of https://github.com/3liz/lizmap-web-client/pull/1894 with: 
* Do not activate combobox or autocomplete on disabled select
* Fix widget relation value for tram_stop_id in tram_stop_works

### DIFF
--- a/lizmap/www/assets/js/edition.js
+++ b/lizmap/www/assets/js/edition.js
@@ -1377,33 +1377,38 @@ OpenLayers.Geometry.pointOnSegment = function(point, segment) {
             // Manage child form
             if ( editionLayer['parent'] != null ){
                 var parentInfo = editionLayer['parent'];
-                var relationRefField = parentInfo['relation'].referencingField;
-                var parentFeatProp = parentInfo['feature'].properties[relationRefField];
+                var relation = parentInfo['relation'];
+                var relationRefField = relation.referencingField;
+                var parentFeatProp = parentInfo['feature'].properties[relation.referencedField];
 
                 var select = form.find('select[name="'+relationRefField+'"]');
                 if( select.length == 1 ){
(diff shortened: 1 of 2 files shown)


lizmap/www/assets/js/edition.js
```diff
@@ -1377,33 +1377,38 @@ OpenLayers.Geometry.pointOnSegment = function(point, segment) {
             // Manage child form
             if ( editionLayer['parent'] != null ){
                 var parentInfo = editionLayer['parent'];
-                var relationRefField = parentInfo['relation'].referencingField;
-                var parentFeatProp = parentInfo['feature'].properties[relationRefField];
+                var relation = parentInfo['relation'];
+                var relationRefField = relation.referencingField;
+                var parentFeatProp = parentInfo['feature'].properties[relation.referencedField];
```
laurentj, line 1382: sorry, I don't see why you made this change. What you write do exactly the same thing that the old code, with more code. don't you? Did I miss something?

rldhont: You have probably missed the difference between `referencingField` and `referencedField`.

* `referencingField` is the field's name in the child layer
* `referencedField` is the field's name in the parent layer

Here the `referencingField` is used to get the form control and the `referencedField` to get the value to link child to the parent.

nboisteault (approved):
Tested. OK ! Thanks.

lizmap/www/assets/js/edition.js
rldhont: @laurentj Do you approve my changes ?

rldhont:
@nboisteault thanks

PR #1903, opened 2020-11-03
Use new config param 'allow_without_geom'

### LINKED ISSUE #1665: Editing: now it is allowed to add a record without geom
* add a PostGIS layer to a project
* make it editable
* Lizmap llows to add a record without a geometry, without warning.
AFAIR previous versions issued a warning in this case, asking the user to add a geometry.
I think the current behaviour is less desirable, as it leads to useless entries.
LM 3.3.6, QGIS 3.10, pg 11.

nboisteault:
ref https://github.com/3liz/lizmap-plugin/pull/302
Fix #1665

### DIFF
--- a/lizmap/modules/lizmap/controllers/edition.classic.php
+++ b/lizmap/modules/lizmap/controllers/edition.classic.php
@@ -618,8 +618,8 @@ class editionCtrl extends jController
 
         // Check the form data and redirect if needed
         $check = $form->check();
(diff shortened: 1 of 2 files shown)


mdouchin (approved):

rldhont (approved):

rldhont:
@nboisteault you can merge

PR #1911, opened 2020-11-10
[tests] Install Lizmap as a QGIS Server plugin

### LINKED ISSUE #1910: QGIS Server plugins in docker for tests
We need a way to install QGIS Server plugins in docker for tests. Lizmap as a QGIS server plugin is needed to test drill down lists in edition for example.

- To have a volume declared to map a directory containing QGIS Server plugins to another in docker.
- Set `QGIS_PLUGINPATH` accordingly
- Install Lizmap as a QGIS server plugin by default
- Test everything is ok with `SERVICE=LIZMAP&REQUEST=GETSERVERSETTINGS` as a query string (https://github.com/3liz/lizmap-plugin#lizmap-server-api)
- profit? :)

nboisteault:
Fix #1910

### DIFF
--- a/tests/run-docker
+++ b/tests/run-docker
@@ -8,7 +8,16 @@ export LIZMAP_GROUP_ID=${LIZMAP_GROUP_ID:-$(id -g)}
 CMD=$1
 
 if [ "$CMD" == "" ]; then
-    CMD="up -d"
+  CMD="up -d"
+fi
+
+# Install or update Lizmap as a QGIS Server plugin
+if [ "$CMD" != "stop" ]; then
+  curl "https://packages.3liz.org/pub/lizmap-qgis-plugin/dev/lizmap-qgis-plugin.dev.zip" --output lizmap-qgis-plugin.dev.zip --silent
+  unzip -qq lizmap-qgis-plugin.dev.zip
+  rm lizmap-qgis-plugin.dev.zip
+  rm -rf qgis-server-plugins/lizmap/
+  mv lizmap qgis-server-plugins/
 fi
 
 if [ "$LZMBRANCH" == "" ]; then


tests/run-docker
```diff
@@ -8,7 +8,16 @@ export LIZMAP_GROUP_ID=${LIZMAP_GROUP_ID:-$(id -g)}
 CMD=$1
 
 if [ "$CMD" == "" ]; then
-    CMD="up -d"
+  CMD="up -d"
+fi
+
+# Install or update Lizmap as a QGIS Server plugin
+if [ "$CMD" != "stop" ]; then
```
laurentj, line 15: you should verify first that the directory qgis-server-plugins/lizmap/ does not already exist, else each time we launch the stack, it will have a download. This is not really that we want IMHO

laurentj: You could also do this statement when `$CMD="build"`.

`if [ !- d qgis-server-plugins/lizmap/  -o "$CMD" == "build" ]; then`

laurentj (changes requested):

tests/run-docker
nboisteault: I did that to always have the last Lizmap plugin version in a simple manner.

laurentj: well, if you want the latest version, you rebuild, and that's all. but you should not install the plugin each time we up the stack

laurentj (changes requested):

Gustry:
In the meantime, I have included in 3.4 a very small script helper to install plugins.
You can simply remove it when you want with this PR.

(Atlasprint, WfsOutputExtension should be added).

dmarteau:
QGSRV_SERVER_PLUGINPATH must be declared  in environnement in `docker-run`

Gustry:
I think we can close this one : https://github.com/3liz/lizmap-web-client/commit/7f33744a872d933b229af0c4dc785ec42bdd9e2c#diff-59830ebc3a4184110566bf1a290d08473dfdcbd492ce498b14cd1a5e2fa2e441 from @laurentj 
and close #1910 ?

PR #1916, opened 2020-11-12
Add 'filter layer by user' manual tests

nboisteault:
@laurentj @lhirlimann @Gustry @mdouchin @rldhont  : this is how a manual test might be added.
We have : 

- a *.md file with checkboxes to validate each tests
- a QGIS project + Lizmap *.cfg
- a dump (structure + data) of the schema containing tests data made from pgadmin4

For the dump, I'm not totally sure if we should have one SQL file for each QGIS + Lizmap project or a single one as I made.
- Several files can make easier to follow each test added but we have to make a dump of every tables/views needed
- One file is easier to make as we just make the dump from the schema. But it will harder to follow each change.

When everything is ok, I'll change docker configuration to automate what can be automated (drop/load sql dump(s) for example).
Then we could configure CI.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,7 @@ docker/*.zip
 tests/units/vendor/
 tests/units/composer.lock
 tests/qgis-projects/*
-!tests/qgis-projects/.empty
+!tests/qgis-projects/tests
 tests/qgis-server-plugins/*
 !tests/qgis-server-plugins/.empty
 assets/node_modules/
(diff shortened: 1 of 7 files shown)


rldhont:
Can you provide some lizmap config ?
* For exemple the admins group does not have the right : Overwrite filter by login for the repository containing this projet.

Can you provide some screen shots of the map ?

nboisteault:
> Can you provide some lizmap config ?
> 
> * For exemple the admins group does not have the right : Overwrite filter by login for the repository containing this projet.
> 
> Can you provide some screen shots of the map ?

This PR is just to validate the procedure. We'll add some more tests later.

PR #1921, opened 2020-11-19
[Bugfix] With lizmap plugin server side, override login by filter is not respected

rldhont:
The lizmapOGCRequest does not check if the user has the right to override the login by filter.

If the user has the right to override the login by filter, lizmap has to send it to QGIS Server with an extra parameters `LIZMAP_OVERRIDE_FILTER`.

* Funded by Grand Narbonne


rldhont:
Needed by https://github.com/3liz/lizmap-plugin/pull/306

nboisteault (approved):
Tested : OK

PR #1925, opened 2020-11-22
[Bugfix] With lizmap plugin server side, override login by filter is not respected in Headers

rldhont:
To finalize  #1921 add override login by filter in request headers


nboisteault:
@rldhont what behavior should it change ?

rldhont:
It is not for fcgi QGIS, it is to provide info in headers. Can you test if it does not introduce a regression ?

nboisteault (approved):
No regression seen with loginFilteredLayers option.

PR #1936, opened 2020-11-24
[refactor] Proxy/Request namespace and context

alagroy-42:
This PR is just a namespace change and the abstraction of Jelix with the `AppContextInterface`.
The `Proxy::getMap` has been moved to `WMS`  class because it was the only one using it.

### DIFF
--- a/lizmap/modules/admin/controllers/maps.classic.php
+++ b/lizmap/modules/admin/controllers/maps.classic.php
@@ -547,7 +547,7 @@ class mapsCtrl extends jController
     public function removeCache()
     {
         $repository = $this->param('repository');
-        $repoKey = \Lizmap\Request\Proxy::clearCache($repository);
+        $repoKey = \Lizmap\Request\Proxy::clearCache(lizmap::getRepository($repository));
         if ($repoKey) {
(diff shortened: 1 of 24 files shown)


laurentj (changes requested):
* lizmapOGCRequest::build should be outside this class. Probably the best place is lizmapProxy? (with the method name `buildOGCRequest()`
* lizmapOGCRequest should be an abstract class as it is not usable as is, I guess? (I don't know what OGC means...)
* I would prefer these names for the classes, because it will be more searchable in the code (but keep them in the `\Lizmap\Request` namespace): 
   * `OGCRequest` (or `OGCRequestAbstract`?) instead of `OGC` 
   * `WMSRequest` instead of `WMS`
   * `WFSRequest` instead of `WFS`
* some comments on request classes are missing: we don't know what means "OGC", "WMS" etc.. Comments on the class
should contain the meaning of the acronym, an url to the documentation of the protocol/format (wikipedia or else), and a summary about the purpose of the protocol/format
* I don't think static properties for services and appcontext in request classes are a good thing. it complexifies the code and it mades the use of the classes complicated as well as tests.
* The move of the map method should be done in a separate commit
* verify the doc comment of all method/properties you changed/added, some `@param`, `@return` are wrong or are missing for instance.

PR #1937, opened 2020-11-25
Fix comparison of modified values in qgis forms

laurentj:
Some values like `'0'` was not considered different from the empty value `''`.


rldhont (approved):

nboisteault (approved):
Tested. OK !

PR #1947, opened 2020-11-30
[Refactor] Proxy/Request namespace and context

alagroy-42:
#1936 with the new files already moved.

### DIFF
--- a/lizmap/modules/lizmap/lib/Request/Proxy.php
+++ b/lizmap/modules/lizmap/lib/Request/Proxy.php
@@ -51,6 +51,64 @@ class Proxy
         return self::$appContext;
     }
 
+    public static function build($project, $params, $requestXml = null)
+    {
+        $service = null;
+        $request = null;
+
+        // Check request XML
+        if ($requestXml && substr(trim($requestXml), 0, 1) == '<') {
+            $requestXml = trim($requestXml);
+        } else {
+            $requestXml = null;
+        }
+
+        // Parse request XML
+        if ($requestXml) {
+            $xml = simplexml_load_string($requestXml);
+            if ($xml) {
+                $request = $xml->getName();
+                if (property_exists($xml->attributes(), 'service')) {
+                    // OGC service has to be upper case for QGIS Server
+                    $service = strtoupper($xml['service']);
+                }
+            } else {
+                $requestXml = null;
+            }
+        }
+
+        // Check parameters
+        if (!$requestXml && isset($params['service'])) {
+            // OGC service has to be upper case for QGIS Server
+            $service = strtoupper($params['service']);
+            if (isset($params['request'])) {
+                $request = strtolower($params['request']);
+            }
+        }
+
+        if ($service == null) {
+            return null;
+        }
+        $params['service'] = $service;
+        if ($request !== null) {
+            $params['request'] = $request;
+        }
+        if ($service == 'WMS') {
+            return new WMSRequest($project, $params, self::setServices(), self::setAppContext(), $requestXml);
+        }
+        if ($service == 'WMTS') {
+            return new WMTSRequest($project, $params, self::setServices(), self::setAppContext(), $requestXml);
+        }
+        if ($service == 'WFS') {
+            return new WFSRequest($project, $params, self::setServices(), self::setAppContext(), $requestXml);
+            // Not yet
+        //} else if ($service == 'WCS') {
+        //    return new lizmapWCSRequest($project, $params, $requestXml)
+        }
+
+        return null;
+    }
+
     /**
      * Normalize and filter request parameters.
      *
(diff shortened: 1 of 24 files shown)


lizmap/modules/lizmap/lib/Request/WFSRequest.php
```diff
@@ -12,6 +12,9 @@
 
 namespace Lizmap\Request;
 
+/**
+ * https://en.wikipedia.org/wiki/Web_Feature_Service.
```
laurentj, line 16: the url should be behind `@see`

```
* @see https://en.wikipedia.org/wiki/Web_Feature_Service.
```

lizmap/modules/lizmap/lib/Request/WFSRequest.php
```diff
@@ -288,6 +300,10 @@ public function getfeature()
         return $this->getfeatureQgis();
     }
 
+    /**
+     * https://en.wikipedia.org/wiki/Web_Feature_Service#Static_Interfaces
+     * Queries Qgis Server for getFeature.
```
laurentj, line 305: decription should be above the `@see` (that you should add for the url).

Fix this for all comments you added

lizmap/modules/lizmap/lib/Request/Proxy.php
```diff
@@ -21,6 +21,94 @@ class Proxy
      */
     protected static $_profiles = array();
 
+    protected static $services = null;
+
+    protected static $appContext = null;
+
+    public static function setServices($services = null)
```
laurentj, line 28: The function name is not related on how the method is used, as it is used to get services.
I think we should have a `getServices()`, which return the current `self::$services` if not null,else `\lizmap::getServices()`. And we should have a `setServices($services)` to set  `self::$services`.

lizmap/modules/lizmap/lib/Request/Proxy.php
```diff
@@ -21,6 +21,94 @@ class Proxy
...
+                self::$services = $services;
+            } else {
+                self::$services = \lizmap::getServices();
+            }
+        }
+
+        return self::$services;
+    }
+
+    public static function setAppContext($appContext = null)
```
laurentj, line 41: same as setServices: we should have a `getAppContext()` and a `setAppContext($appContext)`

lizmap/modules/lizmap/lib/Request/Proxy.php
```diff
@@ -725,19 +528,24 @@ protected static function declareRedisProfile($ser, $cacheName, $repository, $pr
         }
 
         // Create the virtual cache profile
-        jProfiles::createVirtualProfile('jcache', $cacheName, $cacheParams);
+        self::setAppContext()->createVirtualProfile('jcache', $cacheName, $cacheParams);
     }
 
     /**
      * @param mixed $repository
+     * @param mixed $lrep
```
laurentj, line 536: the previous @param should be removed. And the type should be indicated.

lizmap/modules/lizmap/lib/Request/WMSRequest.php
```diff
@@ -847,4 +866,292 @@ protected function gfiGmlToHtml($gmldata, $configLayer)
...
+
+            // Keep original param value
+            $originalParams = array('width' => $params['width'], 'height' => $params['height']);
+            // Replace width and height before requesting the image from qgis
+            $params['width'] = $metatileSizeX * $params['width'] + 2 * $metatileBuffer;
+            $params['height'] = $metatileSizeY * $params['height'] + 2 * $metatileBuffer;
+        }
+
+        // Build params before send the request to Qgis
+        $builtParams = http_build_query($params);
```
laurentj, line 1061: Probably we could use Proxy::constructUrl() here ?

laurentj (changes requested):

lizmap/modules/lizmap/lib/Request/WMSRequest.php
alagroy-42: Done in next PR.

laurentj: okay

lizmap/modules/lizmap/lib/Request/Proxy.php
```diff
@@ -21,6 +21,96 @@ class Proxy
      */
     protected static $_profiles = array();
 
+    protected static $services = null;
+
+    protected static $appContext = null;
+
+    public static function setServices($services)
```
laurentj, line 28: Type of parameters is missing (in an `@param` or directly into the code).
Please document parameters and return values for all new methods.

laurentj (changes requested):
still have some issues on documentation. It will be ok after that

lizmap/modules/lizmap/lib/Request/Proxy.php
laurentj: the syntax is `@param <type> <parameter name>`

`@param \lizmapServices $services`. No two `@param` for the same parameter.

PR #1948, opened 2020-11-30
[refactor] Proxy and XXXRequest Refactorisation

alagroy-42:


lizmap/modules/lizmap/lib/Request/OGCRequest.php
```diff
@@ -271,4 +259,27 @@ protected function getcapabilities()
             'cached' => $cached,
         );
     }
+
+    protected function logXmlError($xmldata)
```
laurentj, line 263: The name of the method does not correspond to what it does. `loadXmlString` is a better name IMHO.

laurentj (changes requested):

laurentj (approved):

PR #1960, opened 2020-12-02
FixBug: mandatory relational value by default

drillinP:
* In Lizmap 3.3.12-pre, there is a default relational value type field bug.
* This PR fix this bug

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisForm.class.php
+++ b/lizmap/modules/lizmap/classes/qgisForm.class.php
@@ -1161,7 +1161,10 @@ class qgisForm implements qgisFormControlsInterface
             $dataSource = new jFormsStaticDatasource();
 
             // required
-            if (!$formControl->valueRelationData['allowNull']) {
+            if (array_key_exists('notNull', $formControl->valueRelationData)
+                and $formControl->valueRelationData['notNull']
+            ) {
+                jLog::log('notNull '.$formControl->valueRelationData['notNull'], 'error');
                 $formControl->ctrl->required = true;
             }
             // combobox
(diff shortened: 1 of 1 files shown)


lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -1161,7 +1161,10 @@ private function fillControlFromValueRelationLayer($fieldName, $formControl)
             $dataSource = new jFormsStaticDatasource();
 
             // required
-            if (!$formControl->valueRelationData['allowNull']) {
+            if (array_key_exists('notNull', $formControl->valueRelationData)
+                and $formControl->valueRelationData['notNull']
+            ) {
+                jLog::log('notNull '.$formControl->valueRelationData['notNull'], 'error');
                 $formControl->ctrl->required = true;
             }
```
drillinP, line 1169: I think the field is require when `notNull` is check and not `allowNull`

lizmap/modules/lizmap/classes/qgisForm.class.php
```diff
@@ -1173,7 +1176,7 @@ private function fillControlFromValueRelationLayer($fieldName, $formControl)
 
             // Add default empty value for required fields
             // Jelix does not do it, but we think it is better this way to avoid unwanted set values
-            if ($formControl->ctrl->required) {
+            if ($formControl->ctrl->required or $formControl->valueRelationData['allowNull']) {
```
drillinP, line 1179: If the field is not required but the `allowNull` is checked the relational value list will not have an empty value.
I add `or $formControl->valueRelationData['allowNull']` in the `if` to add empty value when `allowNull` is checked .

rldhont (approved):

lizmap/modules/lizmap/classes/qgisForm.class.php
rldhont: Are you sure that `notNull` is in `valueRelationData` ?

PR #1943, opened 2020-11-27
[bugfix] defining undefined variables in lizmapWMSRequest

alagroy-42:
* Ticket : fixes #1924

PR #1962, opened 2020-12-03
[tests] Proxy/XXXRequest

alagroy-42:
Here come the unit tests for the `Lizmap\Request` namespace classes. There is still some methods untested particualrly the ones that are retrieving data from an external service or the one that are handling a big volume of data that I don't have access to (XML/GML etc...) If you send me samples from these data (some valid and some invalid if it's possible), I will be able to write these tests.


tests/units/classes/Request/ClassesForTests.php
```diff
@@ -0,0 +1,146 @@
...
+        return $this->loginFilters;
+    }
+
+    public function setData($key, $value)
+    {
+        $this->data[$key] = $value;
+    }
+}
+
+class jTpl
```
laurentj, line 28: You cannot redefine classes that already exists, like jTpl or lizmapTiler, else we couldn't test components which are using them and need real results from them.And we couldn't test these classes too (at least lizmapTiler).

laurentj (changes requested):

lizmap/modules/lizmap/lib/Request/WMSRequest.php
```diff
@@ -188,7 +187,8 @@ protected function getcontext()
             array('repository' => $this->repository->getKey(), 'project' => $this->project->getKey())
         );
         $sUrl = str_replace('&', '&amp;', $sUrl).'&amp;';
-        $data = preg_replace('/xlink\:href=".*"/', 'xlink:href="'.$sUrl.'&amp;"', $response->data);
+        $data = $response->data;
```
laurentj, line 190: why this change?

laurentj (approved):

lizmap/modules/lizmap/lib/Request/WMSRequest.php
alagroy-42: https://github.com/3liz/lizmap-web-client/pull/1943#pullrequestreview-541093390

PR #1969, opened 2020-12-04
Update required value

drillinP:
Update to allow empty value or not  when multiple select is use


rldhont:
can you do the fix on 3.3 and not master ?

drillinP:
> can you do the fix on 3.3 and not master ?

the change is for 3_3. wrong branch in pr sorry

rldhont (approved):

PR #1970, opened 2020-12-04
Adding test project

drillinP:
*Adding test project after fixbug concerning the relational value


nboisteault:
@drillinP please make those changes

- Point lizmap_3_4 branch for merge. It is just tests so that can not break Lizmap and we'll want to test all 3.4.x versions.
- Default dbname is `lizmap` not `qgis`, so please change that in .qgs
- Make Lizmap project as light as possible (no OSM base layer, no unneeded map tools). Those tests will be added to CI so it has to be fast.
- Create all states to test in QGIS project. We need to avoid QGIS project changes by hand. So things like `Configure this field without allowing the non-null constraints and the default empty value.` have to be removed and you create as many fields as needed to test those states/configuration.
Thanks

PR #1981, opened 2020-12-10
[BUGFIX] qgisVectorLayerDatasource - Improve parsing for layers with sql

mdouchin:
This takes into account:

* layers with no `sql=` to make it compatible with QGIS >= 3.16
* layer with simple schema & table and with `sql=`. In the past, only complex layer with SQL inside the `table` property was covered by the tests.


mdouchin:
Close because wrong base branch !

PR #1998, opened 2020-12-17
Fix foreach warning in getLoginFilters()

nboisteault:
* Funded by 3Liz


lizmap/modules/lizmap/classes/lizmapProject.class.php
```diff
@@ -953,6 +953,10 @@ public function getLoginFilters($layers, $edition = false)
             return $filters;
         }
 
+        if (!$layers) {
```
laurentj, line 956: I prefer to have `if (!is_array($layers))`

laurentj (changes requested):

rldhont (approved):

PR #2016, opened 2021-01-06
Fix docker tests

dmarteau:
Fix docker tests:

* Set user uid/gid as build arguments for phpfpm Dockerfile
* Create `.env` file in `run-docker` instead of exporting variables
* Declare postgres version as compose parameter
* Fix postgres data volume 
* Set default postgres to postgres 11, postgis 2.5


laurentj (approved):
LGTM

PR #2018, opened 2021-01-06
[Bugfix] Draw: import KML does not draw anythings if xml headers

### LINKED ISSUE #2015: Draw: import KML does not draw anythings

1. Go to https://demo.lizmap.com/lizmap_3_4/index.php/view/map/?repository=demo&project=qgis_info
2. Click on 'Draw'
3. Click on 'Import in GeoJSON, GPX, KML' and choose a KML file [kmls.zip](https://github.com/3liz/lizmap-web-client/files/5775147/kmls.zip)
4. The file name is well displayed but nothing on the map and the buttons are still deactivated

### Expected behavior
Display the KML content at the right place

### Screenshots and logs
No logs in the browser console

### Environment

 - Lizmap version: 3.4.0
 - QGIS Server FULL version: 3.10
 - QGIS Desktop FULL version: 3.10
 - OS (Windows, Linux, MacOS, Android…): Linux
 - Browser (Firefox, Chrome…): Ubuntu Firefox
 - Lizmap admin information (users, logs…) stored into Sqlite (default)? PostgreSQL? MySQL?

rldhont:
If the KML file contain the xml header, it was not parsed.

Fixes #2015

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "editor.formatOnSave": true
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "editor.formatOnSave": true
-}
--- a/assets/src/modules/Digitizing.js
+++ b/assets/src/modules/Digitizing.js
@@ -265,7 +265,7 @@ export default class Digitizing {
 
         // Refresh layer
         this._drawLayer.redraw(true);
-        
+
         // Save color
         localStorage.setItem(this._repoAndProjectString + '_drawColor', this._drawColor);
 
@@ -346,10 +346,10 @@ export default class Digitizing {
 
             }
             return `<?xml version="1.0" encoding="UTF-8"?>
-                    <StyledLayerDescriptor xmlns="http://www.opengis.net/sld" 
-                        xmlns:ogc="http://www.opengis.net/ogc" 
-                        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.1.0" 
-                        xmlns:xlink="http://www.w3.org/1999/xlink" xsi:schemaLocation="http://www.opengis.net/sld http://schemas.opengis.net/sld/1.1.0/StyledLayerDescriptor.xsd" 
+                    <StyledLayerDescriptor xmlns="http://www.opengis.net/sld"
+                        xmlns:ogc="http://www.opengis.net/ogc"
+                        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" version="1.1.0"
+                        xmlns:xlink="http://www.w3.org/1999/xlink" xsi:schemaLocation="http://www.opengis.net/sld http://schemas.opengis.net/sld/1.1.0/StyledLayerDescriptor.xsd"
                         xmlns:se="http://www.opengis.net/se">
                         <UserStyle>
                             <FeatureTypeStyle>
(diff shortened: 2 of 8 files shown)


drillinP:
In the import function, there is a parameter called `file`. With this parameter we can get the file name and get the file extension.

nboisteault:
@drillinP Please make one commit for linting and another for your fix. It is easier to read.

drillinP:
@nboisteault It's done, in fact it's just that my vscode formats the code automatically when I save the file

drillinP (approved):
Good for me

rldhont:
@drillinP peux-tu créer 2 fichiers KML pour les tests :+1: 

1. Sans l'en-tête xml
2. Avec l'en-tête xml

assets/src/modules/Digitizing.js
```diff
@@ -472,19 +472,23 @@ export default class Digitizing {
...
         reader.onload = (() => {
             return (e) => {
                 const fileContent = e.target.result;
                 let OL6features;
 
                 // Handle GeoJSON, GPX or KML strings
                 try {
-                    if (fileContent[0] === '{') {
+                    // Check extension to the feature type
+                    if (fileExtension === 'geojson') {
```
rldhont, line 486: What if the extenion is json ?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "editor.formatOnSave": true
```
Gustry, line 2: Should we commit vscode user settings ?

assets/src/modules/Digitizing.js
drillinP: I can make `if (fileExtension === 'geojson' || fileExtension === 'json')` ?

.vscode/settings.json
nboisteault: No, we should not. Please @drillinP  ignore this file by editing `.git/info/exclude` and adding this file.

assets/src/modules/Digitizing.js
rldhont: @drillinP I'll do it

.vscode/settings.json
rldhont: @nboisteault @drillinP I'll do it

Gustry: Why `.git/info/exclude` and not just adding it in the `.gitignore`  in the repository? (the `.vscode` directory) ?

nboisteault: Yes we can do it this way but in this case we'll have to handle every files/directories created by different IDE, tools, etc.
As vscode is very popular, let's add this directory. It will avoid this directory to be committed by potential contributors in future.

rldhont: Done in `.gitignore`

assets/src/modules/Digitizing.js
rldhont: Done

assets/src/modules/Digitizing.js
```diff
@@ -469,29 +469,33 @@ export default class Digitizing {
         setTimeout(function () { URL.revokeObjectURL(a.href); }, 1500);
     }
 
-    import(file){
+    import(file) {
         const reader = new FileReader();
 
+        // Get extension file
+        var fileExtension = file.name.split('.').pop();
```
nboisteault, line 476: Please use avoid `var`. Use `let` or `const`. Here `const` is right.

tests/qgis-projects/tests/qgis_info.qgs
```diff
@@ -0,0 +1,201 @@
+<qgis projectname="" version="3.10.13-A Coruña">
```
nboisteault, line 1: @drillinP could you change file names. `qgis_info` doesn't tell what those files permit to test.

nboisteault (changes requested):

PR #2028, opened 2021-01-11
Add Cypress to CI w/ github actions

nboisteault:
* Funded by 3Liz


Gustry:
@nboisteault This is a good news !

I think, this should had been merged into `master`, we said we don't do any merge from downstream to upstream last week.

Will you do it ?

rldhont:
@Gustry not yet

Issue #2024, opened 2021-01-08
Undefined properties/methods shown by PHPStan


alagroy-42:
These are the last significative errors shown by PHPStan on master : 
```
 ------ ----------------------------------------------------------
  Line   lizmap/modules/action/classes/actionDatasource.class.php
 ------ ----------------------------------------------------------
  44     Call to an undefined method actionDatasource::error().
  48     Call to an undefined method actionDatasource::error().
 ------ ----------------------------------------------------------

 ------ ----------------------------------------------------------
  Line   lizmap/modules/filter/classes/filterDatasource.class.php
 ------ ----------------------------------------------------------
  49     Call to an undefined method filterDatasource::error().
  53     Call to an undefined method filterDatasource::error().
 ------ ----------------------------------------------------------

 ------ ----------------------------------------------------------------
  Line   lizmap/modules/lizmap/controllers/geobookmark.classic.php
 ------ ----------------------------------------------------------------
  16     Access to an undefined property geobookmarkCtrl::$whiteParams.
  81     Access to an undefined property geobookmarkCtrl::$whiteParams.
 ------ ----------------------------------------------------------------
```

@mdouchin, you are the one who worked on those files, can you confirm these are errors and not some tricky inheritance or something that I haven't spot ?

mdouchin:
I will !

alagroy-42:
There are also these 2 undefined properties in `qgisVectorLayerDatasource.class.php`

```

Access to an undefined property qgisVectorLayerDatasource::$provider. | build: lizmap/modules/lizmap/classes/qgisVectorLayerDatasource.class.php#L51
-- | --
Access to an undefined property qgisVectorLayerDatasource::$datasource. | build: lizmap/modules/lizmap/classes/qgisVectorLayerDatasource.class.php#L52
```

Gustry:
Feedback from PHPStan Action :) 

```
Error: Call to an undefined method actionDatasource::error().
Error: Call to an undefined method actionDatasource::error().
Error: Call to an undefined method filterDatasource::error().
Error: Call to an undefined method filterDatasource::error().
Error: Access to an undefined property qgisVectorLayerDatasource::$provider.
Error: Access to an undefined property qgisVectorLayerDatasource::$datasource.
Error: Access to an undefined property qgisVectorLayerDatasource::$datasource.
Error: Access to an undefined property qgisVectorLayerDatasource::$provider.
[...]

PR #2029, opened 2021-01-11
[GH Action] PHPStan setup

alagroy-42:
This PR introduces PHPStan as a github Action to staticly analize the php code on each PR/push on master.
Some errors are fixed, the others are in #2024.

The rule level is set to 0 (the minimum) for now, we can increase it progressively once each level is fixed.

@Gustry we have the choice between using [this action](https://github.com/chindit/actions-phpstan) or [this one](https://github.com/shivammathur/setup-php#phpstan) (that we already use to install dependencies), which one do you think is best ?

### DIFF
--- a/.github/workflows/php-stan.yml
+++ b/.github/workflows/php-stan.yml
@@ -44,5 +44,3 @@ jobs:
       - name: Run PHPStan
         run: phpstan analyse -c ../phpstan.neon
 
-      # - name: PHPStan
-      #   uses: chindit/actions-phpstan@master


.github/workflows/php-stan.yml
```diff
@@ -0,0 +1,31 @@
+name: phpstan
+on:
+  push:
+    branches: [ master ]
```
Gustry, line 4: I would extend that to 3.4 at least (and even 3.3 I think we can)

phpstan.neon
```diff
@@ -0,0 +1,14 @@
+parameters:
+  scanDirectories:
+    - lib/
+    - lizmap/vendor/
+  level: 0
```
Gustry, line 5: Have you tried increasing the level ?
I saw it's from 0 to 7, right ?
What would be the advantage of increasing ?
Should we increasse step by step because there is too much error at level N+1 ?

Gustry (commented):
Nice move, thanks !

phpstan.neon
alagroy-42: Yes, increasing the level brings a lot of error, doing it progressively seems a better idea.
Level 1 checks 'possibly' undefined variables so it will introduce a lot of false positive and level >= 2 checks methods and types based on PHPDocs so we will first need to do/complete rigourously all of them if we want it to be effective.

rldhont:
Some changes need to be backported in release_3_4.

Gustry:
> Some changes need to be backported in release_3_4.

Do you mean before 3.4.1 or after ?

Thanks @alagroy-42 I understand. I will go step by step with phpstan if we find it good.

.github/workflows/php-stan.yml
```diff
@@ -0,0 +1,37 @@
...
+      - uses: actions/checkout@v2
+
+      - name: Setup PHP with tools
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
+          tools: phpstan
+
+      - name: Install dependencies
+        run: composer install --prefer-dist --no-progress --no-suggest
```
laurentj, line 31: is there a way in github workflows to configure a cache of files downloaded  by Composer (or any other packager like npm etc ?)

Gustry: It's possible to setup cache on GH Action.
We can look to that after ?

Gustry:
One part of this work has been merged in release 3.4 with #2041. You should wait for the merge on master and then rebase this branch.

rldhont:
@alagroy-42 can rebase your master branch ?

Gustry:
Thanks for your initial work @alagroy-42 

Merging manually.

PR #2034, opened 2021-01-12
FTS PostgreSQL Search - Use a transaction to avoid PostgreSQL connection issue

mdouchin:
This solves an issue when Lizmap Web Client could not find the `lizmap_search` table: in some cases, the following PostgreSQL queries run by Lizmap, for example when authenticating the user, woul fail with a `ERROR: current transaction is aborted` message.

* Ticket : #2008
* Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmapFtsSearch.listener.php
+++ b/lizmap/modules/lizmap/classes/lizmapFtsSearch.listener.php
@@ -31,20 +31,30 @@ class lizmapFtsSearchListener extends jEventListener
         $profile = 'search';
 
         try {
-            // try to get the specific search profile to do not rebuild it
+            // try to get the specific search profile
             jProfiles::get('jdb', $profile, true);
         } catch (Exception $e) {
-            // else use default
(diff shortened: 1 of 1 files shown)


lizmap/modules/lizmap/classes/lizmapFtsSearch.listener.php
```diff
@@ -38,13 +38,24 @@ protected function checkLizmapFts()
             $profile = null;
         }
 
+        // Lizmap search available only for PostgreSQL
+        $cnx = jDb::getConnection($profile);
+        $dbms = $cnx->dbms;
```
laurentj, line 43: why creating a variable, since it is used only once?

mdouchin: it is indeed useless. I will remove this unnecessary step

nboisteault (approved):
@mdouchin I have just tested it does not break Lizmap but I don't know how to reproduce bug you're fixing

PR #2041, opened 2021-01-13
Fixing errors flagged by PhpStan [For Release 3_4]

drillinP:
* Getting PhpStan from @alagroy-42  [commit](https://github.com/3liz/lizmap-web-client/pull/2029/commits/e71a15e9f00758e304ad72eb1fd7a40539432570)
* Fixing errors flagged by PhpStan

### DIFF
--- a/lizmap/modules/action/classes/actionDatasource.class.php
+++ b/lizmap/modules/action/classes/actionDatasource.class.php
@@ -41,11 +41,11 @@ class actionDatasource
         jClasses::inc('action~actionConfig');
         $dv = new actionConfig($repository, $project);
         if (!$dv->getStatus()) {
-            return $this->error($dv->getErrors());
+            return $dv->getErrors();
         }
         $config = $dv->getConfig();
         if (empty($config)) {
-            return $this->error($dv->getErrors());
+            return $dv->getErrors();
         }
 
         $this->repository = $repository;
--- a/lizmap/modules/filter/classes/filterDatasource.class.php
+++ b/lizmap/modules/filter/classes/filterDatasource.class.php
@@ -46,11 +46,11 @@ class filterDatasource
         jClasses::inc('filter~filterConfig');
         $dv = new filterConfig($repository, $project);
         if (!$dv->getStatus()) {
-            return $this->error($dv->getErrors());
+            return $dv->getErrors();
         }
         $config = $dv->getConfig();
         if (empty($config)) {
-            return $this->error($dv->getErrors());
+            return $dv->getErrors();
         }
 
         $this->repository = $repository;
--- a/lizmap/modules/lizmap/classes/qgisVectorLayerDatasource.class.php
+++ b/lizmap/modules/lizmap/classes/qgisVectorLayerDatasource.class.php
@@ -33,6 +33,10 @@ class qgisVectorLayerDatasource
         'sql' => ' sql=(.*)$',
     );
 
+    protected $provider;
+
+    protected $datasource;
+
     // /!\ in table and geocol regex above, sql= will be removed when search for table or geocol
     // to remove issues
     // better would be to find a robust regex, but it is fragile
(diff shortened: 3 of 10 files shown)


lizmap/modules/action/classes/actionDatasource.class.php
```diff
@@ -41,11 +41,11 @@ public function __construct($repository, $project, $layerId)
         jClasses::inc('action~actionConfig');
         $dv = new actionConfig($repository, $project);
         if (!$dv->getStatus()) {
-            return $this->error($dv->getErrors());
+            return $dv->getErrors();
```
drillinP, line 44: @rldhont **Phpstan** say  the `error` method doesn't exists, so i think we need to return `$dv->geErrors()` directly or define `$this->errors = $dv->geErrors()` and return this ?

lizmap/modules/filter/classes/filterDatasource.class.php
```diff
@@ -46,11 +46,11 @@ public function __construct($repository, $project, $layerId)
         jClasses::inc('filter~filterConfig');
         $dv = new filterConfig($repository, $project);
         if (!$dv->getStatus()) {
-            return $this->error($dv->getErrors());
+            return $dv->getErrors();
```
drillinP, line 49: Same as previous comment

lizmap/modules/lizmap/classes/qgisVectorLayerDatasource.class.php
```diff
@@ -33,6 +33,10 @@ class qgisVectorLayerDatasource
         'sql' => ' sql=(.*)$',
     );
 
+    protected $provider;
+
+    protected $datasource;
```
drillinP, line 38: **PhpStan** Say that ``` Access to an undefined property qgisVectorLayerDatasource::$provider. ``` and ```
 Access to an undefined property qgisVectorLayerDatasource::$datasource.``` [for this two line](https://github.com/3liz/lizmap-web-client/pull/2041/commits/76e291380881b866d6e079da0a8fbcbf0160c350#diff-520ea96b02586d47aa1d3ab69edf6f419050a52814f16d8f6e8e9a4f00d53bd7L51-L52)

lizmap/modules/lizmap/controllers/geobookmark.classic.php
```diff
@@ -11,6 +11,8 @@
  */
 class geobookmarkCtrl extends jController
 {
+    protected $whiteParams;
```
drillinP, line 14: Here **PhpStan** fllagged the error `Access to an undefined property geobookmarkCtrl::$whiteParams.` so i think we need to define `protected $$whiteParams`

lizmap/modules/action/classes/actionDatasource.class.php
rldhont: @mdouchin can you confirm ?

lizmap/modules/filter/classes/filterDatasource.class.php
rldhont: @mdouchin can you confirm ?

rldhont (changes requested):
Can you run php-cs-fixer to fix the test.

rldhont:
@drillinP can you also rebase the branch ?

Gustry:
You should include all the content of the action, CF https://github.com/3liz/lizmap-web-client/pull/2029/files with the cache mechanism.

rldhont (approved):

Gustry (approved):
Just LGTM because it's green for phpstan !
We can polish the action later.

I will let PHP devs the real review.

lizmap/modules/action/classes/actionDatasource.class.php
mdouchin: I confirm the proposed fix is OK. By the way, it seems the whole class actionDatasource is never used in Lizmap ! I will review this module code for next point release !

lizmap/modules/filter/classes/filterDatasource.class.php
mdouchin: I confirm the fix is ok

mdouchin (approved):

PR #2043, opened 2021-01-13
[Bugfix] Tab content is not visible on mobile screen

### LINKED ISSUE #2019: Tab content is not visible using Edition on Android (Lizmap 3.4.0)
### What is the bug?
Feature edit does not work on mobile devices if the edit form is organized in tabs (Drag and drop designer). Simply tab content does not show on mobile devices. On desktop interface it works without any problem.

### Environment

 - Lizmap version: 3.4.0
 - QGIS Server FULL version: 3.10
 - QGIS Desktop FULL version: 3.10.6
 - OS (Windows, Linux, MacOS, Android…): error on Android (phone), works on Windows (desktop)
 - Browser (Firefox, Chrome…): Chrome

rldhont:
Fixes #2019 Tab content is not visible using Edition on Android (Lizmap 3.4.0)

The way lizmap checks the edition form tab visibility was not the right way. Lizmap checks visibility instead of checks if tab has not display:none

* Funded by 3liz

To test in a desktop browser:
* [ ] Go to a map with creation capabilities and a form with QGIS Drag an Drop form configuration and tabs
* [ ] Activate mobile display
* [ ] Reload the page to get the mobile UI
* [ ] Go to Edition and click on Add
* [ ] Go to Edition to display the form and check that the form content is well displayed

### DIFF
--- a/lizmap/www/assets/js/edition.js
+++ b/lizmap/www/assets/js/edition.js
@@ -1523,7 +1523,11 @@ OpenLayers.Geometry.pointOnSegment = function(point, segment) {
                 }
             }
             // Check li (tabs) visibility
-            if (form.children('ul.nav-tabs').find('li:visible').length == 0 ) {
+            var visibleTabs = form.children('ul.nav-tabs').find('li').filter(
+                function(){
+                    return $(this).css('display') !== 'none';
+                });
+            if (visibleTabs.length == 0 ) {
                 // No tabs visible, hide the tab content
                 $('#'+form.attr('id')+'-tab-content').hide();
             } else {


rldhont:
To test you can use the dnd_form project https://github.com/3liz/lizmap-web-client/blob/release_3_4/tests/qgis-projects/tests/dnd_form.qgs

drillinP (approved):
Tested with many local project the form content is well displayed for me.

nboisteault (commented):
@rldhont ok with what I read but I can't test before tomorrow afternoon.

mdouchin (approved):
Ok for me, the manual tests are passing. But in the case of a geometryless table (dnd test), when you clic on the button Add, it opens the form but afterward the dock is hidden (to let the user digitize a geometry, but there is not). So I approve because it fixes the bug, but we need to catch the "geometryless" weird behaviour for a next point release and "just" show the form in this case. cc @nboisteault

PR #2048, opened 2021-01-14
[Tests] Enhance Drag'N Drop Edition form test to avoid regression

### LINKED ISSUE #2038: Add a test for the fix related to the PR #2013
The PR #2013 called "[Bugfix] Attribute edition sets to null unedited fields " has fixed a bug in 3.3.x and would be merged into release_3_4
We must add a test in 3.4 (or in master and cherry-pick into 3.4) for this fix

rldhont:
Fixes #2038

### DIFF
--- a/tests/qgis-projects/tests/dnd_form.sql
+++ b/tests/qgis-projects/tests/dnd_form.sql
@@ -64,9 +64,8 @@ ALTER TABLE ONLY tests_projects.dnd_form ALTER COLUMN id SET DEFAULT nextval('te
 -- Data for Name: dnd_form; Type: TABLE DATA; Schema: tests_projects; Owner: lizmap
 --
 
-COPY tests_projects.dnd_form (id, field_in_dnd_form, field_not_in_dnd_form) FROM stdin;
-1	test	test
-\.
+INSERT INTO tests_projects.dnd_form (id, field_in_dnd_form, field_not_in_dnd_form) VALUES
+    (1, 'test', 'test');
 
 
 --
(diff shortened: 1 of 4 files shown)


tests/qgis-projects/tests/dnd_form.sql
```diff
@@ -65,7 +65,7 @@ ALTER TABLE ONLY tests_projects.dnd_form ALTER COLUMN id SET DEFAULT nextval('te
 --
 
 COPY tests_projects.dnd_form (id, field_in_dnd_form, field_not_in_dnd_form) FROM stdin;
-1	test	test
+1   test    test
```
mdouchin, line 68: There is an error here because of the replacement of TAB by spaces -> the SQL does not work

mdouchin (changes requested):

tests/qgis-projects/tests/dnd_form.sql
mdouchin: Do not forget to fix the other COPY statement

Gustry: > There is an error here because of the replacement of TAB by spaces -> the SQL does not work

So we should a test on the test ;-)

rldhont: Fix with `INSERT INTO ... VALUES`

tests/qgis-projects/tests/dnd_form.sql
```diff
@@ -84,6 +83,73 @@ ALTER TABLE ONLY tests_projects.dnd_form
...
+
+ALTER TABLE ONLY tests_projects.dnd_form_geom ALTER COLUMN id SET DEFAULT nextval('tests_projects.dnd_form_geom_id_seq'::regclass);
+
+
+--
+-- Data for Name: dnd_form_geom; Type: TABLE DATA; Schema: tests_projects; Owner: lizmap
+--
+
+INSERT INTO tests_projects.dnd_form_geom (id, field_in_dnd_form, field_not_in_dnd_form, geom) VALUES
+    (1, 'test_geom, 'test_geom, '01010000206A080000BF599997AB39254116EA7038651D5841');
```
mdouchin, line 134: syntax error: quotes missing !

rldhont: Fixed thanks

drillinP (approved):

mdouchin (approved):

PR #2054, opened 2021-01-16
Cypress ci fix

nboisteault:


nboisteault:
@Gustry I have tried the backport bot but it failed. There is this message here : 
`Warning: Unexpected input(s) 'bot_token_key', valid inputs are ['bot_username', 'bot_token', 'github_token']`
It is weird as `bot_token_key` is in this action doc.
Did you test to backport manually as written above?

Gustry:
I have tried, no error indeed. Sad ...

I think you can skip this warning.

Gustry:
Ok, the error is another part, CF logs.

Gustry:
I think your issue is due to the update of a workflow file **within** this PR.
For security reason, updating a workflow file with a token require a special token scope, and the bot 3liz-bot hasn't this scope in its token.

Let's check if it's the reason with another pull request that will be backported. So I won't update the token because this PR is quite a special use-case.

I will backport this PR manually.

Issue #2055, opened 2021-01-18
Apply php-cs-fixer 2.18.0


rldhont:
The new php-cs-fixer version 2.18.0 discovered a lint error on lizmap/modules/admin/classes/config.listener.php
```
-        if (/*!$event->himself && */jAcl2::check('acl.user.view')) {
+        if (/*!$event->himself && */ jAcl2::check('acl.user.view')) {
```

Gustry:
Is the version of PHP-Cs-Fixer not fixed in the source code ?

PR #2056, opened 2021-01-20
Popup children fetchall

nboisteault:
Fetch all children popup instead of using jQuery post.
This allows having new `lizmappopupallchildrendisplayed` event launched when all children popup have been fetched.

* Funded by Grand Narbonne


Gustry:
Lint PHP due to https://github.com/3liz/lizmap-web-client/issues/2055

PR #2076, opened 2021-01-28
Adding new test to the feature of filtering by tags

drillinP:
New PR to add test to the feature of filter by tags.

### DIFF
--- a/tests/qgis-projects/tests/test_tags.md
+++ b/tests/qgis-projects/tests/test_tags.md
@@ -1,0 +1,7 @@
+For this test you need the two empty projects (test_tags_nature_flower, test_tags_nature_tree).
+This is to test the feature to filter by tags.
+
+On lizmap in the projects page activate search by tags in the search bar.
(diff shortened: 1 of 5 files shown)


nboisteault:
@drillinP Please rebase this PR with `master` and target `master` not `release_3_4`.

drillinP:
@nboisteault I have a problem with some file when i want rebase with `master` in local. some files change on their own like: `lizmap/var/config/admin/config.ini.php`

nboisteault:
@drillinP Do `git status`. If those files have just spaces that differ, you can `git restore` them.

drillinP:
@nboisteault I use git restore but this problem appear when I use git merge

Gustry:
@drillinP When writing an error, you should include the result of `git status` for instance, add the full error message and also your workflow of git commands. There are so many ways and commands to work with git ;-)
And I don't think `git merge` is the good one :)

PR #2092, opened 2021-02-05
[Bugfix] Assert $layer exists

nboisteault:


nboisteault:
@rldhont This commit has to be made in the same file but in another path in release_3_4. How do I do the backport?

rldhont:
@nboisteault Once this PR merged, you can opne a new one for release_3_4 with _manually backport #2092_ as description

rldhont (approved):

Gustry:
I'm not a PHP dev, but it seems these 70 edited lines could be replaced by 3 lines ?
```php
if (!$layer) {
    return array();
}
```
?
IMHO, it's better visibility :
* keeping our code to the left side with not too much indentation
* exiting from the code as soon as we can

Sorry for the noise if I'm wrong, just trying to read the PR :)

rldhont:
@Gustry you are right
```php
if (!$layer) {
    return $result;
}
```
this code could have been used

PR #2104, opened 2021-02-11
Fix the sqlite to pgsql migrator

laurentj:
* Funded by 3liz


rldhont (approved):

nboisteault (approved):
Tested with Grand Narbonne. OK

PR #2133, opened 2021-02-19
Run refreshDatatableSize after button Refresh was pressed or when a dock is oppened or closed

ygorigor:
Target version: LWC 3.4.4, LWC 3.5

If bottom dock is opened, when UI size is changed, columns width of active table will be recomputed.

Without this patch if we press on button `Refresh`  the width of table columns is resetting to 0. 

This commit will fix this.


rldhont:
Hi @ygorigor thanks for your contribution.

Can you add a description to your pull request ?

Can you describe the bug fixed ?

Can you describe the targeted version of lizmap web client ?

rldhont (approved):

PR #2137, opened 2021-02-19
Fix geobookmark sql: remove explicit public schema

laurentj:
The table geobookmark should not be created explicitly into the public schema.

All Lizmap tables must be installed in the same schema, e.g. in the
first schema indicated into the search_path of the connection.

* Funded by 3Liz


rldhont (approved):

rldhont:
@laurentj can be merged

PR #2169, opened 2021-02-26
Fix: WMTS GetTile request with 0 as parameter value

rldhont:
The WMTS GetTile request with TILEMATRIX=0&TILEROW=0&TILECOL=0 returned
```
<ows:ExceptionReport xsi:schemaLocation="http://www.opengis.net/ows/1.1 http://schemas.opengis.net/ows/1.1.0/owsExceptionReport.xsd" version="1.0.0" xml:lang="en">
<ows:Exception exceptionCode="MissingParameter">
<ows:ExceptionText>The parameter TileMatrixId is mandatory!</ows:ExceptionText>
</ows:Exception>
</ows:ExceptionReport>
```
Even if the tileMatrixId 0 is a valid value.

* Ticket : #
* Funded by

### DIFF
--- a/lizmap/modules/lizmap/lib/Request/WMTSRequest.php
+++ b/lizmap/modules/lizmap/lib/Request/WMTSRequest.php
@@ -98,8 +98,8 @@ class WMTSRequest extends OGCRequest
 
         foreach ($params as $var => $param) {
             ${$var} = $this->param($param);
-            if (!${$var}) {
-                \jMessage::add('The parameter '.$var.' is mandatory!', 'MissingParameter');
+            if (${$var} === '' || ${$var} === null) {
+                \jMessage::add('The parameter '.$param.' is mandatory!', 'MissingParameter');
 
                 return $this->serviceException();
             }


rldhont:
To test it, run index.php/lizmap/service/?repository=testsrepository&project=test_cache&LAYERS=Quartiers&STYLES=défaut&VERSION=1.0.0&EXCEPTIONS=application%2Fvnd.ogc.se_inimage&FORMAT=image%2Fpng&DPI=96&TRANSPARENT=true&SERVICE=WMTS&REQUEST=GetTile&LAYER=Quartiers&STYLE=default&TILEMATRIXSET=EPSG%3A3857&TILEMATRIX=0&TILEROW=0&TILECOL=0 in test environment

@nboisteault can test it with lizmap-web-client 3.4 and 3.3 to check if a backport is needed.

lizmap/modules/lizmap/lib/Request/WMTSRequest.php
```diff
@@ -98,7 +98,7 @@ public function gettile()
 
         foreach ($params as $var => $param) {
             ${$var} = $this->param($param);
-            if (!${$var}) {
+            if (!${$var} && !is_numeric(${$var})) {
```
laurentj, line 101: As LayerName is not numeric (I guess), so it will fail because of LayerName.

I suggest to have this test instead: `if (${$var} === '' || ${$var} === null) {`

laurentj (changes requested):

lizmap/modules/lizmap/lib/Request/WMTSRequest.php
rldhont: Thanks changes done. I will rebase the code.

lizmap/modules/lizmap/lib/Request/WMTSRequest.php
```diff
@@ -98,7 +98,7 @@ public function gettile()
 
         foreach ($params as $var => $param) {
             ${$var} = $this->param($param);
-            if (!${$var}) {
+            if (${$var} === '' || ${$var} === null) {
                 \jMessage::add('The parameter '.$var.' is mandatory!', 'MissingParameter');
```
nboisteault, line 102: @rldhont `$var` should be replaced by `$param`

nboisteault (changes requested):
Tested with 3.4 =>ok. No need to backport to 3.3.

lizmap/modules/lizmap/lib/Request/WMTSRequest.php
nboisteault: @rldhont This comment is for line 102 and not 101.

nboisteault (approved):

PR #2181, opened 2021-03-04
Fix 'Undefined index' notice

nboisteault:
* Funded by 3Liz

@rldhont @mdouchin @laurentj  just a simple review without test should be enough. If you run form_edition_spec.js with cypress you'll get this notice if needed.

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisFormControl.class.php
+++ b/lizmap/modules/lizmap/classes/qgisFormControl.class.php
@@ -238,6 +238,8 @@ class qgisFormControl
             $this->required = false;
         }
 
+        $markup = '';
+
         if ($this->fieldDataType != 'geometry') {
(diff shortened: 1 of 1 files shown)


lizmap/modules/lizmap/classes/qgisFormControl.class.php
```diff
@@ -308,7 +310,7 @@ public function __construct($ref, $edittype, $prop, $aliasXml = null, $defaultVa
                         $markup = 'time';
                     }
                 }
-            } else {
+            } elseif (in_array($this->fieldEditType, $this->qgisEdittypeMap)) {
```
mdouchin, line 313: I would add a else condition which uses the default 'input' markup

mdouchin (changes requested):

mdouchin (approved):

PR #2188, opened 2021-03-09
Fix regression in qgisFormControl constructor

rldhont:
Followup bf4ded5

Use array_key_exists instead of in_array for key identification in array.

For users the issue was that the Hidden Field defined in QGIS Form is not hidden in Lizmap form.
Use form_edition_all_field_type to check it.


mdouchin (approved):
👍

PR #2192, opened 2021-03-12
[BUGFIX] Get only projects metadata for the landing page

### LINKED ISSUE #2191: Projects page : it takes too much memory
### What is the bug?

When there is many projects to display, PHP can crash because of lack of memory.

This is because Lizmap loads all projects files whether from files or from the cache. So it can eat rapidely all the memory.

### Expected behavior

No crash because of lack of memory when displaying the projects list

### Proposition

For the project list, there should be a loop that reads projects one after the other. For each project, it should extract only informations needed for the display, and it should release the project object or any other variables that contain project data, in order to not eat more memory at each item.

mdouchin:
This will lower the memory usage when listing all projects information when building the landing page (or listing the projects of a repository). Fix #2191 

* Ticket : #2191
* Funded by 3liz

### DIFF
--- a/lizmap/modules/admin/classes/listProjectDatasource.class.php
+++ b/lizmap/modules/admin/classes/listProjectDatasource.class.php
@@ -22,13 +22,13 @@ class listProjectDatasource extends jFormsDynamicDatasource
         $criteria = $form->getData($this->criteriaFrom[0]);
         if ($criteria && array_key_exists($criteria, $this->data)) {
             $rep = lizmap::getRepository($criteria);
-            $projects = $rep->getProjects();
-            foreach ($projects as $p) {
-                $pOptions = $p->getOptions();
-                if (property_exists($pOptions, 'hideProject') && $pOptions->hideProject == 'True') {
+            // Get projects metadata
+            $metadata = $rep->getProjectsMetadata();
+            foreach ($metadata as $meta) {
+                if ($meta->getHidden()) {
                     continue;
                 }
-                $pdata[$p->getData('id')] = (string) $p->getData('title');
+                $pdata[$meta->getId()] = $meta->getTitle();
             }
         }
 
--- a/lizmap/modules/lizmap/classes/lizmapProject.class.php
+++ b/lizmap/modules/lizmap/classes/lizmapProject.class.php
@@ -139,6 +139,16 @@ class lizmapProject
         return $this->proj->getXmlLayer($layerId);
     }
 
+    /**
+     * Get the minimum needed project information for some pages (landing page, admin project listing).
+     *
+     * @return ProjectMetadata
+     */
+    public function getMetadata()
+    {
+        return $this->proj->getMetadata();
+    }
+
     public function findLayerByAnyName($name)
     {
         return $this->proj->findLayerByAnyName($name);
--- a/lizmap/modules/lizmap/classes/lizmapRepository.class.php
+++ b/lizmap/modules/lizmap/classes/lizmapRepository.class.php
@@ -111,13 +111,36 @@ class lizmapRepository
         return $this->repo->update($data, $ini);
     }
 
-    public function getProject($key)
+    /**
+     * Get a project by key.
+     *
+     * @param string $key           the project key
+     * @param bool   $keepReference if we need to keep reference in projectInstances
+     *
+     * @return null|Project null if it does not exist
+     */
+    public function getProject($key, $keepReference = true)
     {
-        return $this->repo->getProject($key);
+        return $this->repo->getProject($key, $keepReference);
     }
 
+    /**
+     * Get the repository projects.
+     *
+     * @return Project[]
+     */
     public function getProjects()
     {
         return $this->repo->getProjects();
+    }
+
+    /**
+     * Get the repository projects metadata.
+     *
+     * @return ProjectMetadata[]
+     */
+    public function getProjectsMetadata()
+    {
+        return $this->repo->getProjectsMetadata();
     }
 }
--- a/lizmap/modules/lizmap/lib/Project/Project.php
+++ b/lizmap/modules/lizmap/lib/Project/Project.php
@@ -467,6 +467,16 @@ class Project
         return $this->qgis->getData($key);
     }
 
+    /**
+     * Get the minimum needed project information for some pages (landing page, admin project listing).
+     *
+     * @return ProjectMetadata
+     */
+    public function getMetadata()
+    {
+        return new ProjectMetadata($this);
+    }
+
     public function getProj4($authId)
     {
         return $this->qgis->getProj4($authId);
--- a/lizmap/modules/lizmap/lib/Project/Repository.php
+++ b/lizmap/modules/lizmap/lib/Project/Repository.php
@@ -189,7 +189,15 @@ class Repository
         return $modified;
     }
 
-    public function getProject($key)
+    /**
+     * Get a project by key.
+     *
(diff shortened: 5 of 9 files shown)


mdouchin:
Not sure automatic backport to 3.4 and 3.3 will be easy, since the Lizmap classes refactoring has moved many project and repository functions in the lib directory. I will investigate

mdouchin:
@rldhont please add the correct milestone(s) and tag(s) as requested by the recent LWC workflow

mdouchin:
@laurentj I could have removed the method `getProjects` since it is not used in the code anymore, but thought we could still have a need in the future ?

Gustry:
> Not sure automatic backport to 3.4 and 3.3 will be easy, since the Lizmap classes refactoring has moved many project and repository functions in the lib directory. I will investigate

If the bot fail to 3.4, then manual cherry-pick, you can still do a PR if needed and use the bot to go to 3.3. Up to you then.

lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -467,6 +467,36 @@ public function getData($key)
         return $this->qgis->getData($key);
     }
 
+    /**
+     * Get the minimum needed project information for some pages (landing page, admin project listing).
+     *
+     * @return array of project metadata
+     */
+    public function getMetadata()
```
laurentj, line 475: I really want an object, not an array as a returned value. With an array, it's hard elsewhere in the code, to identify what it contains, where it is build etc.. I doesn't help to maintain the code. With a class/object, and with the help of an editor, we don't have such issues. We can have typed parameters on methods we gives this object etc.. So we have more robust code. 

I really want to eradicate all these kind of arrays we have in our code.

laurentj (changes requested):

laurentj:
> @laurentj I could have removed the method `getProjects` since it is not used in the code anymore, but thought we could still have a need in the future ?

Keep it for the moment.. We don't know if some modules are using it. And the patch is for stable branch, so we must not introduce some API incompatibilities

lizmap/modules/lizmap/lib/Project/Project.php
mdouchin: Ok, I will add a new class projectMetadata

laurentj: The new class should have getter methods. To ease the usage, the constructor may accept an array like you build..
Something like this:
```
namespace Lizmap\Project;
class ProjectMetadata {

  function __construct($repository, $id, $metadata);

  function getTitle() {... }
...
}
```

lizmap/modules/admin/classes/listProjectDatasource.class.php
```diff
@@ -22,13 +22,13 @@ public function getData($form)
...
-                $pOptions = $p->getOptions();
-                if (property_exists($pOptions, 'hideProject') && $pOptions->hideProject == 'True') {
+            // Get projects metadata
+            $metadata = $rep->getProjectsMetadata();
+            foreach ($metadata as $meta) {
+                if ($meta->getHidden()) {
                     continue;
                 }
-                $pdata[$p->getData('id')] = (string) $p->getData('title');
+                $pdata[$meta->getId()] = (string) $meta->getTitle();
```
laurentj, line 31: we shouldn't have to do `(string)`.

lizmap/modules/lizmap/lib/Project/Repository.php
```diff
@@ -205,11 +213,18 @@ public function getProject($key)
...
+            $this->projectInstances[$key] = $proj;
+        }
 
         return $proj;
     }
 
+    /**
+     * Get the repository projects.
+     *
+     * @return array of lizmapProject
```
laurentj, line 226: the right syntax is `@return Project[]`

lizmap/modules/lizmap/lib/Project/Repository.php
```diff
@@ -189,7 +189,15 @@ public function update($data, $ini)
...
     }
 
-    public function getProject($key)
+    /**
+     * Get a project by key.
+     *
+     * @param string $key           the project key
+     * @param bool   $keepReference if we need to keep reference in the repository property projectInstances
+     *
+     * @return null|lizmapProject null if it does not exist
```
laurentj, line 198: `@return null|Project` (but yes it is `lizmapProject into 3.4/3.3)

lizmap/modules/lizmap/lib/Project/Repository.php
```diff
@@ -249,4 +267,52 @@ public function getProjects()
 
         return $projects;
     }
+
+    /**
+     * Get the repository projects metadata.
+     *
+     * @return array of project metadata
```
laurentj, line 274: `@return ProjectMetada[]`  (or lizmapProjectMetadata for 3.4/3.3

the goal is to help editors ;-)

lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -467,6 +467,16 @@ public function getData($key)
         return $this->qgis->getData($key);
     }
 
+    /**
+     * Get the minimum needed project information for some pages (landing page, admin project listing).
+     *
+     * @return array of project metadata
```
laurentj, line 473: not an array, a single metada

`@return ProjectMetada`

lizmap/modules/lizmap/classes/lizmapRepository.class.php
```diff
@@ -111,13 +111,36 @@ public function update($data, $ini)
...
+    public function getProject($key, $keepReference = true)
     {
-        return $this->repo->getProject($key);
+        return $this->repo->getProject($key, $keepReference);
     }
 
+    /**
+     * Get the repository projects.
+     *
+     * @return array of lizmapProject
```
laurentj, line 130: `@return Project[]`

lizmap/modules/lizmap/classes/lizmapRepository.class.php
```diff
@@ -111,13 +111,36 @@ public function update($data, $ini)
...
+     */
     public function getProjects()
     {
         return $this->repo->getProjects();
     }
+
+    /**
+     * Get the repository projects metadata.
+     *
+     * @return array of projects metadata
```
laurentj, line 140: `@return ProjectMetada[]`

lizmap/modules/lizmap/classes/lizmapRepository.class.php
```diff
@@ -111,13 +111,36 @@ public function update($data, $ini)
...
     }
 
-    public function getProject($key)
+    /**
+     * Get a project by key.
+     *
+     * @param string $key           the project key
+     * @param bool   $keepReference if we need to keep reference in projectInstances
+     *
+     * @return null|lizmapProject null if it does not exist
```
laurentj, line 120: `@return null|Project`

lizmap/modules/lizmap/classes/lizmapProject.class.php
```diff
@@ -139,6 +139,11 @@ public function getXmlLayer($layerId)
         return $this->proj->getXmlLayer($layerId);
     }
 
+    public function getMetadata()
```
laurentj, line 142: add a doc comment with `@return ProjectMetada`

laurentj (approved):
Ok for me, after fixing doc comments

nboisteault (approved):

PR #2199, opened 2021-03-17
[BUGFIX] Get only projects metadata for the landing page

mdouchin:
Avoid high memory usage when there is a lot of projects

Manual backport 05c2eb9d5bace5c442e12704a37e3ad64ec6c11f in release_3_4
* Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmapProject.class.php
+++ b/lizmap/modules/lizmap/classes/lizmapProject.class.php
@@ -9,6 +9,8 @@
  *
  * @license Mozilla Public License : http://www.mozilla.org/MPL/
  */
+use Lizmap\Project\ProjectMetadata;
+
 class lizmapProject extends qgisProject
 {
     /**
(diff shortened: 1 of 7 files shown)


lizmap/modules/lizmap/classes/lizmapProject.class.php
```diff
@@ -9,6 +9,8 @@
  *
  * @license Mozilla Public License : http://www.mozilla.org/MPL/
  */
+use Lizmap\Project\ProjectMetadata;
```
mdouchin, line 12: @laurentj I needed to add this `use` here which I did not need in master . Do you know why ?

rldhont (approved):

rldhont:
waiting for @laurentj  review ?

lizmap/modules/lizmap/classes/lizmapProject.class.php
laurentj: Because 
1. you created the file ProjectMetadata.php into the lizmap/lib/ directory, which is declared into the module.xml, as a directory which have classes having a namespace starting with `Lizmap\`, and so that are auto-loadable
2. in ProjectMetadata.php, you declared that the class is into the namespace `Lizmap\Project`

All is fine, you can keep as is.

laurentj (approved):
looks good to me

nboisteault (approved):

PR #2229, opened 2021-04-07
Docker: change base image tag

dmarteau:
## Change base image tag from alpine:3.11 to alpine:3

Strict alpine version is too restrictive: it prevents CI builds to take into accounts custom infrastructure choices. 

## Follow convention for php override config name.

Change the name of the custom lizmap override config file dynamically created at startup from the PHP_INI env variable to follow convention.

### DIFF
--- a/docker/lizmap-entrypoint.sh
+++ b/docker/lizmap-entrypoint.sh
@@ -9,7 +9,7 @@ LIZMAP_ADMIN_EMAIL=${LIZMAP_ADMIN_EMAIL:-root@local.localhost}
 
 # php ini override
 if [ ! -z $PHP_INI ]; then
-    echo -e "$PHP_INI" > $PHP_INI_DIR/conf.d/lizmap-php-0.ini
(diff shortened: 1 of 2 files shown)


docker/lizmap-entrypoint.sh
```diff
@@ -35,7 +35,7 @@ fi
 # Update localconfig and lizmapConfig
 update-config.php
 
-# Set up Configuration  
+/etc/php7/php.ini# Set up Configuration
```
laurentj, line 38: bad copy-paste ?

laurentj (changes requested):

docker/lizmap-entrypoint.sh
dmarteau: Arf, un doigt qui a ripé sur le clavier....

dmarteau: Fix pushed...

laurentj (approved):

PR #2230, opened 2021-04-07
Fix: tabs in anchored popup were not toggling

nboisteault:
* Ticket support 1389
* Funded by 3Liz


drillinP (approved):
Tested with the test project and its work for me

PR #2235, opened 2021-04-12
Fix filter support 1116

nboisteault:
* Ticket : support 1116


nboisteault:
@mdouchin I just need a simple code review, you can skip tests. Thanks.

PR #2242, opened 2021-04-14
New config parameter to enable metrics

laurentj:
The debugMode parameter activates metrics, but also the logging of all
http requests to qgis. This is not what we would want. A new configuration
parameter `metricsEnabled` allow to enable only metrics, and
debugMode is not anymore used for metrics.

* Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmap.class.php
+++ b/lizmap/modules/lizmap/classes/lizmap.class.php
@@ -433,6 +433,10 @@ class lizmap
      */
     public static function logMetric($label, $start = 'index')
     {
+        if (!self::getServices()->areMetricsEnabled()) {
+            return;
+        }
+
         // Choose from when to calculate time: index, request or given $start
         if ($start == 'index') {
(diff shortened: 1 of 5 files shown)


laurentj:
I forgot some unit tests on lizmapServices, and I need to fix lints...

laurentj:
Ready to be reviewed

rldhont:
@laurentj can you add a print screen ?

rldhont (approved):

rldhont:
> @laurentj can you add a print screen ?

Sorry, forgot my request. It's ok, no ui.

Gustry:
Side question, how do we set this new parameter ?
Any update in the provided template files in `lizmap/var/config/*.ini.php.dist` ?

laurentj:
> Side question, how do we set this new parameter ?

```
[services]

metricsEnabled = 1
```
into lizmapConfig.ini.php

> Any update in the provided template files in `lizmap/var/config/*.ini.php.dist` ?

No, it is a feature only for us.

PR #2243, opened 2021-04-15
Fix logger: no more page processing interruption when it fails

laurentj:
The call of the dao method getDistinctCounter may fail if the sqlite database is locked or else. The exception was not catched and stopped the processing of the page.

* Funded by 3Liz


rldhont (approved):

laurentj:
I will do a backport into 3.4 and 3.3. This is the same patch, but as the class name has been changed into master, I cannot use the same patch.

PR #2256, opened 2021-04-28
Fix potential issues when calling OGC requests with unexpected method

laurentj:
Even if the request type name is checked into service.classic.php, when we used one of the OGC requests object independently, it could be processed with the name of any methods of the object.

Only methods corresponding to the request should be able to be executed. This is why I choose the solution to prefix these methods, so no other methods can be executed.

This naming allows too to identify easily methods of request types.

* Funded by 3liz


laurentj:
I see that this changes could break some module like wps_module. I  should find a solution to not break them.

laurentj:
It's ok finally.

rldhont (approved):

PR #2270, opened 2021-05-04
Add support for Drop-In lizmap configuration

dmarteau:
## Support for drop-in configuration files in Lizmap docker image:

This PR add the capability to use drop-in config files that will merge to the lizmap configuration files at startup. It allows for customized local  persistent configuration which will be available at startup without the risk to be removed on container restart.

It allows for the container to provide its own set of configuration independently of the custom local configuration (modules configuration is a good exemple): it simplifies the the deployment of module configuration without  having to tamper with the main configuration files.

It use environment variables for defiining the directories to load files from. If defined, all files ending in `*.php.ini` are merged to the appropriate config file

## Changes

* Add `load_include_config` in `update-config.php`
* Use `LIZMAP_LOCALCONFIG_INCLUDE` env var for files merged to `localConfig.ini.php`
* Use `LIZMAP_LIZMAPCONFIG_INCLUDE` env var for files merged to `lizmapConfig.ini.php`

## Tests

Tested with [lizmap-docker-compose](https://github.com/3liz/lizmap-docker-compose)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -248,21 +248,6 @@ docker-release: check-factory
 docker-hub:
 	cd docker && $(FACTORY_SCRIPTS)/push-to-docker-hub.sh --clean
 
-LIZMAP_USER:=$(shell id -u)
-
-docker-run:
-	rm -rf $(shell pwd)/docker/.run
-	docker run -it --rm -p 9000:9000 \
-    -v $(shell pwd)/.run/config:/www/lizmap/var/config \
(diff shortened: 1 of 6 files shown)


laurentj (approved):
LGTM

laurentj:
These new environment variables are not documented. Could you document them into docker/README.md ?

dmarteau:
> These new environment variables are not documented. Could you document them into docker/README.md ?

Done

nboisteault (approved):

PR #2291, opened 2021-05-18
Docker: configure log metrics

dmarteau:
## Enable configuration for log metrics in docker

If `LIZMAP_LOGMETRICS` env var is set then use the value as metrics logger and enable metric logging.


nboisteault (approved):
As far as I understand it LGTM.

Issue #2280, opened 2021-05-15
Order of fields in filter not always the same


gerald2545:
### What is the bug?
in lizmap QGIS plugin, I set an order for the fields that can be used in filter.
Randomly, this order is not respected in LWC

### Steps to reproduce the issue
Create a postgis layer with some fields and add some records
define some fields in filter form
in LWC, display the filter form : the order is not always the same

### Expected behavior
in LWC, the order of the fields should be always the same

### Screenshots and logs
the reason is that for each field in the filter form a request is done against the database : the display order of the fields in the form corresponds to the SQL return order

### Environment

 - Lizmap version: 3.4.3
 - QGIS Server FULL version: 3.10
 - QGIS Desktop FULL version: 3.16.5
 - OS (Windows, Linux, MacOS, Android…): Docker sur linux
 - Browser (Firefox, Chrome…):Firefox
 - Lizmap admin information (users, logs…) stored into Sqlite

PR #2296, opened 2021-05-21
fix the order of the filter fields in the form

gerald2545:
Fix the order of the fields in the filter form, in case of a postgis layer (before, was dependant to the SQL response order)
* Ticket : #2280
* Funded by gerald2545

### DIFF
--- a/lizmap/www/assets/js/filter.js
+++ b/lizmap/www/assets/js/filter.js
@@ -140,19 +140,20 @@ var lizLayerFilterTool = function() {
                 var field_item = filterConfig[o];
                 if( 'title' in field_item && field_item.layerId == layerId ){
                     formFilterLayersSorted.push(field_item);
+                    $("#filter div.tree").append('<div id="filter-field-order'+String(field_item.order)+'"></div>');
                 }
             }
 
             // Add form fields
             for(var conf in formFilterLayersSorted){
-                var field_item = formFilterLayersSorted[conf];
-                getFormFieldInput(field_item);
+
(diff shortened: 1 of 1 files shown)


rldhont (approved):

nboisteault:
> in case of a postgis layer (before, was dependant to the SQL response order)

@gerald2545 Thanks for your PR. What do you mean in this quote? Before which change it was dependant to the SQL response order?

gerald2545:
> 
> 
> > in case of a postgis layer (before, was dependant to the SQL response order)
> 
> @gerald2545 Thanks for your PR. What do you mean in this quote? Before which change it was dependant to the SQL response order?

You're welcome.  
the order of the fields in the filter form was random. With this PR, the order of the fields is always the same (teh one defined in the lizmap QGIS plugin), whatever the time used by each SQL request to create the field contents (min/max in the slider, uniq values in the list...).

nboisteault (approved):

nboisteault:
@gerald2545 Good work, I merge.
Note for the future, code part creating every filter needs to be refactored. We should use `Promise.all()` in order to fetch every query for every filter then add responses to the DOM at once when all received.

rldhont:
Backport done:
* 3.4 https://github.com/3liz/lizmap-web-client/commit/62f3ceee34c8750a38dbdda87d33af41729ecff8
* 3.3 https://github.com/3liz/lizmap-web-client/commit/1ca3000abce5115e5c4853ebe3f97a7739b3b143

PR #2336, opened 2021-06-11
Fix: get non-exposed fields via WFS w/ QGIS 3.16

### LINKED ISSUE #2304: Do not expose via WFS is not tacking into account in Form
### What is the bug?
The checkbox **Do not expose via WFS** in the Vector layer properties in fields tab seems to be not taking into account in the edition form.

1. Go to **Fields** tab in the **Vector layer properties** in QGIS Desktop
2. Click on **Do not expose via WFS** in Configuration column
3. Publish the project to lizmap
4. Add an element to the layer
5. See that the field is in the form

### Expected behavior
Like with QGIS 3.10, the checkbox **Do not expose via WFS** has to be tacking into account in the edition form.

### Environment

 - Lizmap version: 3.4.*
 - QGIS Server FULL version: 3.16.*
 - QGIS Desktop FULL version: 3.16.*

nboisteault:
QGIS 3.16 introduce a new way to declare non-exposed fields in qgs

* Ticket : Fix #2304 
* Funded by 3Liz

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisProject.class.php
+++ b/lizmap/modules/lizmap/classes/qgisProject.class.php
@@ -798,7 +798,14 @@ class qgisProject
                     $layer['constraints'] = $constraints;
                     $layer['wfsFields'] = $wfsFields;
 
-                    $excludeFields = $xmlLayer->xpath('.//excludeAttributesWFS/attribute');
+                    // Do not expose fields with HideFromWfs parameter
+                    // Format in .qgs has changed in QGIS 3.16
+                    if ($this->qgisProjectVersion >= 31600) {
+                        $excludeFields = $xmlLayer->xpath('.//field[contains(@configurationFlags,"HideFromWfs")]/@name');
+                    } else {
+                        $excludeFields = $xmlLayer->xpath('.//excludeAttributesWFS/attribute');
+                    }
+
                     if ($excludeFields && count($excludeFields) > 0) {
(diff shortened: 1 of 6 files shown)


nboisteault:
@laurentj @rldhont 

1. Should I look and test for QGIS Server version? If yes, where do I get this value?

2. I could add a new end2end test for this case but maybe PHP unit test is a better candidate. Your thoughts?

Gustry:
Hum, sorry, I knew this change when I saw the PR in QGIS, but  I didn't thought that LWC will parse the XML for this setting.
Not only for WFS, but also for WMS. Did you check ?

> Should I look and test for QGIS Server version?

Not strong opinion, but it would be nice indeed. A little bit of milliseconds less to parse the XML ?
Also to have a cleaner code abotu new/legacy API.

>  If yes, where do I get this value?

For the QGIS Project version, you have it :
* at the top of the XML file. It might be already stored in the project in an attribute ?
* In the CFG file, in the metadata section at the top of the file

> I could add a new end2end test for this case but maybe PHP unit test is a better candidate. Your thoughts?

PHP unittest is indeed better indeed. It's just a parsing issue. Given a XML string and check you have expected fields ?

rldhont:
@nboisteault you can use `$this->getQgisProjectVersion() >= 31600` or `$this->qgisProjectVersion >= 31600` to avoid unnecessary `xpath` request.

rldhont (approved):

Issue #2327, opened 2021-06-04
Notice: WMSExtent could be not defined in QGIS Project properties


rldhont:
### What is the bug?
```
Trying to access array offset on value of type null
  /srv/lizmap_web_client-3.4.3/lizmap/modules/lizmap/classes/qgisProject.class.php	500
Trying to access array offset on value of type null
  /srv/lizmap_web_client-3.4.3/lizmap/modules/lizmap/classes/qgisProject.class.php	501
Trying to access array offset on value of type null
  /srv/lizmap_web_client-3.4.3/lizmap/modules/lizmap/classes/qgisProject.class.php	502
Trying to access array offset on value of type null
  /srv/lizmap_web_client-3.4.3/lizmap/modules/lizmap/classes/qgisProject.class.php	503
```

These notices are for
```php
            $WMSExtent = $qgsLoad->properties->WMSExtent->value[0];
            $WMSExtent = $qgsLoad->properties->WMSExtent->value[1];
            $WMSExtent = $qgsLoad->properties->WMSExtent->value[2];
            $WMSExtent = $qgsLoad->properties->WMSExtent->value[3];
```

This mean that the `WMSExtent` element is not defined in the QGIS Project `properties` element

And `$WMSExtent` can stay an empty string because it is only use in template to be deisplay.

PR #2338, opened 2021-06-14
Check if wmsextent is not null

drillinP:
* Ticket : #2327 
* Adding check to test if wmsextent is not null

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisProject.class.php
+++ b/lizmap/modules/lizmap/classes/qgisProject.class.php
@@ -499,10 +499,12 @@ class qgisProject
             }
             $WMSKeywordList = implode(', ', $values);
 
-            $WMSExtent = $qgsLoad->properties->WMSExtent->value[0];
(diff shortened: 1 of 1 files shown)


lizmap/modules/lizmap/classes/qgisProject.class.php
```diff
@@ -499,10 +499,12 @@ protected function readWMSInformation($qgsLoad)
             }
             $WMSKeywordList = implode(', ', $values);
 
-            $WMSExtent = $qgsLoad->properties->WMSExtent->value[0];
-            $WMSExtent .= ', '.$qgsLoad->properties->WMSExtent->value[1];
-            $WMSExtent .= ', '.$qgsLoad->properties->WMSExtent->value[2];
-            $WMSExtent .= ', '.$qgsLoad->properties->WMSExtent->value[3];
+            if(!is_null($WMSExtent)){
```
rldhont, line 502: It is `$qgsLoad->properties->WMSExtent` to test!

rldhont (changes requested):

rldhont (approved):

PR #2343, opened 2021-06-14
[BUGFIX] Print - Print the map in the project projection, not in LWC map projection

mdouchin:
This avoid for example to print a map in Pseudo Mercator (EPSG:3857)
when there are some external baselayers such as OpenStreetMap in the Lizmap configuration

This fix allows to print in the real scale: you can use your ruler 📏 !

* Funded by 3liz


rldhont (approved):

lizmap/www/assets/js/map.js
```diff
@@ -3527,6 +3527,18 @@ var lizMap = function() {
...
     var w = size.width / 72 / unitsRatio * aScale / 2;
     var h = size.height / 72 / unitsRatio * aScale / 2;
+    var printInProjectProjection = true;
+    if ('printInProjectProjection' in lizMap.config.options && lizMap.config.options.printInProjectProjection != 'True') {
+      printInProjectProjection = false;
+    }
+    if (printInProjectProjection
+      && lizMap.config.options.qgisProjectProjection.ref != lizMap.config.options.projection.ref
+    ) {
+      var printDpi = parseFloat(document.querySelector('#print-dpi').value);
```
nboisteault, line 3537: `printDpi` is never used.

lizmap/www/assets/js/map.js
```diff
@@ -3554,6 +3566,18 @@ var lizMap = function() {
...
     var w = size.width / 72 / unitsRatio * aScale;
     var h = size.height / 72 / unitsRatio * aScale;
+    var printInProjectProjection = true;
+    if ('printInProjectProjection' in lizMap.config.options && lizMap.config.options.printInProjectProjection != 'True') {
+      printInProjectProjection = false;
+    }
+    if (printInProjectProjection
+      && lizMap.config.options.qgisProjectProjection.ref != lizMap.config.options.projection.ref
+    ) {
+      var printDpi = parseFloat(document.querySelector('#print-dpi').value);
```
nboisteault, line 3576: `printDpi` is never used.

lizmap/www/assets/js/map.js
```diff
@@ -3975,20 +4039,56 @@ var lizMap = function() {
...
+            draw_feature.geometry.transform(map.projection, project_projection);
+          }
+          highlightGeom.push(formatWKT.write(draw_feature));
           highlightSymbol.push(lizMap.mainLizmap.digitizing.getFeatureDrawnSLD(index));
 
         }
+      }
+      // Print the chosen drawn rectangle in dashed pale white
+      var printDrawBox = false;
+      if ('printDrawBox' in lizMap.config.options && lizMap.config.options.printDrawBox == 'True') {
```
nboisteault, line 4061: Where `lizMap.config.options.printDrawBox` is set?

lizmap/www/assets/js/map.js
```diff
@@ -3527,6 +3527,18 @@ var lizMap = function() {
     var unitsRatio = OpenLayers.INCHES_PER_UNIT[units];
     var w = size.width / 72 / unitsRatio * aScale / 2;
     var h = size.height / 72 / unitsRatio * aScale / 2;
+    var printInProjectProjection = true;
+    if ('printInProjectProjection' in lizMap.config.options && lizMap.config.options.printInProjectProjection != 'True') {
```
nboisteault, line 3531: Where `lizMap.config.options.printInProjectProjection` is set?

nboisteault (changes requested):

lizmap/www/assets/js/map.js
mdouchin: Could not test if we needed to use the hardcoded DPI 96 or the value of the chosen DPI, since I encountered a bug with QGIS Server 3.16 which does not take the DPI parameter into account (exported PDF or images have the same size)

lizmap/www/assets/js/map.js
mdouchin: It was added there so that an external JS script could set this property for debugging purpose. Not sure if we need to keep it ? If not, we need to remove the corresponding code @rldhont any thought ?

mdouchin:
@rldhont @nboisteault I simplified the code. I removed the `printDrawBox` option (and kept the code in a local branch) and remove the unnecessary listener on `print-dpi` change since the draw box does not depend on it

mdouchin:
The default behaviour in 3.5 will be "print in project projection". For the backport in 3.4, we could use the other default "do as before" but let the possibility to change the behaviour by using the JS console of a browser and set

```javascript
lizMap.config.options.printInProjectProjection = 'True'
```

lizmap/www/assets/js/map.js
Gustry: Does a eslint checks would help here ? to clean step by step the LWC code base ?
Maybe enabling eslint on only modified file in a commit ?

mdouchin: Sure, but here it was made on purpose, to let dev who review test different configurations. I removed it after discussion and further tests yesterday by force pushing

nboisteault (approved):
Ok for Lizmap 3.5. Just fill CHANGELOG-3.5.md here and CHANGELOG-3.4.md in backport to clearly explain what have changed.

PR #2345, opened 2021-06-16
Action - Improvements in Lizmap Web Client action module

mdouchin:
* Remove useless JS code which sent the action `options` parameters (they are retrieved safely by the PHP back-end)
* When the result GeoJSON contains a field `message`, it is displayed in Lizmap message bar at the same time the received geometry is drawn
* A new action property `confirm` allows the editor to have a confirm message displayed with its text content before performing the action
* A new Lizmap event `actionResultReceived` is triggered when the action result is received with the action properties and the returned GeoJSON feature
* Documentation has been improved.

* Funded by SNCF Réseau Grand-Est

### DIFF
--- a/lizmap/www/assets/js/action.js
+++ b/lizmap/www/assets/js/action.js
@@ -32,49 +32,42 @@ function getItemByName(layerId, name){
 
 function getActionData(layerId, fid, name){
 
-    // Empty actionLayer
-    var layer = lizMap.layers['actionLayer'];
-    layer.destroyFeatures();
-
-    // Do nothing if geometry was already set
-    // This allow to delete the current geometry
-    if(
-        action_current_object == layerId + '.' + fid + '.' + name
-    ){
-        //console.log("no api call");
-        action_current_object = null;
-        return true;
-    }
-
     var options = {
         "layerId": layerId,
         "featureId": fid,
         "name": name
     };
     var item = getItemByName(layerId, name);
-    //console.log(item);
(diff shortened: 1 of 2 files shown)


mdouchin:
@rldhont @nboisteault I would like to add this in 3.5

mdouchin:
I modified Lizmap Documentation: 
* https://github.com/3liz/lizmap-documentation/commit/8bf1c61e1b8b14cf1ae2a67445899d2a06d1b966
* https://github.com/3liz/lizmap-documentation/commit/96e8f8d7a5905fe110979f3b689c6f7725296255

with also this GIF

publish-configuration-action-popup

nboisteault (approved):
Just a detail for later code, we should all use `let` and `const` instead of `var`.

lizmap/www/assets/js/action.js
```diff
@@ -159,27 +164,56 @@ function addActionButton(layerId, fid, item, popupitem){
...
+
+        // Empty actionLayer: do it if button & action was active or not
+        var layer = lizMap.layers['actionLayer'];
+        layer.destroyFeatures();
+
         var val = $(this).val();
         var vals = val.split('.');
         var layerId = vals[0];
         var fid = vals[1];
         var name = vals[2];
```
nboisteault, line 193: Tip: since ES6 you could replace by `const [layerId, fid, name] = vals;` 
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

PR #2347, opened 2021-06-17
[BUGFIX][BACKPORT FROM 3.5][DEACTIVATED] Print - Print the map in the project projection, not in LWC map projection

mdouchin:
This fix has been backported from 3.5. It is not activated by default in 3.4.4
You need to set the variable
`lizMap.config.options.printInProjectProjection = 'True'`
in your web browser JS console to test this new behaviour


mdouchin:
I have not modified the CHANGELOG 3.4 as this change is "hidden" and not active by default

mdouchin:
@rldhont @nboisteault Please tell me if the title of this PR (and related commit) is not too much bazooka

nboisteault (approved):

PR #2359, opened 2021-06-24
Fix dataviz notice

nboisteault:

### DIFF
--- a/lizmap/modules/dataviz/classes/datavizPlot.class.php
+++ b/lizmap/modules/dataviz/classes/datavizPlot.class.php
@@ -122,17 +122,19 @@ class datavizPlot
         if (!empty($plotConfig['plot']['x_field'])) {
             $x_fields[] = $plotConfig['plot']['x_field'];
         }
-        $traces = $plotConfig['plot']['traces'];
-        foreach ($traces as $trace) {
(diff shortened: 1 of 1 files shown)


lizmap/modules/dataviz/classes/datavizPlot.class.php
```diff
@@ -136,53 +136,6 @@ private function parsePlotConfig($plotConfig)
                     $z_fields[] = $trace->z_field;
                 }
             }
-        } else {
-            // LEGACY CODE:  LIZMAP < 3.4
```
Gustry, line 140: We need to keep that to read CFG file made with < 3.4.
Not all CFG are updated. Otherwise, we need from our users to open **all** projects on a server, open Lizmap and save lizmap. They are plenty of projects on a server.

nboisteault: I see. I'll handle that.

Gustry: To remove this kind of code, you need to provide a table with all Lizmap projects and their version.

For now, there is no way to know, in a long list of lizmap projects sitting on a server, each LWC target version.

PR #2370, opened 2021-07-05
Resurrect lizmapProxy and lizmapOGCRequest classes for modules compatibility

rldhont:
Some modules, like cadastre and wps client, need lizmapProxy and lizmapOGCRequest.
These classes has been removed and replaced by namespace classes. To keep compatibility,
These classes has been resurrected to extends the new classes.

### DIFF
--- a/CHANGELOG-3.5.md
+++ b/CHANGELOG-3.5.md
@@ -37,5 +37,6 @@
 ### Backend
 
 - Major refactoring of Lizmap source code done by @alagroy-42
+- Keep lizmapProxy and lizmapOGCRequest classes for modules compatibility
 - Upgrade Jelix to 1.6.32
 - Improve testings using Docker, Cypress, PHPUnit etc
--- a/lizmap/modules/lizmap/classes/lizmapOGCRequest.class.php
+++ b/lizmap/modules/lizmap/classes/lizmapOGCRequest.class.php
@@ -1,0 +1,40 @@
+<?php
+/**
+ * Manage OGC request.
+ *
+ * @author    3liz
+ * @copyright 2015-2021 3liz
+ *
+ * @see      http://3liz.com
+ *
+ * @license Mozilla Public License : http://www.mozilla.org/MPL/
+ */
+/**
+ * dummy class for compatibility.
+ *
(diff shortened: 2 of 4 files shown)


CHANGELOG-3.5.md
```diff
@@ -1,5 +1,9 @@
 # Changelog Lizmap 3.5
 
+## Unreleased
```
nboisteault, line 3: Why this section?

rldhont: To prepare next release, the section Unreleased can contain the changelog of the next release.

Gustry: For now, we have one changelog file per version @rldhont so this change is not really needed.

It's not the best because difficult to reuse the content of these files. It's not following the convention https://keepachangelog.com/en/1.0.0/ but this is another topic.

nboisteault: Your PR is for 3.5 so you should put description in 3.5 part.

rldhont: Sorry I was thinking 3.5.0 has been already released, I was fixing stable release.

rldhont: I have updated the changelog

lizmap/modules/lizmap/classes/lizmapOGCRequest.class.php
```diff
@@ -0,0 +1,41 @@
...
+    /**
+     * constructor.
+     *
+     * @param lizmapProject $project    the project has a lizmapProject Class
+     * @param array         $params     the OGC request parameters array
+     * @param string        $requestXml the OGC XML Request as string
+     */
+    public function __construct($project, $params, $requestXml = null)
+    {
+        //print_r( $project != null );
```
nboisteault, line 29: This can be removed. For the rest LGTM. Test with module-cadastre and it'll be ok.

nboisteault (approved):

lizmap/modules/lizmap/classes/lizmapOGCRequest.class.php
rldhont: fixed

PR #2377, opened 2021-07-08
Fix: Add hash in custom assets querystring

### LINKED ISSUE #2373: [Bug]: custom JS and CSS and cache refresh
### What is the bug?

When modifying a custom JS or CSS file for a project, modified files are not loaded until cache is refreshed (ctrl + F5).
Possible solution: file hash could be added as a URL parameter here : https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/view/controllers/lizMap.classic.php#L397
PHP has this function : https://www.php.net/manual/en/function.hash-file.php

1. Modify a custom asset, JS or CSS.
2. Reload (F5) page
3. Old asset version is loaded not new one

### Lizmap version

3.4.4

### QGIS desktop version

all

### QGIS server version

all

### Operating system

all

### Browsers

Firefox, Chrome, Safari, Microsoft Edge

### Browsers version

all

### Relevant log output

_No response_

nboisteault:
Add file modification time in custom assets querystring.
It busts cache and avoid user to force reload (ctrl + F5) to have fresh assets.

* Fix #2373 
* Funded by 3Liz

### DIFF
--- a/lizmap/modules/view/controllers/lizMap.classic.php
+++ b/lizmap/modules/view/controllers/lizMap.classic.php
@@ -404,6 +404,7 @@ class lizMapCtrl extends jController
                                     array(
                                         'repository' => $lrep->getKey(),
                                         'project' => $project,
+                                        'mtime' => filemtime($filename),
                                         'path' => $jsRelPath,
                                     )
                                 );


laurentj:
IIMHO, using hash_file is not the best solution. It reads the entire file. It means the file is readed twice to be displayed into the web page : one time with hash_file, the second time during the call of the getmedia url. 

It could be ok on a single instance with tiny JS/CSS files and used by only some users, but it can be problematic on a server with big JS/CSS files and/or with many users. 

I propose to use the modification time of the file (hashed with md5?) instead the hash of its content.

PR #2382, opened 2021-07-16
Fix some log messages about XML loading

rldhont:
Firstly fix `Object of class LibXMLError could not be converted to string` in OGCRequest.

Secondly add error log messages about QGIS Project XML loading

Then add error log messages about WMS GetCapabilities loading in WMTS classes.

* Funded by CRIGE-PACA

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmapTiler.class.php
+++ b/lizmap/modules/lizmap/classes/lizmapTiler.class.php
@@ -80,15 +80,30 @@ class lizmapTiler
             $wmsResult = $wmsRequest->process();
             // Http code error
             if (($wmsResult->code / 100) >= 4) {
+                $errormsg = 'An error has been raised when loading WMS GetCapabilities: HTTP Code '.$wmsResult->code;
+                \jLog::log($errormsg, 'error');
+
                 return null;
             }
             $wms = $wmsResult->data;
             // empty data or service exception, WMS not available
             if (empty($wms) or preg_match('/ServiceExceptionReport/', $wms)) {
+                $errormsg = 'An error has been raised when loading WMS GetCapabilities: ServiceExceptionReport';
+                $errormsg .= '\n'.$wms;
+                \jLog::log($errormsg, 'error');
+
                 return null;
             }
 
-            $wms_xml = simplexml_load_string($wms);
+            $wms_xml = \Lizmap\App\XmlTools::xmlFromString($wms);
+            if (!is_object($wms_xml)) {
+                $errormsg = '\n'.$file.'\n'.$wms_xml;
+                $errormsg = 'An error has been raised when loading WMS GetCapabilities:'.$errormsg;
+                \jLog::log($errormsg, 'error');
+
+                return null;
+            }
+
             $wms_xml->registerXPathNamespace('wms', 'http://www.opengis.net/wms');
             $wms_xml->registerXPathNamespace('xlink', 'http://www.w3.org/1999/xlink');
 
@@ -109,6 +124,7 @@ class lizmapTiler
             'tileMatrixSetList' => null,
             'layerTileInfoList' => null,
         );
+
         if (is_array($tileMatrixSetList) && is_array($layers)) {
             $tileCapabilities->tileMatrixSetList = $tileMatrixSetList;
             $tileCapabilities->layerTileInfoList = $layers;
--- a/lizmap/modules/lizmap/lib/App/XmlTools.php
+++ b/lizmap/modules/lizmap/lib/App/XmlTools.php
@@ -1,0 +1,95 @@
+<?php
+/**
+ * XML tools for Lizmap.
+ *
+ * @author    3liz
+ * @copyright 2021 3liz
+ *
+ * @see      https://3liz.com
+ *
+ * @license Mozilla Public License : http://www.mozilla.org/MPL/
+ */
+
+namespace Lizmap\App;
+
+class XmlTools
+{
+    /*
+     * Interprets a string of XML into an object
+     *
+     * @param string $xml_str a well-formed XML string
+     *
+     * @return SimpleXmlElement|string an object of class SimpleXMLElement with properties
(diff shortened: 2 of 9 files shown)


nboisteault (approved):
@rldhont it looks good to me but we have no project to replicate the issue and test your PR.

mdouchin (changes requested):
I think we should add a method in a utility class to manage the building of the `$errormsg` to avoid to duplicate four times the code (with small changes). Probably could be a method of `lizmap/modules.lizmap/lib/Request/Proxy.php` or in another more appropriate class.

lizmap/modules/lizmap/classes/lizmapTiler.class.php
```diff
@@ -80,15 +80,54 @@ public static function getTileCapabilities($project)
...
+                $errormsg = 'An error has been raised when loading WMS GetCapabilities: ServiceExceptionReport';
+                $errormsg .= '\n'.$wms;
+                \jLog::log($errormsg, 'error');
+
                 return null;
             }
 
+            $use_errors = libxml_use_internal_errors(true);
             $wms_xml = simplexml_load_string($wms);
+            if ($wms_xml === false) {
```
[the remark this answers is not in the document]
rldhont: Ok I'll do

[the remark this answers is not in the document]
rldhont: done

rldhont:
@nboisteault @mdouchin can you re-review the code ?

mdouchin (approved):
👍 thanks

nboisteault (approved):
Could you add some unit tests?

laurentj (changes requested):
These new methods are not related at all to Lizmap, and so should not be into the `lizmap` class. I would prefer to have them in the `tools` class or better, into a new `\Lizmap\App\XmlTools` class (into the lib directory of the lizmap module).

rldhont:
@laurentj thanks for information

rldhont:
@laurentj new class `\Lizmap\App\XmlTools` created with tests

@nboisteault  unit tests addded

lizmap/modules/lizmap/lib/App/XmlTools.php
```diff
@@ -0,0 +1,97 @@
+<?php
+/**
+ * app context informations inside a Jelix environment.
```
laurentj, line 3: wrong comment, and wrong copyright year ;-)

lizmap/modules/lizmap/lib/App/XmlTools.php
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+namespace Lizmap\App;
+
+class XmlTools
+{
+    /*
+     * Interprets a string of XML into an object
+     *
+     * @param $xml_str a well-formed XML string
```
laurentj, line 20: `string` is missing. The good syntax is `@param string $xml_str  a well-formed XML string`

lizmap/modules/lizmap/lib/App/XmlTools.php
```diff
@@ -0,0 +1,97 @@
...
+     *
+     * @return ($xml, $msg) Array containing $xml an object of class SimpleXMLElement with properties
```
laurentj, line 22: the good syntax here is : 

```
* @return array  Array containing a SimpleXMLElement object with properties containing the data held 
*                                  within the XML document, or false on failure ; and a string containing the error message.
```

lizmap/modules/lizmap/lib/App/XmlTools.php
```diff
@@ -0,0 +1,97 @@
...
+        $msg = '';
+        if (!$xml) {
+            $msg = self::xmlErrorMsg();
+        }
+
+        return array($xml, $msg);
+    }
+
+    /*
+     * Interprets an XML file into an object
```
laurentj, line 39: same issue in `@param` and `@return` syntax. See https://docs.phpdoc.org/latest/guide/guides/types.html

lizmap/modules/lizmap/lib/App/XmlTools.php
```diff
@@ -0,0 +1,97 @@
...
+
+        return array($xml, $msg);
+    }
+
+    /*
+     * Get XML error message
+     *
+     * Build an error message based on LibXMLError object
+     *
+     * @return $msg a string containing the error message.
```
laurentj, line 64: `@return string  the error message.`

lizmap/modules/lizmap/lib/Request/OGCRequest.php
```diff
@@ -312,23 +312,26 @@ protected function process_getcapabilities()
         );
     }
 
+    /*
+     * Interprets a string of XML into an object
+     *
+     * @param $xmldata a well-formed XML string
```
laurentj, line 318: `string` keyword is missing in both `@param`

laurentj (changes requested):
It's ok to me, except DOC comments. It is important to respect the right syntax, to allow IDE to help on auto-completion and type checks, and so to improve the code quality. It helps also to generate a good API documentation (for the day when we will generate such documentations).

Probably we could improve the return type of `xmlFromString` and `xmlFromFile` by returning only one value (`@return SimpleXmlElement|string`). So, instead to do `if (!$xml)`, we could do `if(is_string($xml))` or `if(!is_object($xml))`.  With this returned value, there is not a unused variable (which exists here in both case, error or not), and it makes the signature of the method more readable.

I let you to decide to do the change or to keep as is, but at least  change the comments :-)

Gustry:
> It is important to respect the right syntax, to allow IDE to help on auto-completion and type checks, and so to improve the code quality. It helps also to generate a good API documentation (for the day when we will generate such documentations).

Oh yes, big +1.
Can we revive #1755 in the mid term?

laurentj (approved):
LGTM. Thank you :-)

rldhont:
Thanks to all

PR #2390, opened 2021-07-22
[Tests] Enhance WKT parser tests

rldhont:
Use dedicated assertion instead of only assertTrue or assertFalse.


laurentj (approved):
Yes, it's better indeed :-)

Issue #2391, opened 2021-07-22
[Bug]: potential issue about wmsGetCapabilitiesUrl and wmtsGetCapabilitiesUrl?


laurentj:
### What is the bug?

In the code of the Project class, [we retrieve](https://github.com/3liz/lizmap-web-client/blob/69584d060124d4212468379e0fc8128f8d3e6d44/lizmap/modules/lizmap/lib/Project/Project.php#L1545) the `wmsGetCapabilitiesUrl` and the `wmtsGetCapabilitiesUrl` data from the `QgisProject` object.

However, into [QgisProject](https://github.com/3liz/lizmap-web-client/blob/69584d060124d4212468379e0fc8128f8d3e6d44/lizmap/modules/lizmap/lib/Project/QgisProject.php) there is not set of such properties into the `data` property.

Am I missing something ? Or is there a bug here, and we don't retrieve the value from the right object?

All these `getData` methods from Project, ProjectMetadata and QgisProject are a nightmare to know which data we can retrieve, what are their type etc... I'm preparing a patch to deprecate them and to replace them by some explicit methods for each data, and I discover this issue during my work.

### Lizmap version

3.5.0-rc.1

laurentj:
Same issue, few lines below those I pointed: there is a `$this->qgis->getData('label')` but nowhere in QgisProject a such value is set.

For the wmsGetCapabilitiesUrl and the wmtsGetCapabilitiesUrl data, I think it should be retrieved from the `Project` objet itself, as [there are set here](https://github.com/3liz/lizmap-web-client/blob/69584d060124d4212468379e0fc8128f8d3e6d44/lizmap/modules/lizmap/lib/Project/Project.php#L312).

laurentj:
Ok, i see that the `label`  should comes from the `Repository` object, not from the project object.

I will make a patch.

nboisteault:
@laurentj As you said it is a nightmare. I think this issue is related to https://github.com/3liz/lizmap-web-client/issues/2153.
I banged my head on it and ask @mdouchin for help but you can join the party if you want :)

PR #2394, opened 2021-07-22
Fix regressions in the Project class

laurentj:
* metadata panel has wrong WMS/WMTS url and wrong label
* in checkAclByUser(): options were not readed from the right object

* Ticket : #2391
* Funded by 3Liz

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/Project.php
+++ b/lizmap/modules/lizmap/lib/Project/Project.php
@@ -122,7 +122,7 @@ class Project
     protected $formFilterLayers = array();
 
     /**
-     * @var array
+     * @var object
      */
     protected $editionLayers = array();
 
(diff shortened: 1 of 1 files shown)


lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -122,7 +122,7 @@ class Project
     protected $formFilterLayers = array();
 
     /**
-     * @var array
+     * @var object
```
nboisteault, line 125: Why object ?

nboisteault (approved):

lizmap/modules/lizmap/lib/Project/Project.php
laurentj: Because it is. The content is loaded from the cfg file, with `json_decode` which returns array as objects since we don't pass the `JSON_OBJECT_AS_ARRAY` flag. And our code acts on this property as an object.

laurentj: Sorry, in fact, some code use this property as an array, and some other code use this property as an object. This is very strange :-/

rldhont (approved):

PR #2398, opened 2021-07-22
Replace getData() methods by getter methods

laurentj:
We have few objects having a `getData($key)` method. It returns a value corresponding to a key.

Having these kind of methods is not good for the code:
* we never know which keys value we can give
* we don't know the type of the return value
* as current `getData()` implementations return null if there is no value for the given key, we don't know if it is because the value does not exist or if it is because the key is invalid.
* we cannot specify the type of the return value in a doc comment for IDE, to help auto-completion and type checks

The result is a code that is not very understandable, not robust, not very maintainable.

These patches propose to implement methods for each "data" we want to get from the object, allowing to have documentation about returned value, allowing to know naturally the available data.

The `getData()` methods are not removed, and only marked as deprecated. Some modules may still use them. They will be removed in Lizmap 4.0.

* Funded by 3Liz


rldhont (approved):

lizmap/modules/lizmap/lib/Project/Repository.php
```diff
@@ -100,6 +100,44 @@ public function getKey()
...
+            return false;
+        }
+        if (is_bool($value)) {
+            return $value;
+        }
+        if (is_numeric($value)) {
+            return intval($value) > 0;
+        }
+        $strVal = strtolower($value);
+        if ($strVal === 'true' && $strVal === 't'
```
nboisteault, line 131: @laurentj it is `&&` not `||` ?

nboisteault (commented):
One comment for the moment but I'll test soon.

lizmap/modules/lizmap/lib/Project/Repository.php
rldhont: replaced by `in_array()`

rldhont (changes requested):

mdouchin (changes requested):

rldhont (approved):

rldhont:
@mdouchin your review is needed

mdouchin (approved):

PR #2402, opened 2021-07-26
Fix and add tests for advanced forms

nboisteault:
- Fix: boolean checked value for checkbox which was sent was incorrect

* Funded by 3Liz


nboisteault:
Note : some commits have to be manually backport to release_3_4.

PR #2458, opened 2021-08-31
[Bugfix] WMS GetFeatureInfo Request can return XML

rldhont:
After refactoring, Lizmap was only able to return HTML for WMS GetFeatureInfo Request.


Gustry (commented):
Can you move the line L338 `$rep = '';` down next to the `if` or remove it. It would be easier to debug and see if the variable was not initialized or used (when using an IDE).

rldhont:
@Gustry I cannot, it used L372 `$rep .= $this->gfiGmlToHtml($data, $configLayer);`

Gustry:
Indeed, my ctrl+f failed ;-)
Sorry for the noise

PR #2461, opened 2021-09-01
API - Factorize ACL checks about edition layers

laurentj:
The code to do some ACL checks for edition layers is duplicated at several place (including in at least the address module). This patch reorganize the code so we now have new methods on Project to manipulate edition layers. There is for example the method `canCurrentUserEdit` that could be used into the address module.

This patch improve also the memory usage of classes that have some properties like `cachedProperties` that could be `static`.

* Funded by 3Liz

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmapProject.class.php
+++ b/lizmap/modules/lizmap/classes/lizmapProject.class.php
@@ -241,12 +241,22 @@ class lizmapProject
 
     public function hasEditionLayers()
     {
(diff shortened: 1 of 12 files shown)


mdouchin (commented):
Reading the changes: seems OK. Need more time to check locally

rldhont (approved):

nboisteault (approved):
Code LGTM. Cypress ok but we need tests concerning ACL.

drillinP (approved):

PR #2472, opened 2021-09-02
Fix bug of boolean checkbox when inserting

drillinP:
* Fix bug checkbox in the form. On insertion when the checkbox is not checked, Lizmap considers it as unmodified. On this PR I modified the edit.classic.php file when saving in the database, if featureId is null we get all the controls else if not null we get only the modified controls.
* Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/controllers/edition.classic.php
+++ b/lizmap/modules/lizmap/controllers/edition.classic.php
@@ -716,8 +716,14 @@ class editionCtrl extends jController
         // And get returned primary key values
         $pkvals = null;
         if ($check) {
-            // Save to database with modified controls
-            $pkvals = $qgisForm->saveToDb($feature, $form->getModifiedControls());
+            // Check if featureId is null to get all controls or only modified controls
+            if ($this->featureId == null) {
+                // Save to database with all controls
+                $pkvals = $qgisForm->saveToDb($feature, $form->getControls());
+            } else {
+                // Save to database with modified controls
+                $pkvals = $qgisForm->saveToDb($feature, $form->getModifiedControls());
+            }
         }
 
         // Some errors where encoutered


nboisteault (changes requested):
@drillinP Could you add comments on what is `getModifiedControls` and why it has been created?

rldhont (commented):
Can you add Docstring and tests ?

lizmap/modules/lizmap/lib/Form/QgisForm.php
```diff
@@ -1031,6 +1031,22 @@ public function deleteFromDb($feature)
...
+    public function getModifiedControls()
+    {
+        $modifiedControls = $this->getForm()->getModifiedControls();
+        $form = $this->form;
+        foreach ($this->formControls as $fieldName => $formControl) {
+            if (($this->formControls[$fieldName]->fieldEditType === 7
+            || $this->formControls[$fieldName]->fieldEditType === 'CheckBox')
+            && $this->formControls[$fieldName]->fieldDataType === 'boolean'
+            && !array_key_exists($fieldName, $modifiedControls)) {
+                $modifiedControls[$fieldName] = $this->getForm()->getControl($fieldName);
```
laurentj, line 1043: You add the control into the modifiedControls list, without checking if it's state has been modified?

drillinP: yes because if checkbox is not checked the value is `null` and not `false`

laurentj: `getModifiedControls` returns controls for which value has been changed. So you should add your checkbox into $modifiedControls **only** if it's "value" has changed, so only if its state has changed (checked -> unchecked or unchecked -> checked).

And probably the issue has to be fixed into the `getModifiedControls` method. Is there a simple test case somewhere? (and/or a ticket?)

rldhont: @laurentj here is the use case:
* boolean field in datatable with not null constraint but no default value defined
* the edit widget for the field is a checkbox without default value defined
* when the form is build for creation the checkbox is unchecked
* when the form is send for creation the checkbox can be still unchecked
* in lizmap the code set the default values to get field with default values in the modified controls list, but here the field has no default value
* the checkbox will not be in the modified control list if the status checked/unchecked has not chenged but lizmap use this list to only set modified fields even if in creation and needs checkbox value even if the status has not changed.
* In this case, the creation failed because no values has been set for the boolean not null without default value with a checkbox control.

To fix the issue, we can:
* define a new method for QgisForm - rename this `getModifiedControls` by something like `getControlsForSave`
* remove this method and update `saveToDb` to get checkbox values in addition to modifiedControls

laurentj: Ok I see the point. However, when inserting a new feature (a new record), you should not use 'initModifiedControls' and 'getModifiedControls', as it is to know modified values for an **existing** record. You should use them only if `$this->featureId` has a value, into `saveFeature`. In the case of a creation, you should use all controls of the form to save data.

laurentj (approved):

rldhont (approved):
Can you merge the commits and keep the last commit description ? Thanks

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -728,7 +728,12 @@ public function saveFeature()
         $pkvals = null;
         if ($check) {
             // Save to database with modified controls
```
nboisteault, line 730: @drillinP this comment should be moved.

nboisteault (approved):
@drillinP I have tested locally and it works as expected. Good job!

Gustry:
I didn't follow everything, but does it fix the manual test about checkbox in the form?
If yes, can you remove the notes we did about these failing tests ?

nboisteault:
@Gustry @drillinP I'll remove the manual test and replace it by a Cypress one when merged.

drillinP:
@laurentj can I merge the pr?

PR #2473, opened 2021-09-03
Command wmts:cleaning - to clear cache with command line

rldhont:
Ths Lizmap WMTS command line only contains methods to create cache and nothing to clear it.
The commit add a new tool to clean cache.

### DIFF
--- a/lizmap/modules/lizmap/lib/Request/Proxy.php
+++ b/lizmap/modules/lizmap/lib/Request/Proxy.php
@@ -646,6 +646,54 @@ class Proxy
         return false;
     }
 
+    /**
+     * @param string $repository
+     * @param string $project
+     *
+     * @return boolean the project cache has been clear or not
+     */
+    public static function clearProjectCache($repository, $project)
+    {
+        // Storage type
+        $ser = self::getServices();
+        $appContext = self::getAppContext();
+        $cacheStorageType = $ser->cacheStorageType;
+        $clearCacheOk = false;
+
+        // Cache root directory
+        if ($cacheStorageType != 'redis') {
+            $cacheRootDirectory = $ser->cacheRootDirectory;
+            if (!is_dir($cacheRootDirectory) or !is_writable($cacheRootDirectory)) {
+                $cacheRootDirectory = sys_get_temp_dir();
+            }
+
+            // Directory where cached files are stored for the project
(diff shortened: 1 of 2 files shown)


laurentj:
Why using FakeServerConf? What is the issue without it? FakeServerConf should be used only for unit tests.

rldhont:
Because I was using it in this control. Because I think it was necessary.

lizmap/modules/lizmap/lib/Request/Proxy.php
```diff
@@ -690,6 +702,8 @@ public static function clearLayerCache($repository, $project, $layer)
             $appContext->flushCache($cacheName);
         }
         $appContext->eventNotify('lizmapProxyClearLayerCache', array('repository' => $repository, 'project' => $project, 'layer' => $layer));
+
+        return $layer;
```
laurentj, line 706: Why returning the layer id? just return true.

lizmap/modules/lizmap/lib/Request/Proxy.php
```diff
@@ -675,13 +682,18 @@ public static function clearLayerCache($repository, $project, $layer)
                     }
                 }
                 closedir($handle);
+                if (count($results) === 0) {
+                    return false;
```
laurentj, line 686: Since we want to remove cache, having no existing cache is not a failure IMHO. It just says that we don't have to remove anything. So you should return true here.

you should return false if opendir fails for example.

lizmap/modules/lizmap/lib/Request/Proxy.php
```diff
@@ -675,13 +682,18 @@ public static function clearLayerCache($repository, $project, $layer)
...
+                }
                 foreach ($results as $rem) {
                     if (is_dir($rem)) {
                         \jFile::removeDir($rem);
                     } else {
                         unlink($rem);
                     }
                 }
+            } else {
+                return false;
```
laurentj, line 696: same here, you should return true.

laurentj (changes requested):
One more thing: you remove the cache of tiles. Should the tile capabilites cache be removed too?

drillinP (approved):

laurentj (approved):

drillinP:
Can we merge ?

PR #2479, opened 2021-09-07
Cleanup of Project and ProjectConfig classes

laurentj:
After the refactoring, some properties of QgisProject or ProjectConfig were still duplicated into the Project class. The goal of this patch is to remove them, allowing to save bytes and to have a consistent code.

Improved also the ProjectConfig class, to accept only data in its constructor, instead of having a filename and optional data. It eases tests and this is the responsability to Project to load data from the cfg file if there is no cache.

* Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmapProject.class.php
+++ b/lizmap/modules/lizmap/classes/lizmapProject.class.php
@@ -25,8 +25,6 @@ class lizmapProject
      */
     protected $proj;
 
-    const CACHE_FORMAT_VERSION = 1;
-
     /**
      * constructor.
      *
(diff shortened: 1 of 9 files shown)


nboisteault (approved):
@laurentj It is ok for me. Could you manually run Cypress in Github for this branch?
For this click 'Actions' tab in your Github, then 'Cypress' in Workflows column, Run workflow button, then you select this branch, and click 'Run workflow'

image

laurentj:
There is not my branch in the list on which cypress can be runned. I'll try locally.

nboisteault:
Cypress tests pass. There is just two tests not passing I fixed in master. So it is ok for me.

rldhont (approved):

Issue #2490, opened 2021-09-10
[Bug]: LWC doesn't display title defined in QGIS Server tab but file name


nboisteault:
### What is the bug?

LWC doesn't display title defined in QGIS Server tab but instead file name since https://github.com/3liz/lizmap-web-client/commit/a74935e5fb8eabb94dc1b57d7161dbbc57649161

### Steps to reproduce the issue

Display index.php/view/ page, Montpellier project should display "Montpellier - Transports" not "Montpellier".

### Lizmap version

master

### QGIS desktop version

all

### QGIS server version

all

### Operating system

all

### Browsers

Firefox, Chrome, Safari, Microsoft Edge

### Browsers version

all

nboisteault:
@laurentj https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/lib/Project/Project.php#L396
this should be `return $this->qgis->getTitle();` as `return $this->qgis->getKeywordList();` above in code.

PR #2491, opened 2021-09-10
Fix Project: remove latest duplicated properties

laurentj:
After the big refactorization of 2020, some properties were still duplicated between Project and QgisProject but not always with same values. It causes some unexpected results for some properties like when getting the title or the abstract.

This patch remove latest duplicated properties from Project to avoid confusion and to have lighter cache content.

* Ticket : #2490
* Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/Project.php
+++ b/lizmap/modules/lizmap/lib/Project/Project.php
@@ -30,18 +30,6 @@ class Project
     protected $cfg;
 
     /**
-     * @var array services properties
-     */
-    protected $properties = array(
-        'repository',
-        'id',
-        'title',
-        'abstract',
-        'proj',
-        'bbox',
-    );
-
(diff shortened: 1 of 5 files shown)


rldhont (approved):
LGTM

lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -403,7 +340,9 @@ public function getTitle()
      */
     public function getAbstract()
     {
-        return $this->getData('abstract');
+        $abstract = $this->qgis->getAbstract();
+
+        return $abstract ?: '';
```
nboisteault, line 345: `?:` TIL :)

nboisteault (approved):
Tested with Cypress.

drillinP (approved):

Issue #2231, opened 2021-04-07
Serial primary key column sorted as text in attribute table tool


RobiFag:
### What is the bug?
I have a PostGIS table (point) with a serial primary key column and when I sort the column from the attribute table tool , values are sorted as text.  Other numeric columns of different types (es. float) are sorted correctly.

### Screenshots and logs
![image](https://user-images.githubusercontent.com/25825472/113881406-b1595680-97bc-11eb-9a91-f0da78f7a8a0.png)

### Environment
 - Lizmap version: 3.4.2
 - QGIS Server FULL version: 3.16.5-Hannover
 - PostgreSQL: 13.2
 - PostGIS: 3.1.1

gerald2545:
same problem for me, even for column defined as integer : 
![image](https://user-images.githubusercontent.com/10177830/121346715-be163a00-c926-11eb-83fe-55acee5a819d.png)

josemvm:
i confirm 👎

josemvm:
maybe a qgis server issue...
on qgis server 3.10 it works fine!
@rldhont, please, can you do something about it?
thanks

josemvm:
it's lwc regression from here https://github.com/3liz/lizmap-web-client/pull/1165

nboisteault:
@josemvm @gerald2545 Could you please give us postgresql and postgis versions used?

nboisteault:
With QGIS Server 3.16,

I can **not** confirm issue with PostgreSQL 11.12 and Postgis : 2.5.5
But I can confirm issue with PostgreSQL 13.4 and Postgis : 3.1.3

Now looking for the bug...

nboisteault:
@mdouchin I'll need some help on this one.

josemvm:
@nboisteault here i'm on postgresql 13.2 and postgis 3.1

josemvm:
LWC 3.3 works fine with these versions: postgresql 13.2 and postgis 3.1

josemvm:
partially related to https://github.com/3liz/lizmap-web-client/issues/1635 ...

josemvm:
hi @nboisteault is not possible to fix this issue into the next bug fix release (3.4.5)?
thanks

PR #2502, opened 2021-09-20
Fix: integer key column sorted as text in attribute table

nboisteault:
lizMap.translateWfsFieldValues() change `data` type integer to string
so we return original `data` if its value has not changed or is null
it avoids wrong sorting in attribute table column
* Ticket : #2231
* Funded by 3Liz

### DIFF
--- a/CHANGELOG-3.4.md
+++ b/CHANGELOG-3.4.md
@@ -1,9 +1,10 @@
 # Changelog Lizmap 3.4
 
-## Unreleased
+## 3.4.6 - Unreleased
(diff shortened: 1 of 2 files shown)


rldhont (approved):

mdouchin (approved):
Thanks

rldhont:
@nboisteault keep Unreleased

Gustry:
> @nboisteault keep Unreleased

I have added in the release process the fact to add the date in the changelog.

PR #2348, opened 2021-06-17
[FEATURE] Filter data of layers by polygon depending on the user groups

mdouchin:
This is a new feature, which requires:

* Lizmap plugin >= 3.6.0.
* QGIS Server >= 3.4.0, but is optimized for QGIS 3.16 and above

It allows to filter the layers data spatially by testing the intersection
of the features against the features of the configured polygon layer.

The filtering polygons are selected for the authenticated user based on a chosen field containing a list of user groups.
* At present, if the user is not authenticated, the filtered layers will show no data at all.
* The filter can be deactivated for some groups of users with the already existing right
`lizmap.tools.loginFilteredLayers.override` labeled "See all the data of the filtered layers (attribute or spatial filters)"

* Funded by Valabre, Centre de gravité de la formation des métiers de la Sécurité Civile, de la Recherche, des Nouvelles Technologies et de la Prévention dans le domaine des risques naturels. htps://www.valabre.com

PR #2504, opened 2021-09-20
Remove app context from OGCRequest constructor

laurentj:
Requests objects should rely on the same app context of the project object.
It is not consistent to have different appContext objects (and it could
brake some unit tests).

* Funded by 3Liz

### DIFF
--- a/lizmap/modules/dataviz/classes/datavizPlot.class.php
+++ b/lizmap/modules/dataviz/classes/datavizPlot.class.php
@@ -459,7 +459,7 @@ class datavizPlot
                 $wfsparams['EXP_FILTER'] = $exp_filter;
             }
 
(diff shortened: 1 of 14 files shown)


laurentj:
This is a draft, i will update the patch after the landing of #2348

mdouchin (approved):
Thanks ! @rldhont we need to check the modules for the use of the OGCRequest class

laurentj:
@mdouchin  I already check: they are using deprecated classes of requests so no impact. However, they should use the new classes AND old classes at the same time to be compatible with lizmap 3.4 and lower.

rldhont (approved):

rldhont:
LGTM thanks @laurentj

PR #2517, opened 2021-09-27
Draw: add save button

nboisteault:
Some users complain about draw being always saved in localstorage.
Button is off by default so users can decide if they want draw to be saved

* Funded by paid support


mdouchin (approved):
🦎 💃

CHANGELOG-3.4.md
```diff
@@ -1,5 +1,9 @@
 # Changelog Lizmap 3.4
 
+## 3.4.7 - Not released
```
Gustry, line 3:
```suggestion
## Unreleased

### Added

```

We are following the format : https://keepachangelog.com/en/1.0.0/
So it's need to be `## Unreleased` until it's not released otherwise, the parsing will break.

The parsing will accept subsection such as `Added`, `Changed` etc

Gustry: Sorry, can you rebase because I have updated the release 3.4 branch last week, but I have fixed this one already.
I have updated master just now.

PR #2547, opened 2021-10-20
[BUGFIX] External WMS layers - respect the image format of the source WMS layer

mdouchin:
* If the QGIS WMS layer requests image/jpeg, we must request the same format
* We must not respect the image format configured in the plugin Layers tab
  to avoid requesting a format not available in the WMS server
* This Improves performance of external WMS layers in JPEG since the
  JPEG images are many times smaller than PNG images

* Funded by 3liz


nboisteault:
@mdouchin Please add manual test, I'll make Cypress ones after.

PR #2573, opened 2021-11-02
Use Makefile to run test stack

dmarteau:
Use makefile for running tests stack:

- Better  separation of commands (separate env from docker commands)
- Easier use of direct docker-compose command (use project name as environment variable in env file)
- Better clean up (use `docker-compose down` for real clean up)
- Keep `run-tests` for compatibility

### DIFF
--- a/tests/Makefile
+++ b/tests/Makefile
@@ -1,0 +1,56 @@
+#
+# Makefile for running tests
+#
+
+SHELL:=bash
+.ONESHELL:
+.PHONY: env
+
+export LZMBRANCH:=$(shell git rev-parse --abbrev-ref HEAD)
+
+PREFIX:=lizmap-$(LZMBRANCH)-tests
+
+LIZMAP_USER_ID ?= $(shell id -u)
+LIZMAP_GROUP_ID ?= $(shell id -g)
+LZMPOSTGISVERSION ?= 11-2.5
+LZMQGSRVVERSION ?= 3.16
+LZMPGPORT ?= 8132
+LZMQGSRVPORT ?= 8131
+LZMWEBPORT ?= 8130
+LIZMAP_ADMIN_LOGIN ?= admin
+LIZMAP_ADMIN_EMAIL ?= admin@localhost.local
+LIZMAP_ADMIN_DEFAULT_PASSWORD_SOURCE ?= /srv/etc/admin.conf
+
+env:
+	@echo "Creating environment file for docker-compose"
+	@cat <<-EOF > .env
+	LIZMAP_USER_ID=$(LIZMAP_USER_ID)
+	LIZMAP_GROUP_ID=$(LIZMAP_GROUP_ID)
+	LZMPOSTGISVERSION=$(LZMPOSTGISVERSION)
+	LZMQGSRVVERSION=$(LZMQGSRVVERSION)
+	LZMBRANCH=$(LZMBRANCH)
+	LZMPGPORT=$(LZMPGPORT)
+	LZMQGSRVPORT=$(LZMQGSRVPORT)
+	LZMWEBPORT=$(LZMWEBPORT)
+	LIZMAP_ADMIN_LOGIN=$(LIZMAP_ADMIN_LOGIN)
+	LIZMAP_ADMIN_EMAIL=$(LIZMAP_ADMIN_EMAIL)
+	LIZMAP_ADMIN_DEFAULT_PASSWORD_SOURCE=$(LIZMAP_ADMIN_DEFAULT_PASSWORD_SOURCE)
+	COMPOSE_PROJECT_NAME=$(PREFIX)
+	EOF
+
+
+build: 
+	if [ ! -d qgis-server-plugins/lizmap ]; then
+		./add_server_plugins.sh
+	fi
+
+up: env
+	docker-compose up  -V --force-recreate -d
+
+stop:
+	docker-compose stop 
+
+reset:
+	docker-compose down -v --remove-orphans
+	./lizmap-ctl clean
+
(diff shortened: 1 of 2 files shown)


rldhont:
@dmarteau can you rebase to fix the PHPStan tests

dmarteau:
> can you rebase to fix the PHPStan tests

Done

laurentj (approved):
lgtm

nboisteault (approved):

drillinP (approved):

rldhont:
backports ?

tests/run-docker
```diff
@@ -12,34 +10,14 @@ else
...
-LZMQGSRVVERSION=${LZMQGSRVVERSION:-3.16}
-LZMBRANCH=${LZMBRANCH}
-LZMPGPORT=${LZMPGPORT:-8132}
-LZMQGSRVPORT=${LZMQGSRVPORT:-8131}
-LZMWEBPORT=${LZMWEBPORT:-8130}
-LIZMAP_ADMIN_LOGIN=${LIZMAP_ADMIN_LOGIN:-admin}
-LIZMAP_ADMIN_EMAIL=${LIZMAP_ADMIN_EMAIL:-admin@localhost.local}
-LIZMAP_ADMIN_DEFAULT_PASSWORD_SOURCE=${LIZMAP_ADMIN_DEFAULT_PASSWORD_SOURCE:-/srv/etc/admin.conf}
-End-of-env
+make up "$@"
```
laurentj, line 20: Why "up"?

nboisteault:
We don't backport yet. @laurentj is looking to make a fix on this PR.

tests/Makefile
```diff
@@ -0,0 +1,56 @@
...
+	EOF
+
+
+build: 
+	if [ ! -d qgis-server-plugins/lizmap ]; then
+		./add_server_plugins.sh
+	fi
+
+up: env
+	docker-compose up  -V --force-recreate -d
```
laurentj, line 48: do we really need to use `--force-recreate`? It's annoying, we can loose any changes we made inside the containers (to test a configuration or else).

tests/Makefile
```diff
@@ -0,0 +1,56 @@
...
+	LIZMAP_ADMIN_EMAIL=$(LIZMAP_ADMIN_EMAIL)
+	LIZMAP_ADMIN_DEFAULT_PASSWORD_SOURCE=$(LIZMAP_ADMIN_DEFAULT_PASSWORD_SOURCE)
+	COMPOSE_PROJECT_NAME=$(PREFIX)
+	EOF
+
+
+build: 
+	if [ ! -d qgis-server-plugins/lizmap ]; then
+		./add_server_plugins.sh
+	fi
```
laurentj, line 45: `docker-compose build`, is missing, isn't it?

laurentj:
Sorry, I made a review too quickly yesterday.

tests/Makefile
dmarteau: no, this is aimed to reproduce the  `run-test` command. building the images is done elsewhere and the docker-compose  does not handle build.

tests/run-docker
dmarteau: why not ? this is a make rule that start the stack.

tests/Makefile
dmarteau: > we can loose any changes we made inside the containers

In that case do not use the makefile command but directly the `docker-compose` commands: this is what this PR is about.

tests/Makefile
laurentj: The original script `./run-docker build` did the build by running `docker-compose build` with all environment variables.

tests/Makefile
laurentj: The goal of run-docker was to hide all docker commands with their all parameters. And by forcing to recreate containers at each starts, we loose logs etc

tests/run-docker
laurentj: well, it does not make sens for me to do "up" when we  call "stop" (even for "build")

tests/Makefile
dmarteau: Do as you want.

tests/Makefile
dmarteau: nop, see the diff.

dmarteau: https://github.com/3liz/lizmap-web-client/pull/2573/files#diff-1da304854c4b19a64a89f66263f40f1060ea91eb6f7a0656483b80f192e9bcbbR16

tests/run-docker
dmarteau: > well, it does not make sens for me to do "up" when we call "stop" (even for "build")

Sorry, I do not understand the sentence.

laurentj: when we launch `./run-docker stop` (which called `docker-compose stop` before your change), now it starts containers and then stop containers. It does not make sens for me. Same thing when we do  `./run-docker build` : it starts container whereas it only built containers before the change. 

IMHO it should be only `make "@"`.

dmarteau: > IMHO it should be only make "@".

Do not call make then, no point.  

`docker-run`  has a major problem to recreate env for each passed command, which is a pain when if one wants to customize environnement, and we should be able:
1. recreate the environment only without running any command
2. Use docker-compose commants directly without recreating the environment.

 If you stlill want to pass command directly to docker-compose in `./run-docker then just replace `make up` with `docker-compose $@`

tests/Makefile
nboisteault: @dmarteau when I `./run-docker stop`, containers are recreated and I can see them as started with `docker ps -a`.

dmarteau:
Reedited PR https://github.com/3liz/lizmap-web-client/pull/2604

PR #2591, opened 2021-11-16
Form filter - Prepare the possibility to select more than one combobox item

mdouchin:
If we want to use tools like [SumoSelect](https://hemantnegi.github.io/jquery.sumoselect/) to better display heavily populated combo-boxes, or allow multiple selection, we need to support multiple `<select>` values.

* Funded by 3liz

### DIFF
--- a/lizmap/www/assets/js/filter.js
+++ b/lizmap/www/assets/js/filter.js
@@ -491,7 +491,11 @@ var lizLayerFilterTool = function () {
                     var clist = [];
                     for (var f_val in filterConfig[field_item.order]['items']) {
                         // Get checked status
-                        var achecked = (selectVal == lizMap.cleanName(f_val));
+                        if (Array.isArray(selectVal)) {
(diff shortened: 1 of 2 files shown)


rldhont (approved):

lizmap/www/assets/js/filter.js
```diff
@@ -491,7 +491,11 @@ var lizLayerFilterTool = function () {
                     var clist = [];
                     for (var f_val in filterConfig[field_item.order]['items']) {
                         // Get checked status
-                        var achecked = (selectVal == lizMap.cleanName(f_val));
+                        if (Array.isArray(selectVal)) {
+                            var achecked = ($.inArray(lizMap.cleanName(f_val), selectVal) > -1);
```
nboisteault, line 495: I prefer `var achecked = selectVal.includes(lizMap.cleanName(f_val));`

nboisteault (changes requested):
One small modification otherwise LGTM.

Issue #2582, opened 2021-11-10
[Bug]: Printing twice the same extent does not provide the same


rldhont:
### What is the bug?

When the Print button is clicked twice in a row, the first document get the right extent, the second one get a bad extent.

### Steps to reproduce the issue

1. Go to https://demo.lizmap.com/lizmap_3_5/index.php/view/map/?repository=misc1&project=banyuls-osm
2. Click on Print in the toolbar
3. Click on Print button in the dock
4. Waiting for PDF file generation
5. Check that the PDF get the right extent
6. Click a second time on Print button in the dock
7. Waiting for PDF file generation
8. Check that the PDF get the bad extent

### Lizmap version

3.5.0-RC.3

### QGIS desktop version

3.16.13

### QGIS server version

3.16.13

### Operating system

Debian

### Browsers

Firefox

### Browsers version

Firefox 93

PR #2592, opened 2021-11-16
[Bugfix] Printing twice the same extent does not provide the same

rldhont:
The extent for printing has to be cloned before transform to do not change the printing geometry.

* Ticket : #2582
* Funded by 3liz


nboisteault (approved):
I have tested and it is ok.👍

PR #2604, opened 2021-11-23
Use Makefile to run test stack

dmarteau:
Use makefile for running tests stack:

- Better  separation of commands (separate env from docker commands)
- Easier use of direct docker-compose command (use project name as environment variable in env file)
- Better clean up (use `docker-compose down` for real clean up)
- Keep `run-tests` for compatibility

### DIFF
--- a/tests/run-docker
+++ b/tests/run-docker
@@ -4,20 +4,21 @@ set -e
 
 CMD=$1;
 if [ -z "$CMD" ]; then
-    CMD="up"
+    CMD="up -d"
 else
     shift
 fi
+
+# Re-create environment
+make env
 
 if [ "$CMD" == "reset" ]; then
     # Purge docker conainers/volumes
     make reset
     exit 0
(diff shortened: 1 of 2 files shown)


nboisteault (approved):

nboisteault:
I've tested => ok

tests/run-docker
```diff
@@ -14,10 +14,9 @@ if [ "$CMD" == "reset" ]; then
...
     exit 0
 elif [ "$CMD" == "build" ]; then
-    make build
+    make build-plugins
 fi
 
-make up "$@"
-
-echo "Use 'docker-compose logs <service>' to display logs"
+make env
```
laurentj, line 20: it should be done line 11, else `make reset` will fail if the .env does not exists

dmarteau: Fixed

laurentj (approved):
Ok for me. thank you

Issue #1304, opened 2019-07-26
Improve the upload in feature form


laurentj:
We could replace the code to support the upload field, by the recent upload2 widget of Jelix. It will remove some duplicated code, and will avoid some errors like #572 because the actual code is fragile.

Issue #2427, opened 2021-08-11
upload photos from webcam using a mobile device - stop working


josemvm:
### What is the bug?

lizmap lost the ability to upload images via camera, the user can only choose browser files

### Steps to reproduce the issue

...

### Lizmap version

3.4.4

### QGIS desktop version

3.16.9

### QGIS server version

3.16.9

### Operating system

Android 10

### Browsers

Firefox

### Browsers version

90.1.3

Gustry:
It works perfectly on https://demo.lizmap.com/lizmap/index.php/view/map/?repository=feat1&project=observatoire with my phone and Firefox 90.1.3.

josemvm:
your lwc version?

Gustry:
You can try all of them, it all works for the camera.

https://demo.lizmap.com/lizmap/index.php/view/map/?repository=feat1&project=observatoire => 3.5
https://demo.lizmap.com/lizmap_3_5/index.php/view/map/?repository=feat1&project=observatoire
https://demo.lizmap.com/lizmap_3_4/index.php/view/map/?repository=feat1&project=observatoire
https://demo.lizmap.com/lizmap_3_3/index.php/view/map/?repository=demo&project=observatoire

But actually, I have noticed, I can't choose any file on my phone. So if I understand, it's the reverse from you ?

josemvm:
hi @Gustry thanks for your feedback

i can confirm with your observatoire projects on my phone with firefox.
i have also my own projects that ask for webcam and i have others that only ask for files and i don´t know why...

i' am trying with other browsers (opera, chrome, edge) and all of them ask for webcam, audio or files but not with firefox

laurentj:
The enabling of the webcam capture depends on many parameters.
The first one is [the `capture` HTML attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/capture). In lizmap, we set it to `camera` but it seems this is not any more a good value. HTML specifications have changed many times since few years. In my patch #2617 , I set it to `environment`.

However, the `capture` attribute is not supported by all browsers. It seems Firefox does not support it but perhaps [this compatibility matrix](https://caniuse.com/html-media-capture) is not updated.

The second parameter seems to be the type of file we indicate into the `accept` attribute, so into QGIS. It should be only an image type. Perhaps the type is not set correctly into your projects.

laurentj:
It should be fixed in the incoming Lizmap 3.5. Reopen the ticket if there is still an issue.

PR #2617, opened 2021-11-30
Improvement of upload content into feature forms and image editor for uploaded images

laurentj:
The support of uploaded file has been improved into feature forms. Better mechanism to check the uploaded files.
There is also now an image editor when the user wants to upload an image. The image can be cropped, rotated, and is resized automatically if it is higher than a size we can set into the lizmap configuration.

* Ticket: #1304
* Ticket: #572
* Funded by Mairie d'Avignon


laurentj:
It probably fixes also #2427

nboisteault:
Great work @laurentj ! I have tested and i just have an error message in console when canceling image modification, not sure it is important. Also I noticed that there is no cross at top right to close the image modification window.
Peek 30-11-2021 16-59

There is some work needed for small screens. Do you want me to work on this?

image

laurentj:
@nboisteault  I fixed the JS error, and hide the button.

It would be fine if you could work on the mobile version. But it will not easy, as the dialog has fixed size, and the canvas should have specific size, for rotation etc. (see the code into imageSelector.js)

nboisteault:
@laurentj good! How it possible to translate the `Cancel` button?

image

nboisteault:
Concerning dialog size, it is ok when putting smartphone in landscape position, so we can consider it is not an issue.

nboisteault (approved):
Just the Cancel button to translate if possible.

laurentj:
cancel button translated :-)

nboisteault:
@laurentj to run Cypress in CI you need to allow this action in your settings then:
cypress_manual

PR #2618, opened 2021-11-30
Editing - Fix the value relation widget with multiple text values on feature modification

mdouchin:
In the form widget, when:

* there is a `text` field configured with a "Value relation" widget, 
* with the option `allow multiple` activated
* using a `text` field in the referenced layer for the code checkbox

QGIS adds double quotes around the text values, such as `{"one_value", "second_value"}`.

We need to remove these double-quotes for setting correctly the corresponding checkboxes.

Cypress test added.

Funded by #3liz


nboisteault (approved):
I have not tested but it looks good!

mdouchin:
@rldhont ok for merge ?

rldhont:
@mdouchin after 3.4.8

mdouchin:
up

rldhont (approved):

PR #2620, opened 2021-12-01
Don't show search result if search query is empty

ygorigor:
Backport to 3.5 and 3.4

Without this commit
without_commit

With this commit
with_commit


rldhont (approved):

nboisteault:
@ygorigor We need a message telling there is no results. Could you add a message with `lizMap.addMessage()`. Maybe there are the right locales yet.

ygorigor:
> @ygorigor We need a message telling there is no results. Could you add a message with `lizMap.addMessage()`. Maybe there are the right locales yet.

@nboisteault I added a message info.

rldhont (approved):

PR #2629, opened 2021-12-02
Better header override and CSS clean

nboisteault:


lizmap/www/assets/css/main.css
```diff
@@ -35,7 +20,8 @@ a:hover {
   z-index: 1030;
   margin-bottom: 0px;
   overflow: visible;
-  background : url('img/header-background-medium.jpg') repeat fixed #2B2B2B;
+  background-color: #2B2B2B;
+  background-image : url(img/header-background-medium.jpg);
```
mdouchin, line 24: I would add a background-repeat and the other config to set the "fixed" property, to be sure on big screens, there is no blank (the image must be repeated)

lizmap/www/themes/default/css/main.css
```diff
@@ -2,21 +2,9 @@ body {
...
-a:hover {
-  color:white\9;
-  text-decoration:none\9;
-}
-/* END for ie 8*/
-
 #header {
-  background : url(img/header-background-medium.jpg) repeat fixed #2B2B2B;
+  background-color: #2B2B2B;
+  background-image : url(img/header-background-medium.jpg);
```
mdouchin, line 7: same here ?

mdouchin (changes requested):

lizmap/www/assets/css/main.css
nboisteault: If I set `background-attachment : fixed`, the black color goes above image. Default is `background-attachment : scroll`. As there is no scroll in the element I think we should let default.

nboisteault: I've added `background-repeat: repeat;`.

mdouchin (approved):

PR #2635, opened 2021-12-06
Fix and test layer without geom for form filter

nboisteault:
* Funded by 3Liz


nboisteault:
@mdouchin Form filter as no incidence on attribute table so how can I visually test that a layer without geometry is filtered?

PR #2637, opened 2021-12-07
[edition] rework the form management by using jForms features

laurentj:
The form does not have to be initialize each time we need to display it,
for example to show errors in the form. The form data can be initialize
during its creation. It simplify the code, save space in the session, and
avoid potential issues.

* Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/lib/Form/QgisForm.php
+++ b/lizmap/modules/lizmap/lib/Form/QgisForm.php
@@ -184,27 +184,12 @@ class QgisForm implements QgisFormControlsInterface
         // Set form's private data
         $privateData = array();
 
-        // FIXME why ? there is already these fields into the form. We duplicate data here
-        $privateData['liz_repository'] = $layer->getProject()->getRepository()->getKey();
-        $privateData['liz_project'] = $layer->getProject()->getKey();
-        $privateData['liz_layerId'] = $layer->getId();
-        $privateData['liz_featureId'] = $featureId;
-        $privateData['liz_geometryColumn'] = $this->dbFieldsInfo->geometryColumn;
-
+        // we need some QgisFormControl data in some case where we have only the jForms object,
+        // not the QgisForm object (like in jForms datasource objects etc.)
         $privateData['qgis_controls'] = array();
         foreach ($this->formControls as $fieldName => $formControl) {
             $privateData['qgis_controls'][$formControl->ref] = array(
-                'fieldName' => $formControl->fieldName,
-                'defaultValue' => $formControl->defaultValue,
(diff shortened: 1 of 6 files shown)


mdouchin (commented):
Looks good to me. Do you need me to test manually ?

rldhont:
I'd like to check the why questions added by @laurentj in the code before approving.

nboisteault:
3.5.1 please :)

lizmap/modules/lizmap/lib/Form/QgisForm.php
```diff
@@ -183,6 +183,8 @@ public function __construct($layer, $form, $featureId, $loginFilteredOverride, A
 
         // Set form's private data
         $privateData = array();
+
+        // FIXME why ? there is already these fields into the form. We duplicate data here
```
rldhont, line 187: If data are duplicated here, we have to update:
* `qgisForm.listener.php` that uses them for checking if the jForm is a qgisForm https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/classes/qgisForm.listener.php#L21
* `qgisExpressionUtils.class.php` that uses them for getting lizmap project from form https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/classes/qgisExpressionUtils.class.php#L285
* `controllers/edition.classic.php` that uses them for building a qgisForm to evaluate input fields group visibilities https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/controllers/edition.classic.php#L1456

lizmap/modules/view/forms/edition.form.xml
```diff
@@ -9,7 +9,7 @@
 <hidden ref="liz_proj4"/>
 <hidden ref="liz_geometryColumn"/>
 <hidden ref="liz_status"/>
-<hidden ref="_submit"/>
+<hidden ref="_submit"/> <!-- why ? duplicated field -->
```
rldhont, line 12: To store the submit value in the form: https://github.com/3liz/lizmap-web-client/commit/4ca94407a71328cf0780284c601fee7e4e339aed#diff-4c07e19f3b7ab2f105b286b0564be51333f7d8ae4f98602921f8f1d49aa3a497R830

lizmap/modules/lizmap/lib/Form/QgisForm.php
laurentj: I removed all this private data excepts `['qgis_control'][X]['valueRelationData']` and `qgis_groupDependencies`

nboisteault (approved):
Cypress happy!

PR #2645, opened 2021-12-13
QGIS Vector Layer - New options timeout & setSearchPathFromLayer to set the PostgreSQL connection profile

mdouchin:
This allows Lizmap Web Client modules to set their connection timeout
and avoid them to share the same PHP PostgreSQL connection.
At present, the connection sharing system is only determined
by the `host, user, service, port, dbname`
not by the `search_path`.

See: https://www.php.net/manual/fr/function.pg-connect.php

In some cases, when a `search_path` is needed, we need to avoid
the modules to reuse the same connection to the database.

* Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisVectorLayer.class.php
+++ b/lizmap/modules/lizmap/classes/qgisVectorLayer.class.php
@@ -190,11 +190,15 @@ class qgisVectorLayer extends qgisMapLayer
      * getDatasourceConnection() is not useful, as we could need the profile
      * to give to jDao or other components that need a profile, not a connection
      *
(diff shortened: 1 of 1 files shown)


mdouchin:
In LWC modules, we will need to determine if the function needs the new parameters or not to adapt the code to different LWC versions.

* By checking LWC version or
* Or by checking if the function needs parameters such as 

```php
function need_params($func_name)
{
    $reflect = new ReflectionFunction($func_name);

    return !empty($reflect->getParameters());
}
```

@laurentj your opinion ?

rldhont (approved):

nboisteault (approved):

PR #2651, opened 2021-12-14
Fix postgresql connection share between modules

laurentj:
The search_path is not used for the connection, so layers
and modules can share the same connection although they want to use
a different search_path. A solution is to set a different timeout for
each different search_path.

* Funded by 3liz


mdouchin (approved):
Merci !

nboisteault (approved):
Cypress is happy!

Issue #2611, opened 2021-11-26
Snapping missing when editing existing feature(?)


gioman:
### What is the bug?

I have not followed closely the developments when snapping was added, so forgive me if the following is expected or I'm missing something obvious, so is not a bug.

I have noticed that is possible to activate snapping when **adding** a new feature

![image](https://user-images.githubusercontent.com/1951107/143553678-4481a0f6-5ef8-44b5-82d5-4bc6a9184f2e.png)

but when **editing** an existing feature:

![image](https://user-images.githubusercontent.com/1951107/143553577-be1b6db6-5335-4f1d-9e49-50e1d58a4b95.png)

### Steps to reproduce the issue

1) Set up a project with a layer with editing and snapping.

2) add a new feature, in "digitazion" the option to activate snapping is there

3) save the feature

4) identify a feature

5) choose to edit it > in "digitazion" there is no option to enable snapping

### Lizmap version

3.4.7

### QGIS desktop version

3.16.14

### QGIS server version

3.16.14

### Operating system

Ubuntu 20.04, Windows 10

### Browsers

Firefox, Chrome, Microsoft Edge

### Browsers version

Latest

gioman:
@nboisteault thanks!

PR #2657, opened 2021-12-16
Fix: Snapping missing when editing existing feature

nboisteault:
This part of the code is hard to change and there are regression risks. I'll try to add some more tests later.

* Ticket : #2611
* Funded by 3Liz


mdouchin:
Thanks for looking into this. Could you write a manual test if an automatic "Cypress" one is not easy to add ?

gioman:
@nboisteault many thanks!

nboisteault:
> Thanks for looking into this. Could you write a manual test if an automatic "Cypress" one is not easy to add ?

Done.

mdouchin (approved):

Issue #2661, opened 2021-12-17
Error undefined method getXmlRelation() when using child in form (version 3.5.1-pre)


arno974:
### What is the bug?

When trying to add a child (1-N relation) to an existing feature, the child form keeps loading. The log file raise the error below :

`Call to undefined method Lizmap\Project\Project::getXmlRelation()	/www/lizmap-3-5-1/lizmap/modules/lizmap/lib/Form/QgisForm.php	1399`

The same form (same QGIS project) works fine when using the 3.4.7 lizmap version.

### Steps to reproduce the issue

1- Create a 1-N relation in QGIS
2- Activate edition for both layers in Lizmap Extension
3- Edit a feature and add a child entity (only attributes to modify in my case)
4- The form keeps loading

### Lizmap version

3.5.1-pre

### QGIS desktop version

3.16.8

### QGIS server version

3.16.8

### Operating system

Ubuntu

### Browsers

Firefox

### Browsers version

Firefox 93.0

### Relevant log output

```shell
Call to undefined method Lizmap\Project\Project::getXmlRelation()	/www/lizmap-3-5-1/lizmap/modules/lizmap/lib/Form/QgisForm.php	1399
```

laurentj:
Thanks for your report. The bug is fixed. 
You can try with [the package for the future 3.5.1](https://packages.3liz.org/pub/lizmap/unstable/3.5/lizmap-web-client-3.5.1-pre.zip).

PR #2664, opened 2021-12-20
Fix a regression during the init of relation references into forms

laurentj:
Relation data should be read during the xml parsing, not at runtime.

* Ticket : #2661
* Funded by 3liz


mdouchin (approved):

nboisteault (commented):
I have tested with a local project and it fixed the issue.
Please run with Cypress in CI and give me the link please.

PR #2671, opened 2021-12-30
Create lizmap-feature-toolbar element

nboisteault:
This element replace toolbar in popup and attribute table. It also add the toolbar to children popup.

* Ticket : paid support 1794
* Funded by [Valabre](https://www.valabre.com/)


nboisteault:
@mdouchin do we keep `btn-warning` on activated button? What do you think about `btn-success` instead?

https://getbootstrap.com/2.3.2/base-css.html#buttons

nboisteault:
@mdouchin Should we really not activate feature `filter` when there is a `startupFilter`? Why this limitation?

https://github.com/3liz/lizmap-web-client/blob/master/lizmap/www/assets/js/attributeTable.js#L2974

tests/qgis-projects/tests/feature_toolbar.qgs.cfg
```diff
@@ -128,6 +128,15 @@
             "hideLayer": "False",
             "custom_config": "False",
             "order": 0
+        },
+        "children_layer": {
```
Gustry, line 132: Is-it a known key `children_layer` ?
I see nothing in the source code of the plugin https://github.com/3liz/lizmap-plugin/search?q=children_layer

nboisteault: It is the layer name :)

PR #2691, opened 2022-01-13
Fix: geom can be created on existing feature w/o geom

nboisteault:
* Funded by 3Liz


nboisteault:
@rldhont @mdouchin It works but the logic might be improved. We should not spend many more time on this as I'll remake edition for Lizmap 4.

nboisteault:
Note to myself : I'll have to make some adjustment in test after this PR is merged https://github.com/3liz/lizmap-web-client/pull/2671

mdouchin (approved):

Issue #2680, opened 2022-01-04
[Bug]: Bad relationship value when creating a child from a parent (works well in QGIS)


sigsdis30:
### What is the bug?

When creating a child from a parent, the value entered from relationship reference or relationship value is not the parent value.

Parent table: "pistes" with id serial and code_de_la_piste (unique in the table. Examples: U18_1CG, U18_2CG)
Child table: "controles" with id serial and code_de_la_piste
Relation 1:M between these tables on code_de_la_piste column
[relationship_value.pdf](https://github.com/3liz/lizmap-web-client/files/7807593/relationship_value.pdf)

### Steps to reproduce the issue

In lizmap:
1. Click on "piste" U18_1CG and edit
2. Add an child: the "code_de_la_piste" value isn't U18_1CG but U18_2CG that is the code of another piste (with the same code beginning)

In QGIS project, it works well

### Lizmap version

3.4.8

### QGIS desktop version

3.16.4

### QGIS server version

3.16.8

### Operating system

Windows 10

### Browsers

Chrome

### Browsers version

96.0.4664.110

mdouchin:
We have reproduced this issue too. I will propose a fix ASAP

sigsdis30:
Thank you!

sigsdis30:
Have you made any progress on fixing this bug?

After our first post, we filtered the problematic rows in the table but another problem occurs with WFS: "Lizmap ne peut pas récupérer cet objet via le WFS". We probably made a mistake leaving two layers with the same name (?). We have therefore completely recreated our data structure (schema, tables, views...) as well as the lizmap map. In the new one, it's even worse: almost all objects get wrong parent values !

We plan to hide the map so that it is no longer accessible to users. Without it, they will insert wrong data into the database.

mdouchin:
Sorry for the delay, I am working on it, and plan to release a fix before this weekend.

sigsdis30:
If it can be useful, we make a few observations. We mainly work on two layers: water points (points) and tracks (lines). In both cases, these are views whose identifier is calculated with row_number().
In our production map, the problem only occurs on the tracks. In the new map, it only occurs on water points. In both cases, the problem is related to a bad identifier referenced in the code. The issue [#235](https://github.com/3liz/lizmap-plugin/issues/235) helped us understand.
For example:
[...]

PR #2695, opened 2022-01-17
Attribute table - Fix the link of selected features between a parent and a child layer

mdouchin:
In the attribute table tool, when you select a feature in the parent layer, and one or several features in the child layers, Lizmap should update the child table the right way. This PR fixes a bug.

* Ticket : #2680
* Funded by Valabre


nboisteault (approved):
I have modified the Cypress test to handle parent and child feature having different ids.
I'll backport your fix and not my commit related to Cypress as it is not backportable.

PR #2708, opened 2022-01-20
Use values instead of keys in attribute table

### LINKED ISSUE #747: Value map not used in Table view
In a qgs I have several fields factorized (0,1,2, etc.) with labels configured in Layer properties>Fields>Value map. The labels are shown correctly in the LM popups, but in table view the original numeric values are displayed instead.

nboisteault:
* Fix #747
* Funded by Grand Narbonne


nboisteault:
Supersed by https://github.com/3liz/lizmap-web-client/pull/2722

PR #2713, opened 2022-01-23
Graph view in Lizmap Records - Record Count and Record Detail

sergiuspinheiro:
Visualization of Lizmap Records data in graphic. https://github.com/3liz/lizmap-web-client/issues/2712

I know it's a non-essential feature but it makes it easier to understand the given use of lizmap.

I made some changes to the lizmap code for testing, as well as views in postgres to provide data to the graphs

Thanks


laurentj:
Hi,

Thanks for your patch. We should discuss at 3Liz if we want a such feature.

Anyway, for the moment, we cannot accept your patch as is, because your code is not integrated into Lizmap, and is not using the framework. Few of many consequences of this missing integration are:

* the postgresql connection does not use credentials and configuration from profiles.ini.php
* It seems your sql script is not compatible with Sqlite (even if we prefer Postgresql, we still have some users who are using Sqlite).
* there is no check of user rights. So all scripts inside  `lizmap/www/graph/` are accessible to everybody, and this is not what we want ;-)
* The script lizlog_graph_pgsql_view.sql is not called anywhere during the installation, so views will not be available automatically
* Url to Lizmap assets or URLs to scripts inside `www/graph/`, may be wrong if lizmap is installed in a subdirectory of a document root of a web server (you don't use the ["base path"](https://docs.jelix.org/en/manual-1.6/urls#configuration) provided by Lizmap/Jelix).

etc.

Locales string should be created only into en_US files (and in english). Translations should be written into [Transifex](https://www.transifex.com/3liz-1/lizmap-locales/dashboard/). All locales files into app/locales/ are overwritten by strings coming from Transifex, so your translations may disappear.

To have a  patch with a good code integrated into Lizmap, you should use a controller, some templates, jDb (the API to access to the database), jUrl to forge urls etc. To learn how to use the framework, you can read the [manual of Jelix 1.6](https://docs.jelix.org/en/manual-1.6).

PR #2720, opened 2022-01-24
Convert QGis XML Option value based on type attribute

rldhont:
The QGis XML Option contains a type attribute that contains the QVariant type name.

* QgsXmlUtils::writeVariant https://github.com/qgis/QGIS/blob/master/src/core/qgsxmlutils.cpp
* QMetaType::Type https://github.com/qt/qtbase/blob/5.6/src/corelib/kernel/qmetatype.cpp

Funded by 3liz


laurentj (approved):
lgtm

PR #2722, opened 2022-01-24
Use values instead of keys in attribute table, form filters and dataviz

### LINKED ISSUE #747: Value map not used in Table view

nboisteault:
* Funded by Grand Narbonne
Fix #747


nboisteault:
Here is a video illustrating the new behavior when columns have keys/values defined with `value map`, `value relation` or `relation reference`. (Columns are not in the same order than in QGIS due to a known bug)

https://user-images.githubusercontent.com/2145040/155533448-14a22f01-9cd1-485b-943a-eb45b0f6016e.mp4

mdouchin:
@nboisteault I just pushed on this branch. I have not run the Cypress tests yets, but will add the corresponding tag.

You may go on and apply the "translation" for the Form filter & the Dataviz tools (I would do it only for X values for the dataviz plots at first) cc @Gustry

mdouchin:
I think we should add an option in the Lizmap plugin, on by default, to let the editor decide wether or not to do the translation ? 
What is your opinion on this @Gustry @rldhont ?
End-to-end tests concerning the attribute table tool must probably by be reviewed too

nboisteault:
@mdouchin We also talked about a checkbox/button in attribute table to toggle keys <=> values view. You'd like both?

mdouchin:
> @mdouchin We also talked about a checkbox/button in attribute table to toggle keys <=> values view. You'd like both?

No, I think a checkbox/button in the web UI would be enough

mdouchin:
@nboisteault I added the support for multiple values (when checking the checkbox "Allow multiple" aka "Permettre les sélections multiples) in the QGIS RelationReference form widget)

Not tested with multiple integer codes (tested only with multiple string codes like uids)

nboisteault:
@mdouchin I have this test failing https://github.com/3liz/lizmap-web-client/runs/5509814485?check_suite_focus=true#step:9:492
But it is said in test file `// REMOVE WHEN THE QGIS SERVER BUG HAS BEEN FIXED`
Should I fix this issue or it is not necessary?

nboisteault:
@mdouchin Cypress fails in CI because of notices in PHP : https://github.com/3liz/lizmap-web-client/runs/5540972697?check_suite_focus=true

Could you take a look?

nboisteault:
@mdouchin Please fix PHP notice and we will merge then.

Gustry:
Well done, all green ! 😎

lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -222,6 +228,55 @@ protected function readProject()
...
+     *
+     * @return array Array of layer ids
+     */
+    protected function getLayersWithLabels()
+    {
+        // Keep a list of layer ids for which to replace the code by labels
+        $layersWithLabeledFields = array();
+
+        // Attribute layers
+        foreach ($this->cfg->getAttributeLayers() as $key => $config) {
```
nboisteault, line 254:
```suggestion
        foreach ($this->cfg->getAttributeLayers() as $config) {
```

lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -222,6 +228,55 @@ protected function readProject()
...
+            if ($config->hideLayer == 'True') {
+                continue;
+            }
+            $layersWithLabeledFields[] = $config->layerId;
+        }
+
+        // Dataviz layers
+        foreach ($this->cfg->getDatavizLayers() as $o => $config) {
```
nboisteault, line 262:
```suggestion
        foreach ($this->cfg->getDatavizLayers() as $config) {
```

lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -222,6 +228,55 @@ protected function readProject()
...
+            $layerId = $config->layerId;
+            if (array_key_exists($layerId, $layersWithLabeledFields)) {
+                continue;
+            }
+            $layersWithLabeledFields[] = $config->layerId;
+        }
+
+        // Form filter layers
+        foreach ($this->cfg->getFormFilterLayers() as $o => $config) {
```
nboisteault, line 271:
```suggestion
        foreach ($this->cfg->getFormFilterLayers() as $config) {
```

PR #2723, opened 2022-01-25
Fix Remove user private group from request to QGIS Server

rldhont:
Lizmap provides user login and groups to QGIS Server through headers or parameters. The user private group was in the user group list.

### DIFF
--- a/lizmap/modules/lizmap/lib/App/AppContextInterface.php
+++ b/lizmap/modules/lizmap/lib/App/AppContextInterface.php
@@ -65,11 +65,12 @@ interface AppContextInterface
     public function aclUserGroupsId();
 
     /**
-     * Returns the groups ids of an user.
+     * Retrieve the list of group the given user is member of
+     * in the acl system.
      *
-     * @param string $login The login of the user
+     * @param string $login The user's login
      *
-     * @return array an array containing the user's groups ids
+     * @return array list of group id
      */
     public function aclGroupsIdByUser($login);
 
--- a/lizmap/modules/lizmap/lib/App/AppContextInterface.php
+++ b/lizmap/modules/lizmap/lib/App/AppContextInterface.php
@@ -75,6 +75,26 @@ interface AppContextInterface
     public function aclGroupsIdByUser($login);
 
     /**
+     * Retrieve the list of public group the given user is member of
+     * in the acl system.
+     *
+     * @param string $login The user's login
+     *
+     * @return array list of public group id
+     */
+    public function aclUserPublicGroupsId($login = null);
+
+    /**
+     * Get the private group for the current user or for the given login
+     * in the acl system.
+     *
+     * @param string $login The user's login
+     *
+     * @return string the id of the private group
+     */
+    public function aclUserPrivateGroup($login = null);
+
+    /**
      * Retrieve the list of groups properties, the current user is member of,
      * in the acl system.
      *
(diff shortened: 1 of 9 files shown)


lizmap/modules/lizmap/lib/App/AppContextInterface.php
```diff
@@ -65,14 +65,25 @@ public function aclCheck($role, $resource = null);
...
 
+    /**
+     * Get the private group for the current user or for the given login
+     * in the acl system.
+     *
+     * @param string $login The user's login
+     *
+     * @return string the id of the private group
+     */
+    public function aclUserPrivateGroup($login = null);
```
laurentj, line 85: It would be better to have a new method `aclPublicGroupsIdByUser($login)`. It would avoid to repeat the code that remove the private group from the list of groups given by `aclGroupsIdByUser`.

And this new method could call `aclGroupsIdByUser()`, and just remove id which begins with `"__priv__"`, like [we do in QgisForm:: updateFormByLogin()](https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/lib/Form/QgisForm.php#L1108) .

lizmap/modules/lizmap/lib/Request/OGCRequest.php
```diff
@@ -121,9 +121,16 @@ public function parameters()
             ));
         }
 
-        // Provide user and groups to lizmap plugin access control
+        // Provide user and groups to lizmap plugin access control without private group
         $user = $appContext->getUserSession();
         $userGroups = $appContext->aclUserGroupsId();
+        $userPrivateGroup = $appContext->aclUserPrivateGroup();
+        if ($userPrivateGroup) {
```
laurentj, line 128: there is no need to do this if statement, if we call the new proposed method `aclPublicGroupsIdByUser()` instead of `aclUserGroupsId()`.

lizmap/modules/lizmap/lib/Request/Proxy.php
```diff
@@ -459,6 +459,13 @@ protected static function userHttpHeader()
         // Provide user and groups to lizmap plugin access control
         $user = $appContext->getUserSession();
         $userGroups = $appContext->aclUserGroupsId();
+        $userPrivateGroup = $appContext->aclUserPrivateGroup();
+        if ($userPrivateGroup) {
```
laurentj, line 463: there is no need to do this if statement, if we call the new proposed method `aclPublicGroupsIdByUser()` instead of `aclUserGroupsId()`.

laurentj (changes requested):

laurentj:
Probably we would need to replace other calls of  `aclUserGroupsId()` by the new proposed method `aclPublicGroupsIdByUser()`, like in `qgisExpressionUtils`, `Project`, `QgisForm`, but i'm not confident enough about the business rules that these piece of code implement.

rldhont:
Thanks a lot @laurentj I have no idea how to name the method `aclPublicGroupsIdByUser` to avoid duplicating code. I will do the changes.

laurentj (approved):
lgtm

nboisteault:
@rldhont @mdouchin 3.5.1 or 3.5.2 ?

rldhont:
Cypress test are now green !

laurentj (approved):
lgtm

PR #2726, opened 2022-01-25
Use Lizmap version to refresh some assets cache

nboisteault:
Sometimes users have to force refresh their cache to have up to date assets. This can be avoided by adding the Lizmap version as an URL parameter.

### DIFF
--- a/lizmap/responses/myHtmlMapResponse.class.php
+++ b/lizmap/responses/myHtmlMapResponse.class.php
@@ -23,6 +23,10 @@ class myHtmlMapResponse extends jResponseHtml
 
         $this->title = '';
 
+        // Use Lizmap version to refresh some assets cache
+        $xmlLoad = simplexml_load_file(jApp::appPath('project.xml'));
+        $version = (string) $xmlLoad->info->version;
+
         // CSS
         $css = jApp::config()->jquery['jqueryui.css'];
         foreach ($css as $file) {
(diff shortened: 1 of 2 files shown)


rldhont:
Thanks @nboisteault you probably need to update other files:

* https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/view/controllers/lizMap.classic.php
* https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/lib/Project/Project.php
* others

laurentj (changes requested):
We should have another way to retrieve the version or we should take another value. Loading XML takes too many resources.

nboisteault:
> We should have another way to retrieve the version or we should take another value. Loading XML takes too many resources.

I think this optimization should be handled for 3.5.2 as we need to release 3.5.1 quickly.

nboisteault:
@laurentj should `addCSSLink` and `addJSLink` methods be modified to automatically add the Lizmap version as URL parameter ? How to have the same behavior for assets added this way : https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/dataviz/classes/dataviz.listener.php#L22 ?

@laurentj @mdouchin @rldhont 
There are only one CSS and JS to force refresh for https://github.com/3liz/lizmap-web-client/pull/2492/files to work correctly so I can add `?v=3.5.1` for those for 3.5.1 release. This way we'll have time to think about a good refactoring later.

laurentj:
I propose an alternative to this PR, which takes less IO resources : #2729

PR #2728, opened 2022-01-25
Tests - Remove user and password from the docker stack in QGS files

Gustry:
We should use exclusively the service file when connecting to the PG database in tests.

@nboisteault I suspect you have "login" and "password" and "service" registered in your QGIS Desktop. Please keep only the service file. Thanks ;-)

I can't test the docker stack for now.

I will forward port to 3.5 and master.


nboisteault (approved):
Indeed I had user and password at some point but it is no more the case. Thanks for cleaning.

Gustry:
@nboisteault I added some changes, for instance, to remove SQL owner in the SQL file.

This PR allows us to publish easily on https://demo.snap.lizmap.com/lizmap_3_4/index.php/view/?repository=testsrepository

PR #2729, opened 2022-01-26
Add a revision parameter on assets url for cache

laurentj:
It allows to invalidate the cache in browser of CSS and JS,
each time the temp directory is clean, so each time Lizmap
is upgraded.

* Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmap.class.php
+++ b/lizmap/modules/lizmap/classes/lizmap.class.php
@@ -38,17 +38,17 @@ class lizmap
     /**
      * @var lizmapServices The lizmapServices instance for the singleton
      */
-    protected static $lizmapServicesInstance = null;
+    protected static $lizmapServicesInstance;
 
     /**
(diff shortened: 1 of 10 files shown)


mdouchin (commented):
Ok for me, but we should check if the documentation "how to upgrade" has a strong warning to let the admin understand he should use the `lizmap/install/clean_vartmp.sh` after the upgrade. Il semble que oui: https://docs.lizmap.com/current/fr/install/upgrade.html#launch-the-installer

nboisteault (commented):
I'd really like to have this PR in master and release_3_5 for a while to test it exhaustively.
Conclusion: LGTM but I'm making a small patch for 3.5.1 and this PR should be merged for 3.5.2.

nboisteault:
@laurentj I have made the small patch ebe154a9ef78b024bfdecc1415c04d72dbb471cd
Could you run Cypress on your PR please?

PR #2740, opened 2022-01-31
Attribute table - Fix the export when a selection or a filter is active

mdouchin:
* when the layer is not spatial
* when the layer has parenthesis inside its name

Backported from master to release_3_4 without the Cypress tests.

* Ticket : #
* Funded by


mdouchin:
@nboisteault @rldhont Have not tested yet, just manually backported

mdouchin:
up

rldhont (approved):

PR #2743, opened 2022-02-01
[Bugfix] Edition: geom can be created on existing feature w/o geom on modify only

rldhont:
If the layer has only modify geometry capabilities, geomatry cannot be created on existing feature.

* Funded by 3Liz


nboisteault (approved):
If you have manually tested, it is ok for me.

PR #2754, opened 2022-02-03
Fix value map in 3.5

nboisteault:
* Ticket : #2700
* Funded by 3Liz


nboisteault:
@laurentj I have added unit test failing currently.

PR #2773, opened 2022-02-11
GetprojectConfig as application/json instead of plain/text

rldhont:
The GetProjectConfig was send as plain/text instead of application/json.

This commit fixes it.

* Funded by 3Liz


rldhont:
@nboisteault Cypress has been run and it is green

PR #2796, opened 2022-02-24
Search without private group

rldhont:
Search controller and FTS does not use appContext

* Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmapFts.class.php
+++ b/lizmap/modules/lizmap/classes/lizmapFts.class.php
@@ -11,9 +11,7 @@ class lizmapFts
 {
     protected $sql;
 
-    protected $project;
-
-    protected function setSql()
+    protected function generateSql($project)
     {
 
         // Build search query.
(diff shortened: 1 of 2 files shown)


Gustry:
FTS ? Is-it not linked to only QuickFinder ? Or is-it something else ?

rldhont:
Cypress green

mdouchin (approved):
LGTM

rldhont:
Rebase to get cypress updated to 9.5.0 and Cypress tests are still green.

lizmap/modules/lizmap/classes/lizmapFts.class.php
```diff
@@ -11,9 +11,7 @@ class lizmapFts
 {
     protected $sql;
 
-    protected $project;
-
-    protected function setSql()
+    protected function setSql($project)
```
laurentj, line 14: I recommend to rename this method to something like `generateSql`, as it is not really a setter method (i.e, the parameter is not a value to set).

laurentj (approved):
Ok for me, if the method setSql is renamed.

lizmap/modules/lizmap/classes/lizmapFts.class.php
rldhont: Method renamed

rldhont:
Thanks @laurentj

laurentj (approved):

PR #2769, opened 2022-02-11
OGC Requests: Check VERSION parameter

rldhont:
For OGC Requests, VERSION parameter is mandatory except for GetCapabilities requests.

* Funded by 3Liz

PR #2797, opened 2022-02-25
Fix: Extract Version from OGC XML Request

rldhont:
For OGC WXS, the version parameter is mandatory except for GetCapabilities request.
For OGC WXS, the request can be done with an XML Request. The root document contains a service attribute and can contain a version attribute.
The version has to be extracted from XML request to be checked before request processing.

* Linked to https://github.com/3liz/lizmap-web-client/pull/2769
* Funded by 3Liz

### DIFF
--- a/lizmap/modules/lizmap/lib/Request/Proxy.php
+++ b/lizmap/modules/lizmap/lib/Request/Proxy.php
@@ -86,6 +86,7 @@ class Proxy
     {
         $service = null;
         $request = null;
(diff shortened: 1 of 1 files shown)


rldhont:
Cypress green

rldhont:
Rebase to get cypress updated to 9.5.0 and Cypress tests are still green.

laurentj (approved):
lgtm

PR #2799, opened 2022-02-25
Fix: Use cache for QGIS Server plugins data

rldhont:
The request to QGIS Server to get plugins data is not QGIS Project dependant.
The requests have not to be perform each time lizmap needs to get the plugins list.

* Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisServer.class.php
+++ b/lizmap/modules/lizmap/classes/qgisServer.class.php
@@ -30,6 +30,13 @@ class qgisServer
 
     public function getPlugins($project)
     {
+        $key = 'qgis/server/plugins';
+        $key = jCache::normalizeKey($key);
+        $plugins = jCache::get($key);
+        if ($plugins !== false && $plugins !== null) {
+            return $plugins;
+        }
+
         $plugins = array();
 
         // Check Lizmap plugin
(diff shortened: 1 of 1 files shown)


Gustry:
OK because it's a bugfix.

Please consider these `GetServerSettings` and `GetCapabilitiesAtlas` as deprecated ;-)
The other class to fetch data from the lizmap plugin is the way to go.

rldhont:
Cypress red

rldhont:
Cypress red - Only *should select the single point on map which turns yellow* failed

mdouchin (approved):

rldhont:
Rebase to get cypress updated to 9.5.0 and Cypress tests are now green.

lizmap/modules/lizmap/classes/qgisServer.class.php
```diff
@@ -30,6 +30,13 @@ public function __construct()
 
     public function getPlugins($project)
     {
+        $key = 'qgis/server/plugins';
+        $key = jCache::normalizeKey($key);
+        $plugins = jCache::get($key);
+        if ($plugins !== false) {
```
laurentj, line 36: The implementation of the plugin "file" for jCache is not consistent with other plugins (my fault). It may return null instead of false if the cache does not exist. Prefer to check both `false` and `null`.

laurentj (changes requested):

lizmap/modules/lizmap/classes/qgisServer.class.php
rldhont: Checked that `$plugins` is not `null` and not `false`.

rldhont:
@laurentj changes done and cypress tests are still green

laurentj (approved):

PR #2800, opened 2022-02-28
QGIS - Display the server commit ID if available

Gustry:
if we run a compiled QGIS, it's nice to know which commit we are running from the LWC interface.


mdouchin (approved):
I am curious, why did you prefer the URL
`https://github.com/qgis/QGIS/search?q=30022b436d96446740c62b98fc4434ccf5a8b4ed&type=commits`
to
`https://github.com/qgis/QGIS/commit/30022b436d96446740c62b98fc4434ccf5a8b4ed`
?

Gustry:
I'm currently in a QGIS server branch, not belonging to the QGIS core repository.

Not sure what is the best : 404 or search result.

I think there is one URL where GitHub will look across fork.

PR #2815, opened 2022-03-01
[Bugfix] lizmap service GetProj4 request

rldhont:
The method getProj4 from project return string and not DOM element

* Funded by 3liz


rldhont:
Cypress is green

mdouchin (approved):

PR #2841, opened 2022-03-11
[Bugfix] Edition: update and delete is not authorized for feature id 0

rldhont:
The feature id can be equal to 0 and it is valid but the edition controller does not authorize edition in this case.

Also replace `and` by `&&` and `or` by `||`.

* Funded by 3Liz


laurentj (approved):
r+

PR #2843, opened 2022-03-11
Php-cs-fixer upgrade to 3.7.0

laurentj:
Php-cs-fixer 2.* is not supported anymore.

* Funded by 3liz

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,6 +17,6 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: "PHP-CS-Fixer"
-      uses: docker://oskarstark/php-cs-fixer-ga:2.19.0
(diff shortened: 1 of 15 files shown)


Gustry:
How will we handle the executable `php-cs-fixer` on your local machine ? (when going on master, 3.5 and 3.4 branch)
Does it work well when switching branch with the same php-cs-fixer binary ? Or should we use only the docker image ?

nboisteault (approved):

laurentj:
I think we should add a dev/composer.json file, and declare php-cs-fixer into it. So everybody have the same version.

rldhont (approved):

nboisteault:
@laurentj I did not remember your PR. I did that PR this morning https://github.com/3liz/lizmap-web-client/pull/2993
Does it superseed yours?

PR #2852, opened 2022-03-22
Revert old cache rules

nboisteault:
No more necessary since https://github.com/3liz/lizmap-web-client/pull/2729

### DIFF
--- a/lizmap/modules/view/controllers/translate.classic.php
+++ b/lizmap/modules/view/controllers/translate.classic.php
@@ -23,7 +23,6 @@ class translateCtrl extends jController
         $rep = $this->getResponse('binary');
         $rep->mimeType = 'text/javascript';
         $rep->doDownload = false;
-        $rep->setExpires('+1 year');
 
         $lang = $this->param('lang');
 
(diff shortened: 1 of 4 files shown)


nboisteault:
@laurentj I'm not sure about this change : https://github.com/3liz/lizmap-web-client/pull/2852/files#diff-2507844269c2b3d15c0d54a58d870595caca6fc3bcfad6e5e23e708204ee6833L26
Should it be handled by Lizmap backend or the web server?
With our docker test stack if I remove this line the asset is not cached and there is no etag. Maybe because there is `Content-Type: text/javascript;charset=UTF-8` and not `Content-Type: text/javascript` in header?

laurentj:
@nboisteault as it is a php script, there is no cache by default, so we should set it. However, one year of cache is too high. So we should not remove it, but change its value.

lizmap/modules/view/controllers/translate.classic.php
```diff
@@ -23,7 +23,6 @@ public function index()
         $rep = $this->getResponse('binary');
         $rep->mimeType = 'text/javascript';
         $rep->doDownload = false;
-        $rep->setExpires('+1 year');
```
laurentj, line 26: we should not remote it, but set a lower duration

laurentj (changes requested):

lizmap/modules/view/controllers/translate.classic.php
laurentj: 1 week is a good duration I think

laurentj:
I merge the PR, and will fix the cache duration, in order to have this patch in the release of the week

PR #2853, opened 2022-03-22
Filter by attribute: the attribute can be upper case

rldhont:
Remove the `strtolower` on `filterAttribute` because the field name can be upper case and it has to stay like it is defined in the database.

* Funded by Terre de Provence Agglomération


mdouchin (approved):
Thanks

nboisteault (approved):

PR #2874, opened 2022-03-25
[Tests] end2end and manually: coverage cypress form value relation tests

rldhont:
* Using cy.log instead of cy.context
* Extend cypress filter by geometry
* Extend cypress initial form state to cover manually tests
* Update manually tests because of cypress overage

* Funded by 3liz

### DIFF
--- a/tests/qgis-projects/tests/form_edition_value_relation_field.md
+++ b/tests/qgis-projects/tests/form_edition_value_relation_field.md
@@ -6,42 +6,9 @@ Lizmap plugin installed for QGIS Server is needed
 
 ## Procedure
 
-### Check the default form
-
-* [ ] Click on add a `point` in the *Edition* panel (do not click on the map yet)
-* [ ] Check that the *No expression* available values has **5** options :
-  * [ ] **Zone A1**
-  * [ ] **Zone A2**
-  * [ ] **Zone B1**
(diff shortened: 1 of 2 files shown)


tests/qgis-projects/tests/form_edition_value_relation_field.md
```diff
@@ -8,41 +8,16 @@ Lizmap plugin installed for QGIS Server is needed
...
-  * [ ] an empty value
-* [ ] Check that the *Simple expression* available values has **3** options :
-  * [ ] **Zone A1**
-  * [ ] **Zone B1**
-  * [ ] an empty value
-* [ ] Check that the *Child field* available values has **1** option :
-  * [ ] an empty value
-* [ ] Check that the *Geom expression* available values has **1** option
-  * [ ] an empty value
+Covered by Cypress
```
Gustry, line 11: Why not removing ?
Less thing to read among all these markdown file when doing the manual testing.

rldhont: Because still one part cannot be removed

Gustry: I'm not talking about the file itself.
But the section `### Check the default form` and `### Check the drill-down fields`

Issue #2770, opened 2022-02-11
attribute table -  doesn't export when a selection is active


josemvm:
### What is the bug?

as this comment https://github.com/3liz/lizmap-web-client/issues/2715#issuecomment-1031376587 by @rldhont 

_It is a regression in Lizmap, before 3.4.8 it was using EXP_FILTER and BBOX, now it uses FEATUREID instead of EXP_FILTER but FEATUREID and BBOX cannot be used in the same WFS request_

### Steps to reproduce the issue

make a selection in attribute and try to export to all formats

### Versions

 Lizmap Web Client : 3.4.9
 Lizmap plugin : 3.7.2
 QGIS Desktop : 3.16.16

### QGIS server version, only if the section above doesn't mention the QGIS Server version

3.16.16

### Operating system

Ubuntu 20.04

### Browsers

Firefox

### Browsers version

97

rldhont:
This issue has been fixed for Postgres Layer by this PR https://github.com/3liz/lizmap-web-client/pull/2928

josemvm:
@rldhont thanks! 👍

PR #2889, opened 2022-03-30
Add export button to featureToolBar

nboisteault:
* Ticket : support 1798
* Funded by: Les portes du soleil

### DIFF
--- a/assets/src/components/FeatureToolbar.js
+++ b/assets/src/components/FeatureToolbar.js
@@ -1,8 +1,12 @@
 import { mainEventDispatcher } from '../modules/Globals.js';
+import Utils from '../modules/Utils.js';
 import { html, render } from 'lit-html';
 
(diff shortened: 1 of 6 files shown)


nboisteault:
@mdouchin @Gustry  should I handle only GeoJSON, KML and GPX formats client-side as in digitizing tool or all formats with wfsoutputextension?

Gustry:
As discussed, GPX, KML are more appropriate for this kind of download.
Download as more GIS formats are still available in attribute table.

To not overload with the UI, I proposed a new button https://github.com/3liz/lizmap-plugin/pull/403#issuecomment-1082992430

Do not forget that the WfsOutputExtension is not required when running LWC (so less formats available)

josemvm:
_Download as more GIS formats are still available in attribute table._
https://github.com/3liz/lizmap-web-client/issues/2770
;-)

PR #2892, opened 2022-04-04
[Bugfix] Force to open GetPrint in blank window

rldhont:
Because some browser like Firefox opens PDF content in the internal PDF viewer, lizmap has to force to open GetPrint content in a blank window whixh is a new tab.

* Funded by 3Liz

### DIFF
--- a/lizmap/www/assets/js/map.js
+++ b/lizmap/www/assets/js/map.js
@@ -4709,22 +4709,7 @@ var lizMap = function() {
               } else {
                   var URL = window.URL || window.webkitURL;
                   var downloadUrl = URL.createObjectURL(blob);
-
-                  if (filename) {
-                      // use HTML5 a[download] attribute to specify filename
-                      var a = document.createElement("a");
-                      // safari doesn't support this yet
-                      if (typeof a.download === 'undefined') {
-                          window.location = downloadUrl;
-                      } else {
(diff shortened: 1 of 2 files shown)


nboisteault (approved):
If you have tested it works LGTM

nboisteault:
@rldhont I tried this code above. It opens in a new tab in Firefox. In chrome, you have to accept to open popups for the given domain then it downloads without displaying. We could detect browser is Firefox and use this code.

```js
      var form = document.createElement('form');
      form.setAttribute('method', 'POST');
      form.setAttribute('action', url);
      form.setAttribute('target', '_blank');
      // Ajout des paramètres sous forme de champs cachés
      for(var cle in printParams) {
        if(printParams.hasOwnProperty(cle)) {
          var champCache = document.createElement('input');
          champCache.setAttribute('type', 'hidden');
          champCache.setAttribute('name', cle);
          champCache.setAttribute('value', printParams[cle]);
          form.appendChild(champCache);
        }
      }
      // Ajout du formulaire à la page et soumission du formulaire
      document.body.appendChild(form);
      form.submit();
```

rldhont:
Is window.open not the right solution for every browser ? If the backend provides the right filename ?

nboisteault (approved):

mdouchin (approved):
lgtm

Issue #2767, opened 2022-02-10
JSON in dataviz


aeduard:
### Description
I can't manage to override the default graph parameters in Dataviz Layout. I have a scatter plot generated form 1:N relation, without any aggregation. The graph is displayed as filtered in the parent layer's popup.
Example:
`{"xaxis": {"title": {"text": "Any text"}}}`

I've also tried the sample codes from the lizmap [docs page](https://docs.lizmap.com/current/en/publish/lizmap_plugin/dataviz.html#json-layout).

When displaying the graph from the main dataviz panel, it works. A similar issue was raised in august 2021 in the lizmap mailing list by Immo Blecher.

### Environment

 - Lizmap version: 3.4.6
 - QGIS Server FULL version: 3.16
 - QGIS Desktop FULL version: 3.16
 - OS (Windows, Linux, MacOS, Android…): Ubuntu 20.4
 - Browser (Firefox, Chrome…): Opera

Gustry:
Linked to #2687 ?

rldhont:
#2897 has been merged in release_3_5, the issue will be fixed when the backport to release_3_4 will be merged.

Issue #491, opened 2017-02-07
LWC tools should respect the layers order defined by the plugin


josemvm:
- locate layer
- attribute table
- tooltip layers
- selection
...

it seems to me that the layers order is random within the respective tools dock...

josemvm:
moreover, in the case of the tooltip the first layer of the list could be the default layer (*) to be displayed on the tool dock

(*) or there is a possibility in the plugin to set the default layer

mdouchin:
order respected for 

- [ ] edition
- [ ] attribute table
- [ ] locate by layer
- [ ] tooltip layers
- [ ] selection

josemvm:
hi @mdouchin @rldhont 

could you please fixe the next item (or items)?
is very useful when it comes to an extensive and thematic variety list

thank you in advance for your support

josemvm:
hi

please, could you fixe the next item (**attribute table**)?
it's very useful for organizing the layers list by themes

thanks in advance

josemvm:
just for remember you @rldhont @mdouchin :-)

nboisteault:
@josemvm layers order in attribute table works ok for me. Can you test ?

josemvm:
hi @nboisteault 

look at montpellier project:

![imagem](https://user-images.githubusercontent.com/10053874/46278800-264dd480-c55f-11e8-8cff-be8caf90ce67.png)

![imagem](https://user-images.githubusercontent.com/10053874/46278855-572e0980-c55f-11e8-817f-7d6e5d1ef66e.png)

thanks

nboisteault:
@josemvm You're right.
"attribute table" fixed by https://github.com/3liz/lizmap-web-client/pull/978.
Can you test it ?

nboisteault:
Locate by layer works thanks to this : https://github.com/3liz/lizmap-web-client/blob/bcf2a63e5ec552a46163cf2b5da851233fd867c9/lizmap/www/js/map.js#L2592

nboisteault:
Layer order in every tools has been fixed.
@josemvm : please test and tell us if it is ok. Thanks.

josemvm:
thank you so much @nboisteault 
everythin works fine!!! :-)

if you can take a look at time manager tool...

josemvm:
hi @nboisteault 

no, that's all!
go ahead on time manager and dataviz module 👍 

thanks

rmarzocchi84:
[...]

PR #2897, opened 2022-04-06
Bugfix dataviz popup

tethysgeco:
* Ticket : 
* #2767 --> plot in popup maintains json customization defined in lizmap plugin
* #491 --> plots order in popup respects plot order defined in lizmap plugin
* Tethys S.r.l. www.tethys.srl

### DIFF
--- a/lizmap/www/assets/js/dataviz/dataviz.js
+++ b/lizmap/www/assets/js/dataviz/dataviz.js
@@ -184,7 +184,8 @@ var lizDataviz = function() {
                 }
 
                 // Build plot
-                var plot = buildPlot(target_id, json);
+                // Pass plot_id to inherit custom configurations in child charts
+                var plot = buildPlot(target_id, json, plot_id);
                 $('#'+target_id).prev('.dataviz-waiter:first').hide();
             }
         );
@@ -295,7 +296,7 @@ var lizDataviz = function() {
         return pid;
     }
 
-    function buildPlot(id, conf){
+    function buildPlot(id, conf, pid = null){
         // Build plot with plotly or lizmap
         if(conf.data.length && conf.data[0]['type'] == 'html'){
             buildHtmlPlot(id, conf.data, conf.layout);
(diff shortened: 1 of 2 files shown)


Gustry:
Thanks for the contribution. I will let @mdouchin check this PR

rldhont:
Thanks @tethysgeco 

@nboisteault do cypress tests cover this code ?

nboisteault:
@rldhont No, dataviz part lacks tests.

lizmap/www/assets/js/dataviz/dataviz.js
```diff
@@ -295,7 +296,7 @@ var lizDataviz = function() {
         return pid;
     }
 
-    function buildPlot(id, conf){
+    function buildPlot(id, conf, pid = null){
```
rldhont, line 299: This change uses [Default parameters in function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters). The [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters#browser_compatibility) seems OK with Lizmap 3.5 [browser compatibility](https://github.com/3liz/lizmap-web-client/wiki/Versions#web-browsers).

lizmap/www/assets/js/dataviz/dataviz.js
```diff
@@ -295,7 +296,7 @@ var lizDataviz = function() {
         return pid;
     }
 
-    function buildPlot(id, conf){
```
mdouchin, line 298: Please change also the definition of the `buildPlot` function at the end of the file to integrate this new parameter, after
`var obj = {`

mdouchin (changes requested):

lizmap/www/assets/js/dataviz/dataviz.js
rldhont: @mdouchin I have done it.

PR #2904, opened 2022-04-11
[Bugfix] Request WFS: Typename or feature id are mandatory

rldhont:
Lizmap Web Client is checking TYPENAME parameter, but TYPENAME is not mandatory if FEATUREID is defined.

* Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/lib/Request/WFSRequest.php
+++ b/lizmap/modules/lizmap/lib/Request/WFSRequest.php
@@ -20,6 +20,11 @@ class WFSRequest extends OGCRequest
     protected $tplExceptions = 'lizmap~wfs_exception';
 
     /**
+     * @var null|string the requested typename
+     */
+    protected $wfs_typename;
+
+    /**
      * @var qgisMapLayer|qgisVectorLayer
      */
     protected $qgisLayer;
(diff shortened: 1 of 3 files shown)


nboisteault:
Cypress fails in feature_toolbar test because `layerConfig.types` is `undefined`.

https://github.com/3liz/lizmap-web-client/blob/master/lizmap/www/assets/js/attributeTable.js#L1980

rldhont:
@nboisteault cypress is green

lizmap/modules/lizmap/lib/Request/WFSRequest.php
```diff
@@ -238,9 +274,9 @@ protected function process_getfeature()
         }
 
         // Get type name
-        $typename = $this->param('typename');
+        $typename = $this->requestedTypename();
         if (!$typename) {
-            \jMessage::add('TYPENAME is mandatory', 'RequestNotWellFormed');
+            \jMessage::add('TYPENAME or FEATUREID are mandatory', 'RequestNotWellFormed');
```
nboisteault, line 279:
```suggestion
            \jMessage::add('TYPENAME or FEATUREID is mandatory', 'RequestNotWellFormed');
```

tests/end2end/cypress/integration/requests-service-ghaction.js
```diff
@@ -173,4 +220,22 @@ describe('Request service', function () {
             expect(resp.body).to.contain('WFS_Capabilities')
         })
     })
+
+    it('TYPENAME or FEATUREID are mandatory for WFS GetFeature request', function () {
```
nboisteault, line 224:
```suggestion
    it('TYPENAME or FEATUREID is mandatory for WFS GetFeature request', function () {
```

nboisteault (approved):
Not sure but if think `is` is correct not `are`.

PR #2906, opened 2022-04-11
[Bugfix] WFS GetFeature request from Postgres - add BBox

rldhont:
The Feature Bounding box was missing when lizmap building the GeoJSON from Postgres.

* Funded by 3liz


nboisteault (approved):
OK for me when Cypress is ok.

PR #2931, opened 2022-04-19
[Bugfix] Jelix jForms jQuery dynamicFillAjax - keep value

rldhont:
If the content of dynamic fill is updated, the form looses the value.
In Lizmap case, the content of a fill is updated by the geometry and we loose the value set by the PHP part.

* Funded by Ville d'Avignon


nboisteault (approved):

laurentj (approved):
r+

laurentj:
The patch has been backported upstream

PR #2956, opened 2022-04-26
[Bugfix] GetSeletionToken and GetFilterToken request

rldhont:
The methods to get selection or filter token return array not object.

The IDS for selection token has to be trim

Enhancing GetSelectionToken and GetFilterToken with ETag header

* Funded by 3liz

PR #2957, opened 2022-04-26
[Bugfix] attribute table - doesn't export when a selection is active

### LINKED ISSUE #2770: attribute table -  doesn't export when a selection is active

rldhont:
fixes #2770

In the WFS OGC standard FEATUREID and BBOX parameters cannot be mutually set
but in Lizmap, the user can do a selection, based on featureid, and can request
a download, a WFS GetFeature request, based on this selection with a restriction
to map extent, sofeatureid and bbox parameter can be set mutually and featureid
parameter needs to be transform in an expression filter.

A parameter `FORCE_QGIS` has been introduced for GetFeature to compare lizmap
postgres GetFeature response and QGIS Server GetFeature response.

* Funded by 3Liz

### DIFF
--- a/lizmap/modules/lizmap/lib/Request/WFSRequest.php
+++ b/lizmap/modules/lizmap/lib/Request/WFSRequest.php
@@ -473,6 +473,7 @@ class WFSRequest extends OGCRequest
         }
 
         // Building the expression filter
+        $expFilter = '';
         if ($provider == 'postgres') {
             // for postgres layer we can build the expression filter for
             // simple and multi fields key
@@ -480,7 +481,7 @@ class WFSRequest extends OGCRequest
             $keys = preg_split('/\\s*,\\s*/', $dtparams->key);
             if (count($keys) == 1 && !$hasDoubleAtSign) {
                 // for simple field key
-                return '"'.$keys[0].'" IN ('.implode(', ', $pks).')';
+                $expFilter = '"'.$keys[0].'" IN ('.implode(', ', $pks).')';
             }
             if (count($keys) > 1 && $hasDoubleAtSign) {
                 // for multi fields key
@@ -498,11 +499,15 @@ class WFSRequest extends OGCRequest
                     $filters[] = $filter;
                 }
 
-                return implode(' OR ', $filters);
+                $expFilter = implode(' OR ', $filters);
             }
         } else {
             // for other layers with simple field key
-            return '$id IN ('.implode(', ', $pks).')';
+            $expFilter = '$id IN ('.implode(', ', $pks).')';
+        }
+
+        if ($expFilter && $this->validateExpressionFilter($expFilter)) {
+            return $expFilter;
         }
 
         return '';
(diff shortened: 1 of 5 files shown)


lizmap/modules/lizmap/lib/Request/WFSRequest.php
```diff
@@ -346,6 +348,115 @@ protected function process_getfeature()
...
+        // parameter needs to be transform in an expression filter.
+        // The transformation is only available if the QGIS layer has been set.
+        if ($this->param('featureid')
+            && $this->param('bbox')
+            && $this->qgisLayer) {
+            $provider = $this->qgisLayer->getProvider();
+
+            $typename = $this->requestedTypename();
+            $featureid = $this->param('featureid', '');
+
```
laurentj, line 364: I suggest to create a new method, `$expFilter = $this->getExpFilter($featureId, $typename)`, so getfeatureQgis will be more readable and not too long

lizmap/modules/lizmap/lib/Request/WFSRequest.php
```diff
@@ -346,6 +348,115 @@ protected function process_getfeature()
...
+            $exp_filter = '';
+            if (($hasDoubleAtSign && $provider == 'postgres')
+                || !$hasDoubleAtSign) {
+                $fids = explode(',', $featureid);
+
+                // get primary keys values
+                $pks = array();
+                foreach ($fids as $fid) {
+                    $exp = explode('.', $fid);
+                    if (count($exp) == 2 and $exp[0] == $typename) {
```
mdouchin, line 379: We should control which data is given here, because it comes from a parameter, and could be used for SQL injections. QGIS Server probably does the job, but we could perhaps use a small regex to check if the values is simple enough ?

lizmap/modules/lizmap/lib/Request/WFSRequest.php
laurentj: (and you could unit test this method)

rldhont:
@laurentj I have created a specific method and added a unit test.

lizmap/modules/lizmap/lib/Request/WFSRequest.php
rldhont: Wich regExp do you propose ?

rldhont:
All cypress tests passed, to avoid error logs we need https://github.com/3liz/lizmap-web-client/pull/2956

lizmap/modules/lizmap/lib/Request/WFSRequest.php
mdouchin: The method `validateFilter` seems to do the Job. Could you perhaps use it in this context ?

rldhont: `validateFilter is for postgres layer and update the expression to SQL, so I have introduced a `validateExpressionFilter` which is strict because of blockedWords. Key values cannot contain `;` or `--`.

mdouchin (approved):

laurentj (approved):
r+

Issue #2996, opened 2022-05-13
[Bug]: Lizmap 3.5.3 - Error 500


sergiuspinheiro:
### What is the bug?

Hi
I upgraded from lizmap 3.4.8 to lizmap 3.5.3 and I get the following error:

```
lizmap-web-client-3.5.3/lizmap/var/log/errors.log

2022-05-12 15:31:28	127.0.0.1	error	2022-05-12 15:31:28	[0]	syntax error, unexpected ')'	/var/www/lizmap-web-client-3.5.3/lizmap/modules/filter/controllers/service.classic.php	89

2022-05-12 15:31:28	127.0.0.1	error	2022-05-12 15:31:28	[0]	syntax error, unexpected ')'	/var/www/lizmap-web-client-3.5.3/lizmap/modules/dataviz/controllers/service.classic.php	69
```

### Steps to reproduce the issue

When opening a project with Filters and/or Dataviz.
Example in browser I get the error 500
![imagem](https://user-images.githubusercontent.com/52661230/168271453-2ce893f3-eda2-4428-9cd2-66203f0c4057.png)

### Versions

**Versions :**

* Lizmap Web Client : 3.5.3
* Lizmap plugin : 3.7.7
* QGIS Desktop : 3.16.3

Installation on ubuntu 18.04.6
![imagem](https://user-images.githubusercontent.com/52661230/168271562-ee36edbb-9102-4e74-99eb-d70294e4699e.png)

### QGIS server version, only if the section above doesn't mention the QGIS Server version

3.16.3

### Operating system

Ubuntu 18.04.6

### Browsers

Firefox

### Browsers version

Firefox 100.0

### Relevant log output

```shell
lizmap-web-client-3.5.3/lizmap/var/log/errors.log

2022-05-12 15:31:28	127.0.0.1	error	2022-05-12 15:31:28	[0]	syntax error, unexpected ')'	/var/www/lizmap-web-client-3.5.3/lizmap/modules/filter/controllers/service.classic.php	89

2022-05-12 15:31:28	127.0.0.1	error	2022-05-12 15:31:28	[0]	syntax error, unexpected ')'	/var/www/lizmap-web-client-3.5.3/lizmap/modules/dataviz/controllers/service.classic.php	69
```

Gustry:
What is your exact workflow ?
I couldn't replicate the issue so far.

Side note : 
[...]

PR #2997, opened 2022-05-13
Demo - Upgrade projects to 3.16 and Lizmap 3.7.7

Gustry:
While trying #2996 : 

* Update projects to 3.16 (from 3.4 ...)
* Update CFG
* ~Remove old unused QFTS file from QGIS 2, ok ?~


Gustry:
Unrelated phpstan error (maybe because of a new version of photostan 1.7.1)

PR #3002, opened 2022-05-17
Fix cmdline wmts seeding bbox

rldhont:
* WMTS command line: add `-dry-run` option to seeding
* [Tests] Add capabilities to run lizmap command line
* [Tests] end2end: new WMTS command line script
* [Bugfix] WMTS command line: seeding -bbox out of the tile matrix limit
* Skipping some e2e tests when continuous_integration env is set
* [Bugfix] WMTS calculate tiles
* [Tests] Add capabilities to run `docker exec` with lizmap-ctl
* [Tests] Continuous integration for cache manual tests

Fixes: #1469
Funded by 3Liz

PR #3005, opened 2022-05-19
GH Action - Allow running Cypress with some PHP errors

Gustry:
* GH Action - Allow running Cypress with some PHP errors
* Funded by 3Liz


rldhont:
If new cypress tests in #3002 will not failed, this PR could be closed.

PR #3014, opened 2022-05-30
[Bugfix] default attribute table column order not respected

### LINKED ISSUE #2780: [Bug]: default attribute table column order not respected
### What is the bug?

Default attribute table column order is not respected. The order is always alphabetical.
This seems related to a change in QGIS Desktop/Server https://github.com/qgis/QGIS/issues/41290 .
It is not mandatory for GeoJSON to have `properties` in the same order as defined in QGIS so we need another source of truth concerning order.

A workaround is in QGIS to right click the columns in attribute table and `Organize columns`.

1. Open attribute table with this project

[wrong_wfs_geojson_attributes_order.zip](https://github.com/3liz/lizmap-web-client/files/8072885/wrong_wfs_geojson_attributes_order.zip)

### Versions

**Versions :**

* Lizmap Web Client : 3.6.0-pre
* Lizmap plugin : master
* QGIS Desktop : 3.22.1
* QGIS Server and plugins unknown status

3.22.3

### Operating system

Ubuntu 18.04

### Browsers

Chrome

### Browsers version

97

### Relevant log output

_No response_

rldhont:
Fixes #2780

Default attribute table column order is not respected. The order is always alphabetical.
This seems related to a change in QGIS Desktop/Server https://github.com/qgis/QGIS/issues/41290 .
It is not mandatory for GeoJSON to have properties in the same order as defined in QGIS so we need another source of truth concerning order.

The DescribeFeatureType can provide the columns order because it is based on an XML which provides the list of attributes in the datasource order.

* Ticket : #2780
* Funded by Terre de Provence Agglo


lizmap/www/assets/js/attributeTable.js
```diff
@@ -1462,6 +1463,32 @@ var lizAttributeTable = function() {
                         }
                     }
 
+                    var newcolumnsfinal = lizcols.concat(newcolumns)
+                    colToReturn['columns'] = newcolumnsfinal;
+                } else if (
+                    'columns' in config.layers[aName]
+                    && config.layers[aName]['columns']
+                    && !$.isEmptyObject(config.layers[aName]['columns'])
```
nboisteault, line 1471:
```suggestion
                    && Object.keys(config.layers[aName]['columns']).length > 0)
```

nboisteault: We should stop using jQuery.

nboisteault (changes requested):

lizmap/www/assets/js/attributeTable.js
rldhont: @nboisteault I have to learn more about new JavaScript ;-)

rldhont:
Thanks for the fix and the tests @nboisteault

nboisteault (approved):

Gustry:
Slightly too late, but just for your information, I prefer having text based layer instead of shapefile for testing, like geojson for instance (or other text based layer) : 

* we are using git, so the diff on text based layer is way easier and made for git
* geojson is a single file, there isn't any side car files
* geojson can be viewed online, like https://github.com/3liz/lizmap-plugin/blob/master/lizmap/test/data/lines.geojson
* we can keep the file simple for a small use case : https://raw.githubusercontent.com/3liz/lizmap-plugin/master/lizmap/test/data/lines.geojson
* it's not a binary file, better to follow diff, or no file replacement in the git history

Just a few notes for later :)

nboisteault:
@Gustry I can change it but @rldhont will a GeoJson file induces the bug like a shapefile ?

Gustry:
Do not worry, the file is already gitified now, so we can keep it.
It's just for later ;-)

PR #3029, opened 2022-06-02
[Bugfix] Dataviz: dock is provided even if graphics are only in popup

rldhont:
* Funded by 3Liz


Gustry:
So it removes the first checkbox ? 
https://github.com/3liz/lizmap-web-client/blob/master/tests/qgis-projects/tests/dataviz_filtered_in_popup.md

Nice.
LGTM

rldhont:
> So it removes the first checkbox ? https://github.com/3liz/lizmap-web-client/blob/master/tests/qgis-projects/tests/dataviz_filtered_in_popup.md
> 
> Nice. LGTM

Yes it tries to remove the first check box

Gustry (approved):

PR #3020, opened 2022-06-01
Tests - Remove the manual test about filtered popup

Gustry:
Yet another markdown file removed

* Funded by 3Liz

PR #3030, opened 2022-06-02
Tests - Add common functions for cypress

Gustry:
Subset of #3020 

@nboisteault What do you think about this `mapClick` command ?

### DIFF
--- a/tests/end2end/cypress/support/commands.js
+++ b/tests/end2end/cypress/support/commands.js
@@ -56,3 +56,13 @@ Cypress.Commands.add(
         cy._login(Cypress.env('login_user_a'), Cypress.env('password_user_a'))
     }
 )
+
+Cypress.Commands.add(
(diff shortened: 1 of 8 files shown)


tests/end2end/cypress/support/commands.js
```diff
@@ -56,3 +56,9 @@ Cypress.Commands.add(
         cy._login(Cypress.env('login_user_a'), Cypress.env('password_user_a'))
     }
 )
+
+Cypress.Commands.add(
+    'mapClick', (x, y) => {
+        cy.get('#map').click(x - 30, y - 75)
```
rldhont, line 62: Why `x - 30` and `y - 75` ?

rldhont: Thanks @Gustry

rldhont (approved):

PR #3040, opened 2022-06-07
[Bugfix] infinite redirections when no access to the default project in the 'onlyMap' mode

### LINKED ISSUE #3003: [Bug]:  infinite redirections when no access to the default project in the 'onlyMap' mode
### What is the bug?

When the onlyMap mode is activated, the user is redirected from the `view~default:index` action to the `view~map:index` action. But if something is wrong into this controller (especially into the lizmapCtrl controller), the user is redirected to the `view~default:index`. So there are only redirections and after a number of redirection the browser stops and display an error.

The `view~map:index` action does a redirection to `view~default:index` when the `checkAcl` on the default project is falling, or when the default project is hidden.

Solution : before doing these redirections, the onlyMap mode should be checked, and if it is activated, an error should be directly displayed, instead of doing a redirection.

1. set the onlyMap mode
2. set a default project
3. set acl groups on the project (with qgis desktop)
4. login with a user who is not in groups selected for acl
5. infinite redirection occurs

### Versions

Lizmap 3.5 and lizmap 3.4

_No response_

### Operating system

any

### Browsers

Firefox

### Browsers version

any

### Relevant log output

_No response_

rldhont:
fixes #3003

In case of OnlyMaps, the ACL was not check.

Funded by 3Liz

### DIFF
--- a/lizmap/modules/view/controllers/default.classic.php
+++ b/lizmap/modules/view/controllers/default.classic.php
@@ -35,7 +35,7 @@ class defaultCtrl extends jController
             if ($repository && jAcl2::check('lizmap.repositories.view', $repository->getKey())) {
                 try {
                     $project = lizmap::getProject($repository->getKey().'~'.$services->defaultProject);
-                    if ($project) {
+                    if ($project && $project->checkAcl()) {
                         // test redirection to an other controller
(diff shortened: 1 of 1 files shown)


rldhont:
@laurentj gentle ping ?

laurentj (approved):

PR #3046, opened 2022-06-09
[Tests] e2e: check config.options.hideProject: "True"

rldhont:
New tests to check the lizmap config options hide project.
The project is empty: no layers, no other config.

Funded by 3Liz


Gustry (approved):
LGTM,

Side question, when is this option useful because the project is not visible for anyone ?
I thought it was only to make it hidden in the landing page but the project would still be accessible with a direct URL... but it seems it's not the case.

rldhont:
This option is made to hide project in the landing page and provide only OGC services.

nboisteault (approved):

nboisteault:
@rldhont PHPStan has been updated to https://github.com/phpstan/phpstan/releases/tag/1.7.12 . That's why it has failed.

rldhont:
> This option is made to hide project in the landing page and provide only OGC services.

It also avoid map view, only services.

PR #3051, opened 2022-06-10
Minify JS files

nboisteault:
Use terser in Webpack to minify our JS files.
That saves more than 400Ko!
before terser : 649,6 Ko
after terser : 246,7 Ko

Funded by 3Liz

### DIFF
--- a/assets/webpack.common.js
+++ b/assets/webpack.common.js
@@ -2,7 +2,19 @@ const path = require('path');
 
 module.exports = {
     entry: {
-        lizmap: './src/index.js'
+        lizmap: './src/index.js',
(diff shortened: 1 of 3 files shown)


rldhont:
Can the commit _Declare explicitly lizMap and lizEditionErrorDecorator as global variables_ db4b6cb9ad94e6b18f5b051bf094b409b65b2f09 be backoprted to release_3_5  to build cypress tests ?

nboisteault:
I merge, we'll see if files are correctly minified in https://packages.3liz.org/pub/lizmap/unstable/3.6/

PR #3060, opened 2022-06-17
CI - Try PHP 8 when running phpunit

Gustry:


rldhont:
linked to https://github.com/3liz/lizmap-web-client/pull/3063 ?

Issue #925, opened 2018-07-11
Lizmap must use a semantically correct json 


dmarteau:
### What is the bug?
Lizmap and Lizmap plugin produre and consumme a jso which is not semantically correcte (even 
if the syntax is ok)

### Steps to reproduce the issue
- Generate a Lizmap configuration with the Lizmap plugin.
- Look at entry like: `{ "toggled": "True" }`

### Expected behavior
Boolean values should be `true` or `false` not `"True"` or `"False"`

### Environment
 - Lizmap version:  3.2+
 - QGis version: 3+
 
This is a problem if one try to parse and use the Lizmap configuration file on a api basis: as most all the languages have modules to handle json correctly, code that manipulate configuration file has to tweak the logic deal with the ambiguity of string value vs boolean value.

rldhont:
This issue has to be fixed in 3.1.x and here the JS code and PHP has to well understand boolean.

dmarteau:
Now 3.2 is out, what is the status ?

nboisteault:
@rldhont @dmarteau will we fix this for 3.3 ?

rldhont:
@nboisteault it has to be fixed in lizmap 4

Gustry:
Linked to https://github.com/3liz/lizmap-plugin/issues/176

rldhont:
It will be fixed when a lot of legacy JS and PHP code will be replaced by a new API to use the config.

PR #3073, opened 2022-06-21
JavaScript Lizmap Config Parser

rldhont:
To wrap the Lizmap config with classes and simplify the JavaScript codes of Lizmap. 

Funded by 3Liz

To get access to the config, it is available through `lizMap.mainLizmap.initialConfig` so you can do:
```js
// get options config instance
lizMap.mainLizmap.initialConfig.options

// get layers config instance
lizMap.mainLizmap.initialConfig.layers
// The layer names
lizMap.mainLizmap.initialConfig.layers.layerNames
// The layer ids
lizMap.mainLizmap.initialConfig.layers.layerIds
// iterate through base layers config
for (const layerCfg of lizMap.mainLizmap.initialConfig.layers.getLayerConfigs()) {
    // The layer type: layer, group
    layerCfg.type
    // The layer name
    layerCfg.name
    // The layer id: it is the same has name for group instead it is the qgis layer id
    layerCfg.id
    // The base layer title
    layerCfg.title
}
// Get layer config by layer name
lizMap.mainLizmap.initialConfig.layers.getLayerConfigByLayerName('null_island OGRGeoJSON Point')
// Get layer config by layer id
lizMap.mainLizmap.initialConfig.layers.getLayerConfigByLayerId('null_island20200414115730489'))

// get base layers config instance
lizMap.mainLizmap.initialConfig.baseLayers
// The stratup base layer name
lizMap.mainLizmap.initialConfig.baseLayers.startupBaseLayerName
// The base layer names
lizMap.mainLizmap.initialConfig.baseLayers.baseLayerNames
// iterate through base layers config
for (const baseLayerCfg of lizMap.mainLizmap.initialConfig.baseLayers.getBaseLayerConfigs()) {
    // The base layer type: empty, xyz, bing, wmts, lizmap
    baseLayerCfg.type
    // The base layer name
    baseLayerCfg.name
    // The base layer title
    baseLayerCfg.title
}

// and the other config
```


assets/src/modules/config/Theme.js
```diff
@@ -0,0 +1,257 @@
+import { BaseObjectConfig } from './Base.js';
+import { ValidationError } from './../Errors.js';
+
+const layerThemeRequiredProperties = {
+    'style': {type: 'sting'},
```
nboisteault, line 5: `string`

nboisteault: TypeScript is used?

rldhont: @nboisteault no typescript only javascript.

assets/src/modules/config/Metadata.js
```diff
@@ -0,0 +1,39 @@
+import { BaseObjectConfig } from './Base.js';
+
+const optionalProperties = {
+    'lizmap_plugin_version_str': {type: 'string', default: '3.1.8'},
```
Gustry, line 4:
```suggestion
    // Lizmap plugin 3.1.8 is the minimum version when these keys were included in the CFG file
    // At that time, it was LWC 3.2 the stable version.
    'lizmap_plugin_version_str': {type: 'string', default: '3.1.8'},
```

Gustry (commented):
Do you know which tickets this PR might fix/be related to ? There might be a few maybe.
In the middle term, it will close #925 I guess ?

assets/src/modules/config/Metadata.js
rldhont: Fixes

rldhont:
All tests passed!

PR #3074, opened 2022-06-21
Map's initial extent defined by features. Optionally display features popup.

nboisteault:
When the `filter` and `popup` parameters are set in URL, map will zoom to defined features and display popup at startup

Funded by [natur&ëmwelt](https://www.naturemwelt.lu/)


rldhont:
@nboisteault needs tests ?

rldhont:
@nboisteault cypress failed to access lizmap on GitHb Action.

Issue #2694, opened 2022-01-17
[Bug]: form filtering seems to associate selected unique values with AND operator


Mavialle:
### What is the bug?

I apply a form filtering on PostgreSQL layer with input type unique values and specify a splitter text.

In Lizmap project, when I select several values, filter seems to be :

`( "field_thematique" LIKE '%value1%'  AND "field_thematique" LIKE '%value2%'  )`

whereas in documentation, it's OR operator who associate selected values (see [here](https://docs.lizmap.com/3.4/en/publish/lizmap_plugin/form_filtering.html?highlight=filter#unique-values)).

In other words, when I select two values, only points with this two values are displayed on map when it should be all points with at least one of these two values.

### Steps to reproduce the issue

1. Prepare a QGIS project with one PostgreSQL layer having at least a primary key and a field with unique values separate with commas
2. In QGIS project proprieties,  published the PostgreSQL layer as WFS
3. Specify with Lizmap extension a form filtering with following parameters : 
    - layer : the PostgreSQL layer opened in step 1
    - titre : at your convenience
    - type : unique values
    - field : field with unique values separate with commas
    - format : check box
    - splitter text : ,
4. Drop the project and cfg file on the Lizmap instance
5. In Lizmap project, in filter panel, select at least two values

### Lizmap version

3.4.7

### QGIS desktop version

3.10.4

### QGIS server version

3.10.4

### Operating system

Ubuntu

### Browsers

Firefox, Chrome

### Browsers version

Chrome 97.0.4692.71, Firefox 96.0.1

mdouchin:
fixed in #3078

PR #3078, opened 2022-06-22
Form Filter - Use the full cascading filter when possible

mdouchin:
This commit allows to set the 'best' method for triggerring the form filter,
depending on the avaibility of the attribute table configuration
for the filtered layer.

Two methods have been introduced:

* 'simple' means the same behaviour like before 3.6: filter only the layer and
  do not cascade to children or refresh other tools
* 'full' means we first get the features ids for the filter and then trigger the
  Lizmap filter based on the attribute table methods. This will cascade the filter
  to child layers .

The new 'full' method will help to filter many children layers when applying a filter
to the parent layer.

Beware this full method will have lower performance for heavy layers (more than 10000 features).

You can force the method in the JS console with: `filterConfigData.filterMethod = 'simple';`

Funded by: TO COMPLETE


rldhont:
@mdouchin do we have to run cypress ? do we have to add tests ?

nboisteault:
@mdouchin Cypress tests are needed. You can write manual tests and I'll Cypress them if you wish.

mdouchin:
I ran the tests locally and all are green. Do you want me to add a test with a shortname layer in `dataviz-ghaction.js`

nboisteault:
@mdouchin yes! Thanks.

rldhont (approved):

mdouchin:
@nboisteault @rldhont test added and Cypress is green. I will create a new PR for fixing #2694

nboisteault (approved):

PR #3085, opened 2022-06-24
Tests - Add one more about time manager

Gustry:
Do we think the MD file is covered by this cypress test ?

There are still some TODO if we want to continue, but I'm not sure about this token

* Test not intercepting specific request ? @nboisteault I didn't find a way to do it
* Find a way to test the correct request sent to QGIS server, we don't want to test the GetMap response, @rldhont @mdouchin Do you think you can provide the EXP_FILTER (or similar) which is send to QGIS server ?
* Test the FILTERTOKEN, the POST request ?

Funded by 3Liz


Gustry:
As QGIS server response are going through Lizmap Web Client, do you think, **for debug/testing only** (with environment variable or setting), we could add in the response from LWC, in the header the URL GET request which has been used to call QGIS server ?
This would allow in Cypress to check the `GetMap` with the `EXP_FILTER` which has been used in the backend ?

In production, we would'nt have this extra header. It's only for testing in Cypress the generated GetMap request.

What do you think ?

rldhont (changes requested):

rldhont (approved):

PR #3087, opened 2022-06-24
Tests - Add form filter cypress test

Gustry:
Tests - Add form filter cypress test

Funded by 3Liz


Gustry:
Unrelated test failure about IGN GetCapabilities.
We should use WMS from our own Lizmap stack, from another QGIS project, not external WMS.

rldhont (approved):

PR #3089, opened 2022-06-27
Cleanly handle project loading error at startup

nboisteault:
No more "SERVICE NON DISPONIBLE!"  🎉

@Gustry @mdouchin @rldhont Do you have any better proposition for the error message?

Funded by 3Liz
error-at-startup

### DIFF
--- a/assets/src/legacy/map.js
+++ b/assets/src/legacy/map.js
@@ -4672,28 +4672,6 @@ window.lizMap = function() {
   }
 
   /**
-   * PRIVATE function: parseData
-   * parsing capability
-   *
-   * Parameters:
-   * aData - {String} the WMS capabilities
-   *
-   * Returns:
-   * {Boolean} the capability is OK
-   */
-  function parseData(aData) {
-    var format =  new OpenLayers.Format.WMSCapabilities({version:'1.3.0'});
-    capabilities = format.read(aData);
-
-    var capability = capabilities.capability;
-    if (!capability) {
-      $('#map').html('SERVICE NON DISPONIBLE!');
-      return false;
-    }
-    return true;
-  }
-
-  /**
    * PRIVATE function: loadProjDefinition
    * load CRS definition and activate it
    *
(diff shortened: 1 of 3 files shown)


assets/src/legacy/map.js
```diff
@@ -6017,7 +5995,8 @@ window.lizMap = function() {
       });
```
rldhont, line 5995: I think you have to update the upper code. I think you have to check if the response is OK (status in the range 200–299):
```js
      // Get config
      const configRequest = fetch(OpenLayers.Util.urlAppend(lizUrls.config, OpenLayers.Util.getParameterString(lizUrls.params))).then(function (response) {
        if (!response.ok) {
            throw 'Config not loaded: '+response.status+' '+response.statusText
       }
        return response.json()
      });
      // Get key/value config
      const keyValueConfigRequest = fetch(OpenLayers.Util.urlAppend(lizUrls.keyValueConfig, OpenLayers.Util.getParameterString(lizUrls.params))).then(function (response) {
        if (!response.ok) {
            throw 'Key/value config not loaded: '+response.status+' '+response.statusText
       }
        return response.json()
      });
      // Get WMS, WMTS, WFS capabilities
      const WMSRequest = fetch(OpenLayers.Util.urlAppend(service, OpenLayers.Util.getParameterString({ SERVICE: 'WMS', REQUEST: 'GetCapabilities', VERSION: '1.3.0' }))).then(function (response) {
        if (!response.ok) {
            throw 'WMS GetCapabilities not loaded: '+response.status+' '+response.statusText
       }
        return response.text()
      });
      const WMTSRequest = fetch(OpenLayers.Util.urlAppend(service, OpenLayers.Util.getParameterString({ SERVICE: 'WMTS', REQUEST: 'GetCapabilities', VERSION: '1.0.0' }))).then(function (response) {
        if (!response.ok) {
            throw 'WMTS GetCapabilities not loaded: '+response.status+' '+response.statusText
       }
        return response.text()
      });
      const WFSRequest = fetch(OpenLayers.Util.urlAppend(service, OpenLayers.Util.getParameterString({ SERVICE: 'WFS', REQUEST: 'GetCapabilities', VERSION: '1.0.0' }))).then(function (response) {
        if (!response.ok) {
            throw 'WFS GetCapabilities not loaded: '+response.status+' '+response.statusText
       }
        return response.text()
      });
```
Docs https://developer.mozilla.org/en-US/docs/Web/API/Response

nboisteault:
@mdouchin Could you take a look at this PR please? :)

rldhont:
@nboisteault is it possible to test it in cypress by intercepting the requests ?

lizmap/modules/view/locales/en_US/dictionnary.UTF-8.properties
```diff
@@ -1,3 +1,6 @@
+startup.error=An error occurred. Please contact your system administrator.
+startup.goToProject=Go to projects.
```
mdouchin, line 2:
```suggestion
startup.error=An error occurred while loading this map. Some necessary resource may temporarily be unavailable. Please try again later.
startup.goToProject=Go back to the home page.
```


* We must not think the user always knows the administrator nor wants to contact it. I think we should not add this suggestion here
* IMHO We should definitely remove any mention of "project" n LWC and speak about maps.

assets/src/legacy/map.js
```diff
@@ -6421,7 +6421,15 @@ window.lizMap = function() {
         $('#headermenu .navbar-inner .nav a[rel="tooltip"]').tooltip();
         $('#mapmenu .nav a[rel="tooltip"]').tooltip();
         self.events.triggerEvent("uicreated", self);
-
+      })
+      .catch((error) => {
+        console.error(error);
+        const errorMsg = `
+          <p class="error-msg">${lizDict['startup.error']}
+          <br><a href="${lizUrls.basepath}">${lizDict['startup.goToProject']}</a></p>`;
```
mdouchin, line 6429: I would love to have the detail of the error messages displayed in a [<samp>](https://developer.mozilla.org/fr/docs/Web/HTML/Element/samp)  when the current authenticated user is the admin user. This would probably help the admin to understand the issue.

Hopefully it would be simple to add since you throw different messages corresponding to the failed request.

mdouchin (changes requested):
Love to finally get rid of the invisible "service non disponible" :D

PR #3099, opened 2022-06-29
UX - Fix icon not displayed for dataviz panel

Gustry:
Regression #3029

By fixing it, I got one deprecated notice in `lizmap/var/log/errors.log` which must stay empty ...

Funded by 3Liz


Gustry:
Is there a function to parse boolean ?

rldhont:
> Is there a function to parse boolean ?

Not yet but it will be one here https://github.com/3liz/lizmap-web-client/pull/3073

rldhont (approved):

PR #3124, opened 2022-07-19
Update PHP and postgresql containers for tests

laurentj:
The PHP containers is now based on the PHP image of 3liz, and the version of PHP is now configurable with an environment variable set before builds.

The postgresql containers is upgraded to Postgresql 14 / Postgis 3 

Funded by 3Liz


nboisteault (approved):
Cypress is ok so I'm ok :)
Thanks.

PR #3129, opened 2022-07-26
fix: The `editionLayers` of `$configJson` in `Project::getUpdatedConfig` can be `unset`

rldhont:
If there is no edition layers for the user, the `editionLayers` of the `$configJson` is `unset`.
So when lizmap removed layers from config to provide only available configuration for user, the code does not check if the `editionLayers` of `$configJson` is still here.

Funded by 3Liz

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/Project.php
+++ b/lizmap/modules/lizmap/lib/Project/Project.php
@@ -1776,7 +1776,8 @@ class Project
                 unset($configJson->tooltipLayers->{$key});
             }
             // editionLayers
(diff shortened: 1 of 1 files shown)


rldhont:
For backport release_3_4, the fix has to be done here https://github.com/3liz/lizmap-web-client/blob/release_3_4/lizmap/modules/lizmap/classes/lizmapProject.class.php#L2073 but not with the same code.

mdouchin (approved):
Thanks

rldhont:
already fixed in release_3_4

Issue #2972, opened 2022-04-29
Set the minimum Lizmap QGIS server version in the source code


Gustry:
It would be nice to have the minimum Lizmap QGIS server version hardcoded in the source code :

1. check the server information panel is displayed and correct
2. check the Lizmap server version according to the hard coded version
3. add a checkbox in the release process to set the minimum version
4. not enabling the Lizmap landing page and project access if the install is not finished (steps 1 and 2)

Gustry:
When reading the JSON file, please check that the plugin is called `lizmap_server` and not `lizmap`. Maybe raise a warning if `lizmap` is in the JSON ?

PR #3132, opened 2022-07-27
UI - Update to Lizmap server plugin and its version

Gustry:
Linked to #2972 @mdouchin

It would be nice to hardcode as well the QGIS version, so we don't need to translate in Transifex everytime we change the QGIS version (and Lizmap server plugin version as well)

### DIFF
--- a/lizmap/modules/admin/controllers/config.classic.php
+++ b/lizmap/modules/admin/controllers/config.classic.php
@@ -150,6 +150,20 @@ class configCtrl extends jController
                 $form->getControl('allowUserAccountRequests')->setReadOnly();
                 $form->getControl('allowUserAccountRequests')->help = jLocale::get('admin~admin.config.services.allowUserAccountRequest.noemail');
             }
+
+            // These variables are temporary until
+            // https://github.com/3liz/lizmap-web-client/issues/2972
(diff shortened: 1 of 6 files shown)


Gustry:
I didn't find in a XML jform file how to dynamically format a string.

My PHP are quite rusty, trying to manage how to format strings in a template.

mdouchin (approved):
Thanks !

Gustry:
I finally found how to make jform with a dynamic help string.

@laurentj and @mdouchin Where would be the best place to make `$qgisMinimumVersionRequired` and `$lizmapPluginMinimumVersionRequired` global ? (for these two files and later for #2972)

mdouchin:
> I finally found how to make jform with a dynamic help string.
> 
> @laurentj and @mdouchin Where would be the best place to make `$qgisMinimumVersionRequired` and `$lizmapPluginMinimumVersionRequired` global ? (for these two files and later for #2972)

I would say we could add them in the `mainconfig.ini.php`, or better in `lizmapConfig.ini.php` ? Or we could create a new file `lizmapQgisServerVersionTable.ini.php` with these variables to facilitate it update when releasing a new version (I prefer this dedicated file)

Gustry:
Ok, let's try in a next PR, related to #2972 . Merging in the meantime.

PR #3133, opened 2022-07-27
The sender email is required

Gustry:
The field is required but missing the red star

Screenshot from 2022-07-27 21-02-26


laurentj:
This field is not required in some case. So we should not set it as required in the xml form. However, there is probably a regression or an issue. I will investigate.

PR #3139, opened 2022-07-29
Config - Make a single configuration for version using an INI file

Gustry:
Linked to #2972

Need to add a new step when releasing a version to edit this new file with latest version of the Lizmap server plugin version

### DIFF
--- a/lizmap/app/system/mainconfig.ini.php
+++ b/lizmap/app/system/mainconfig.ini.php
@@ -26,6 +26,11 @@ availableLocales="cs_CZ,de_DE,el_GR,en_US,es_ES,eu_ES,fi_FI,fr_FR,gl_ES,hu_HU,it
 ; the locale to fallback when the asked string doesn't exist in the current locale
 fallbackLocale=en_US
 
+[minimumRequiredVersion]
+; The minimum version required about external software to make Lizmap Web Client happy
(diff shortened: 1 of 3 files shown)


laurentj:
Anyway, you should not use `\Jelix\IniFile\IniModifier` to read an ini file, it takes many resources. Just use `$ini = jIniFile::read($filename); $val = $ini['versions']['qgisMinimumVersionRequired']`

Gustry:
Thanks, updated.
I was following @mdouchin comment in https://github.com/3liz/lizmap-web-client/pull/3132#issuecomment-1199376866 about the new file. But indeed, it looks ok like this.

lizmap/app/system/mainconfig.ini.php
```diff
@@ -26,6 +26,10 @@
 ; the locale to fallback when the asked string doesn't exist in the current locale
 fallbackLocale=en_US
 
+[minimumRequiredVersion]
```
mdouchin, line 29: Please add a concise comment above with a `;` to explain the purpose of these new variables

mdouchin (changes requested):
Thanks. Just a small change requested

Gustry:
> The backport to `release_3_5` failed:

Manually backported

Issue #3128, opened 2022-07-22
WMS/WMTS Urls have the wrong server urls. 


dmarteau:
### Is your feature request related to a problem? Please describe

GetCapabilities forwarded from WMS/WMTS urls may be not relevant: capabilities response return the internal urls used by lizmap and this should be reconsidered:

- The server is a backend and may be not exposed (voluntarly) to the end-user so the returned URLS are irrelevants.
- When dockerized with docker-compose, lizmap is not aware of the interface address (ip and port) publicly exposed for the map server, so again, the returned urls are totally irrevelants for the end user

### Describe the solution you'd like
If there is no specific reason to return Capabilities information to the end user, then don't do it:   this i exposing useless/irrelevant/insecure information about backend server 

### Describe alternatives you've considered
Make a proper message indicating that this is only informative data and in may be not relevant or usable as-is. Make it eventually only available for admin for debugging purpose.

### Additional context
See https://github.com/3liz/lizmap-docker-compose/issues/24

Gustry:
Is-it not the new environment variable in QGIS Server settings about service URL ?

dmarteau:
> Is-it not the new environment variable in QGIS Server settings about service URL ?

Yes, you may use either:

Environment variables:
* `QGIS_SERVER_SERVICE_URL`  (native to qgis server)
* `QGIS_SERVER_<SERVICE>_URL` for a per service configuration (native to qgis server)
* [`SERVER_PROXY_URL`](https://docs.3liz.org/py-qgis-server/configuration.html#server-proxy-url) with py-qgis-server but now superseded by variables above.

Headers from proxy
* `X-Qgis-Service-Url` and `X-Qgis-<service>-Url `headers (Qgis server native)
* `X-Forwarded-Url` which is specific to py-qgis-server.

The latter may be used with URL rewriting in proxy.  

Imho, headers would be the best choice for lizmap (no need to configure the server)

rldhont:
@dmarteau thanks, which QGIS Server version supports X-Qgis-Service-Url ?

Gustry:
> @dmarteau thanks, which QGIS Server version supports X-Qgis-Service-Url ?

3.20

rldhont:
The issue still occur when the user is authenticated!

rldhont:
The regexp is not global !

PR #3149, opened 2022-08-03
fix(request): Add X-Qgis-Service-Url header to requests send to QGIS Server

rldhont:
Enhancing lizmap by sending an header with the lizmap ogc service url to QGIS Server.

Partially fix https://github.com/3liz/lizmap-web-client/issues/3128

Funding by 3liz


rldhont:
It is a first step to fix #3128 

Tests have to be added.

rldhont:
Tests will be provided by #3174

Issue #2310, opened 2021-05-28
Editproblems


asdro-f:
Hello,

i can't see the Edit Button in Lizmap...

I created a qgs-project with both shp and spatialite Point layer and put in on my Server, i can see all my Layer (also the spatialite).
I exactly followed the instructions from: https://docs.lizmap.com/current/en/publish/lizmap_plugin/edition.html
But i cant see the edit Button, of course i enabled editing the sql layer in lizmap qgis plugin.

My Log says (when enabled editing from QGIS):
LizMap log
2021-05-28 10:32:23	78.48.179.109	warning	2021-05-28 10:32:23	[2]	file_get_contents(http://127.0.0.1/cgi-bin/qgis_mapserv.fcgi?service=WMS&amp;request=GetCapabilitiesAtlas&amp;map=%2Fhome%2FmyUser%2Fedit%2Fedit%2Eqgs): failed to open stream: HTTP request failed! HTTP/1.1 501 Not Implemented
	/var/www/lizmap-web-client-3.4.0/lizmap/modules/lizmap/classes/lizmapProxy.class.php	249

My Log says (after disable editing from QGIS, spatialite layer still visible):
2021-05-28 10:33:38	78.48.179.109	notice	Spatialite is not available
2021-05-28 10:33:38	78.48.179.109	warning	2021-05-28 10:33:38	[2]	file_get_contents(http://127.0.0.1/cgi-bin/qgis_mapserv.fcgi?service=WMS&amp;request=GetCapabilitiesAtlas&amp;map=%2Fhome%2FmyUser%2Fedit%2Fedit%2Eqgs): failed to open stream: HTTP request failed! HTTP/1.1 501 Not Implemented
	/var/www/lizmap-web-client-3.4.0/lizmap/modules/lizmap/classes/lizmapProxy.class.php	249

Somewhere i read the information, that i wont work with Spatialite (but its described in the lizmap doc, which i linked above)...
So i need a PostGIS for storing my editable Geodata? Or am i missing a required installation.

Thanks for your Help and Best regards

Gustry:
I think there might need some issues with spatialite. #1404 ?
Can you try with postgis ?

asdro-f:
I will try

Gustry:
We are currently dropping Spatialite support for editing capabilities.

Issue #2058, opened 2021-01-22
Remove code related to QGIS < 3


Gustry:
What a nice ticket ;-)

Because we are now on Lizmap 3.5, we should remove code related to QGIS < 3.

* https://github.com/3liz/lizmap-web-client/wiki/Versions#qgis-server

* Clean dropdown menu in the admin panel
* QGIS Composer (which is now QGIS Layout)
* ... ?

Issue #1404, opened 2019-11-26
Can't edit Spatialite layer


Warkorentin:
Hello, 
I followed the documentation in order to create an editable SpatiaLite layer. I did everything from adding an auto-increment column, editing the layer properties and enabling the layer in the projet properties tab, to adding the layer in the lizmap configuration plugin. 
At this point, the layer is showing up on the web client but not the editing tab. I have no clue as to what I should do next. I would like to avoid using postGis if possible.

I'm using qgis v.2.18, lizmap web client in a docker (v.3.1.11) and qgis server v.2.18ltr
Do you have any idea on what I could be doing wrong ?
Thanks in advance

capu81:
I've the same problem!

pcav:
In my understanding the issue is still valid, and SL layer cannot be edited.

PR #3151, opened 2022-08-04
Remove code about spatialite and edition of layer, not supported for a long time in the plugin

Gustry:
In the Lizmap plugin, Spatialite layers for edition capabilities has been disabled for a long time.

So I think this code can be removed ?
Or am I missing something ?

If correct, this can be backported to LWC 3.5.

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/Project.php
+++ b/lizmap/modules/lizmap/lib/Project/Project.php
@@ -1558,18 +1558,7 @@ class Project
             return;
         }
 
-        $editionLayers = $this->cfg->getEditionLayers();
-
-        // Check ability to load spatialite extension
-        // And remove ONLY spatialite layers if no extension found
-        $spatialiteExt = '';
-        if (class_exists('SQLite3')) {
-            $spatialiteExt = $this->getSpatialiteExtension();
-        }
-        if (!$spatialiteExt) {
-            $this->appContext->logMessage('Spatialite is not available', 'error');
-            $xml->readEditionLayers($editionLayers);
-        }
+        $xml->readEditionLayers($this->cfg->getEditionLayers());
     }
 
     protected function readAttributeLayers(QgisProject $xml, ProjectConfig $cfg)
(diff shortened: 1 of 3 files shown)


mdouchin:
Correct !

Gustry:
No more "Spatialite is not available" in the logs ? :)

mdouchin:
No more, but I would like to have @rldhont opinion on this :D

nicogodet:
Oh yeah ! Thanks @Gustry for raising this issue from my IRC messages !

Gustry:
@mdouchin and @rldhont Please double check the PR as I'm not sure about the previous code (or the existing one now)

mdouchin:
Indeed, the previous code seems awkward : the edition Layers were set only when Spatialite was NOT available ??? It seems this PR would remove the annoying "Spatialite not available" but also fix a unknown bug. We should double-check the method `$xml->readEditionLayers`

rldhont:
I am -1 for backporting.

Gustry:
Backporting to 3.6 has been removed. I'm totally fine to keep that for 3.7.

Any idea if the code is correct ?

lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -1502,18 +1502,7 @@ protected function readEditionLayers(QgisProject $xml)
             return;
         }
 
-        $editionLayers = $this->cfg->getEditionLayers();
-
-        // Check ability to load spatialite extension
-        // And remove ONLY spatialite layers if no extension found
-        $spatialiteExt = '';
-        if (class_exists('SQLite3')) {
-            $spatialiteExt = $this->getSpatialiteExtension();
```
mdouchin, line 1511: We should probably get rid of this method too (and search for "spatialite" or "sqlite" in the code too)

mdouchin (changes requested):

lizmap/modules/lizmap/lib/Project/Project.php
Gustry: sqlite can still be used for database backend for Lizmap users and groups.

* https://github.com/3liz/lizmap-web-client/search?q=spatialite
* https://github.com/3liz/lizmap-web-client/search?q=sqlite

If you want further cleaning, I guess a note should be added in separate PR, related to https://github.com/3liz/lizmap-web-client/issues/2058 ?

laurentj:
@rldhont we should backport this patch into lizmap 3.6. It produces tons of logs, that prevent us to audit issues.

Gustry:
No, version 3.6.0 from september 2021.

https://github.com/3liz/lizmap-plugin/blob/master/CHANGELOG.md

> 3.6.3 - 2021-10-07
    Follow up from version 3.6.0 about not supporting spatialite for editing capabilities, the plugin will now remove these lines from the CFG file.


> 3.6.0 - 2021-09-23
 Removing Spatialite from available provider for editing capabilities, only PostgreSQL is supported (fix #361)


Version 3.6.3 was quite strict, removing lines it seems, whatever the version of LWC version. We didn't get any feedbacks in the meantime for this important change.

So indeed, there would be a regression from a feature point of view for an old CFG file made with version 3.5.X from the QGIS plugin and sleeping on a server, which can still be read with Lizmap Web Client 3.6 (and even LWC 3.7)

In September 2021, LWC 3.5 was not out yet : https://github.com/3liz/lizmap-web-client/wiki/Versions#lizmap-web-client

Gustry:
Some people were reporting that edition capabilities with spatialite was not working #1404 #2310 in LWC at that time.

If I remember correctly, that's why we decided to not really fix in the current LWC version (from this time of the ticket, 2019, 2020, 2021), and we decided to drop from the plugin first.

rldhont:
@Gustry you are correct this can be backported to 3.5 because, Lizmap Web Client removes from editionLayers layers with spatialite provider.

rldhont (approved):

rldhont:
@mdouchin we need your review!

mdouchin (approved):
🍾

PR #3159, opened 2022-08-09
UX - Display the URL to use in the Lizmap QGIS plugin

Gustry:
Users are confused which URL to use in the QGIS desktop plugin :

* QGIS server URL from the admin panel ?
* QGIS Server Lizmap API URL from the admin panel ?
* Issue with HTTP redirection ...  https://demo.lizmap.com/lizmap/ versus https://demo.lizmap.com
* Clicking on the small "House" icon to bring to the landing page makes a redirection to "https://demo.lizmap.com/lizmap/index.php/view/" which is also not correct

The correct one is https://demo.lizmap.com/lizmap/ for instance.

So let's display the one to use.
Is `jServer::getServerURI()` the correct one ?

@nboisteault It's not in my JS skills to know how to fetch the icon in the assets and where to put JS. I would need to your help to finish this PR.


Gustry:
@nboisteault Can you check the latest commit ?
I haven't been able to try this button, maybe because of security reason (HTTPS). Did you make it run with `lizmap.local` ?
I also removed the border etc like in other websites : 

Screenshot from 2022-09-16 18-45-46

@laurentj Can you confirm it's `jServer::getServerURI()` we need to use to get server URL ? CF PR description.

nboisteault (approved):

Gustry:
The button is still not active on https://demo.snap.lizmap.com/lizmap_3_6/admin.php/admin/server_information for me. @nboisteault Can you try on the snap website ?

I would need to fix the URL. It's missing one part about the instance name.

nboisteault:
@Gustry https://github.com/3liz/lizmap-web-client/pull/3277

Issue #3154, opened 2022-08-09
[Bug]: Editing data of Array type with PHP 8.1 doesn't (always) work


geographit:
### What is the bug?

When a project has an editable layer with

1.  An array field in Postgis
2. A "customised" Attribute forms (E.g aliases...) 

Trying to edit the layer in Lizmap throws an error in the logs 
`array_key_exists(): Argument #2 ($array) must be of type array`

It is not possible to edit the layer and Lizmap becomes unstable (I noticed some crashes).

### Steps to reproduce the issue

1. Create a postgis table with an array
```
CREATE TABLE test_array (
	id SERIAL NOT NULL PRIMARY KEY,
	my_array real [5] DEFAULT NULL,
 	geom geometry(MULTIPOLYGON, 4326)
	);
```

2. Create a lizmap project with the layer set as editable
3. Style the layer attribute form (add an alias). This will cause the .qgs file to specify the field type in xml  `<editWidget type="List">`

4. In lizmap edit the layer. It should hang and display an error in Lizmap logs.

### Temporary fix

Installing php 7.8 instead of php 8.1 solves the issue.
Also if the layer's attribute form is not modified, the project file (*.qgs) it seems php won't know the data type and there will be no error. 

### Versions

QGIS desktop 3.26 
Lizmap 3.55 
QGIS Server 3.26

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 22 lts

### Browsers

Firefox, Chrome, Safari, Microsoft Edge

### Browsers version

Firefox 89

nworr:
thanks for the detailled steps, 
it's related to a change in [array_key_exists](https://www.php.net/manual/en/function.array-key-exists.php#refsect1-function.array-key-exists-notes) since php 8.0.
quick fix can be made

PR #3161, opened 2022-08-10
quick fix using is_array

nworr:
no more php error in php8
 

Ticket : #3154

### DIFF
--- a/lizmap/modules/lizmap/lib/Form/QgisFormControl.php
+++ b/lizmap/modules/lizmap/lib/Form/QgisFormControl.php
@@ -411,7 +411,7 @@ class QgisFormControl
 
         // Read-only
         if ($this->fieldDataType != 'geometry') {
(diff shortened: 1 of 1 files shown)


mdouchin (approved):
Thanks a lot. Ok for this fix, but we should in the future better understand why this `self::$qgisEdittypeMap` could return something else ;)

laurentj (changes requested):
There is the issue because `$this->fieldEditType` is `List` and we don't support `List` into `$qgisEdittypeMap`. So the test should check first that `self::$qgisEdittypeMap[$this->fieldEditType]` exists before checking the `readonly` parameter.

laurentj (approved):
r+

nworr:
question before merging, 
- are the (backport release) labels correct ?
- need to add a new feature to report unhandled fieldEditType ?

PR #3162, opened 2022-08-10
Form filter - Allow to use a second field for the numeric type

mdouchin:
It was already possible for dates in the previous versions.
Now a second field can also be used for the numeric form widget.
It is useful when the layer features contain two fields describing
a minimum and maximum value of the same property (ex: age, width, depth, etc.)

Linked to https://github.com/3liz/lizmap-plugin/pull/442

### DIFF
--- a/assets/src/legacy/filter.js
+++ b/assets/src/legacy/filter.js
@@ -216,13 +216,22 @@ var lizLayerFilterTool = function () {
 
             // Get the HTML form element for the date field type
             function dateFormInput(field_item) {
-                // max_date = min_date when undefined
-                const max_date = ('max_date' in field_item) ? field_item.max_date : field_item.min_date;
+                // For LWC >=3.7, the min_date and max_date
+                // are replaced by start_field & end_field
+                var start_field_property = 'min_date';
+                var end_field_property = 'max_date';
+                let lizmap_plugin_metadata = lizMap.getLizmapDesktopPluginMetadata();
+                if (lizmap_plugin_metadata.lizmap_web_client_target_version >= 30700) {
+                    start_field_property = 'start_field';
+                    end_field_property = 'end_field';
+                }
+                // end_field = start_field when undefined
+                const end_field = (end_field_property in field_item) ? field_item[end_field_property] : field_item[start_field_property];
 
                 var sdata = {
                     request: 'getMinAndMaxValues',
                     layerId: field_item.layerId,
-                    fieldname: field_item.min_date + ',' + max_date,
+                    fieldname: field_item[start_field_property] + ',' + end_field,
                     filter: ''
                 };
                 $.get(filterConfigData.url, sdata, function (result) {
@@ -278,10 +287,22 @@ var lizLayerFilterTool = function () {
 
             // Get the HTML form element for the numeric field type
             function numericFormInput(field_item) {
+                // Since LWC >=3.7, we can configure a field with the end value
+                var fieldNames = '';
+                let lizmap_plugin_metadata = lizMap.getLizmapDesktopPluginMetadata();
+                if (lizmap_plugin_metadata.lizmap_web_client_target_version >= 30700) {
+                    fieldNames += field_item.start_field;
+                    if ('end_field' in field_item && field_item.end_field != '') {
+                        fieldNames += ',' + field_item.end_field;
+                    }
+                } else {
+                    fieldNames += field_item.field;
+                }
+
                 var sdata = {
                     request: 'getMinAndMaxValues',
                     layerId: field_item.layerId,
-                    fieldname: field_item['field'],
+                    fieldname: fieldNames,
                     filter: ''
                 };
                 $.get(filterConfigData.url, sdata, function (result) {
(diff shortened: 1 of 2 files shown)


Gustry:
Ouch in cypress

> Undefined array key "metadata"\n  /srv/lzm/lizmap/modules/lizmap/lib/Server/Server.php\t56

It seems from my fault ...

assets/src/legacy/filter.js
```diff
@@ -278,10 +278,22 @@ var lizLayerFilterTool = function () {
 
             // Get the HTML form element for the numeric field type
             function numericFormInput(field_item) {
```
Gustry, line 280: On IRC, I understood you would take care of `min_date` => `start_field` and `max_date` => `end_field`. These keys might be merged soon on my part.

It will avoid this screenshot : 
ksnip_20220810-164839

Gustry:
> Ouch in cypress
> 
> > Undefined array key "metadata"\n  /srv/lzm/lizmap/modules/lizmap/lib/Server/Server.php\t56
> 
> It seems from my fault ...

https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/lib/Server/Server.php#L56

The metadata is empty when it's called.

mdouchin:
> > Ouch in cypress
> > > Undefined array key "metadata"\n  /srv/lzm/lizmap/modules/lizmap/lib/Server/Server.php\t56
> > 
> > 
> > It seems from my fault ...
> 
> https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/lib/Server/Server.php#L56
> 
> The metadata is empty when it's called.

I will check

mdouchin:
It seems we must change a little in 3.6 (then use it here) how the full metadata is build and shared by the `Server.php` lib.

At present, if you are not logged as an admin, the returned metadata is 
```json
{"info":{"version":"3.7.0-pre","date":"2021-07-27"},"dependencies":{"jelix":{"version":"1.8.0-alpha.2","minversion":"1.8.0-alpha.2","maxversion":"1.8.0-alpha.2"}},"qgis_server":{"test":"OK","mime_type":"text\/xml; charset=utf-8"},"qgis_server_info":{"error":"NO_ACCESS"}}
```
There is a rule here: https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/lib/Server/Server.php#L157
which explains this response.
I think the class `Server` should not consider if the user is logged or not but always return the full metadata. The controller(s) should do the check:

* in the admin project page, no worries, no check
* for the "map page", in the `lizMap.classic.php` the metadata object is only used for access control (I think but have not checked). If so, no worries, no check
* for the view/app/metadata (the source of the JSON downloaded by the Lizmap plugin) -> the controller should also check and remove the sensitive data if the user is not an administrator. Before returning the metadata, remove sensitive properties: https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/view/controllers/app.classic.php#L32

assets/src/legacy/filter.js
```diff
@@ -278,10 +278,22 @@ var lizLayerFilterTool = function () {
...
+                // Since LWC >=3.6, we can configure a field with the end value
```
Gustry, line 281:
```suggestion
                // Since LWC >=3.7, we can configure a field with the end value
```

assets/src/legacy/filter.js
```diff
@@ -278,10 +278,22 @@ var lizLayerFilterTool = function () {
...
+                var fieldNames = '';
+                let lizmap_plugin_metadata = lizMap.getLizmapDesktopPluginMetadata();
+                if (lizmap_plugin_metadata.lizmap_web_client_target_version >= 30600) {
```
Gustry, line 284:
```suggestion
                if (lizmap_plugin_metadata.lizmap_web_client_target_version >= 30700) {
```

assets/src/legacy/filter.js
```diff
@@ -649,19 +661,41 @@ var lizLayerFilterTool = function () {
                     return true;
                 }
 
-                // field
-                var field = field_item['field'];
+                // fields
+                var startField = '';
+                var endField = null;
+
+                // Since LWC >=3.6, we can configure a field with the end value
```
Gustry, line 668:
```suggestion
                // Since LWC >=3.7, we can configure a field with the end value
```

assets/src/legacy/filter.js
```diff
@@ -649,19 +661,41 @@ var lizLayerFilterTool = function () {
...
+                let lizmap_plugin_metadata = lizMap.getLizmapDesktopPluginMetadata();
+                if (lizmap_plugin_metadata.lizmap_web_client_target_version >= 30600) {
```
Gustry, line 670:
```suggestion
                if (lizmap_plugin_metadata.lizmap_web_client_target_version >= 30700) {
```

Gustry (commented):
We should choose which one ;-)

Gustry (approved):
Not a JS expert, I would say LGTM !

I would be awesome to have a JS unit test to check the SQL generation according to a few config given as JSON. But this would wait for #3073

Gustry (approved):
Double LGTM with the funder

mdouchin:
@rldhont gentle ping

rldhont:
@mdouchin LGTM but a test could be added.

mdouchin:
If it is ok for you, I propose we add JS test after your PR #3073 with a test infrastructure will be merged

Issue #3171, opened 2022-08-21
[Bug]: Wrong group ID in the CFG leads to a PHP error when loading the project


Gustry:
### What is the bug?

The project is not available in the group ID doesn't not exist : 

> An error occurred while loading this map. Some necessary resource may temporarily be unavailable. Please try again later. 

In the PHP logs : 

```
2022-08-21 14:26:39	172.23.0.1	error	[0]	property_exists(): Argument #1 ($object_or_class) must be of type object|string, array given
  /srv/lzm/lizmap/modules/lizmap/lib/Project/Project.php	1780
 /index.php/lizmap/service/getProjectConfig?repository=testsrepository&project=filter_layer_by_user
 array (   'repository' => 'testsrepository',   'project' => 'filter_layer_by_user',   'module' => 'lizmap',   'action' => 'service:getProjectConfig', )
 ref:http://lizmap.local:8130/index.php/view/map?repository=testsrepository&project=filter_layer_by_user

	0	property_exists()	/srv/lzm/lizmap/modules/lizmap/lib/Project/Project.php : 1780
	1	Lizmap\Project\Project->getUpdatedConfig()	/srv/lzm/lizmap/modules/lizmap/controllers/service.classic.php : 751
	2	serviceCtrl->getProjectConfig()	/srv/lzm/lizmap/vendor/jelix/jelix/lib/jelix/core/jCoordinator.class.php : 266
	3	jCoordinator->process()	/srv/lzm/lizmap/www/index.php : 23


2022-08-21 14:27:03	172.23.0.1	deprecated	[8192]	trim(): Passing null to parameter #1 ($string) of type string is deprecated
  /srv/lzm/lizmap/vendor/jelix/jelix/lib/jelix-admin-modules/jauthdb_admin/controllers/default.classic.php	132
 /admin.php/users-admin
 array (   'module' => 'jauthdb_admin', )
 ref:http://lizmap.local:8130/admin.php/auth/account/show?user=admin

	0	trim()	/srv/lzm/lizmap/vendor/jelix/jelix/lib/jelix-admin-modules/jauthdb_admin/controllers/default.classic.php : 132
	1	defaultCtrl->index()	/srv/lzm/lizmap/vendor/jelix/jelix/lib/jelix/core/jCoordinator.class.php : 266
	2	jCoordinator->process()	/srv/lzm/lizmap/www/admin.php : 23


[...]

PR #3181, opened 2022-08-26
fix Unknown group id leads to php error while loading project

nworr:
quick & dirt fix to prevent php error (force object cast)
must be related to a php8 behavior change in property_exists

fix https://github.com/3liz/lizmap-web-client/issues/3171

not sure of labels ...


laurentj:
I'm not sure this is enough. I would prefer a better patch, which should respond to the question : why is `$this->editionLayersForCurrentUser` not an object.

rldhont:
`$this->editionLayersForCurrentUser` is an array but `$configJson` is managed as an `Object` for example https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/lib/Project/Project.php#L1778

rldhont (approved):

laurentj (approved):

rldhont:
Manually backport to master

PR #3203, opened 2022-09-02
[Backport release_3_5] fix(admin): Use normalizedPath instead of realpath to check rootRepositories

rldhont:
Backport #3202

The PHP method `realpath` returns canonicalized absolute pathname. realpath() expands
all symbolic links and resolves references to /./, /../ and extra / characters in the
input path and returns the canonicalized absolute pathname.

The `\Jelix\FileUtilities\Path::normalizePath` resolves references to /./, /../ and
extra / characters in the input path.

In Lizmap, the rootRepositories config parameter could be a symbolic link, in which find
Lizmap repositorires. So Lizmap does not need to expand symbolic link in the admin maps
controller, it only needs to normalize path.

Funded by Ingerop

### DIFF
--- a/lizmap/modules/admin/controllers/maps.classic.php
+++ b/lizmap/modules/admin/controllers/maps.classic.php
@@ -9,6 +9,8 @@
  *
  * @license Mozilla Public License : http://www.mozilla.org/MPL/
  */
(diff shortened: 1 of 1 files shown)


laurentj (approved):

rldhont:
Thansk @laurentj

PR #3211, opened 2022-09-08
Add color picker to theme edition in admin

nboisteault:
It would be convenient to have color pickers instead of text input where people don't always know what to put in.
It is just a quick draft to discuss.
By default color pickers have their default value to `#000000` (black) so if someone saves the form without changing any colors this color will apply to all fields. So we need to display default values when opening the form.

We could use CSS custom properties. It would also allow to use them in custom themes defined in `media`.

Funded by 3Liz

image

### DIFF
--- a/lizmap/modules/admin/locales/en_US/admin.UTF-8.properties
+++ b/lizmap/modules/admin/locales/en_US/admin.UTF-8.properties
@@ -159,6 +159,7 @@ form.admin_section.message.accessControlAllowOrigin.bad.domain=Content is not a 
 menu.lizmap.theme.label=Theme
 configuration.button.modify.theme.label=Modify
 form.admin_theme.h1=Theme
+form.admin_theme.helpcolor=You can select colors with this <a href="https://mdn.github.io/css-examples/tools/color-picker/">colorpicker</a> (e.g. <code>rgb(148 193 31 / 0.5)</code>).
 form.admin_theme.headerLogo.label=Header - logo
 form.admin_theme.headerLogo.help=The image to be used for the header logo (200ko Maximum).
 form.admin_theme.headerLogoWidth.label=Header - logo width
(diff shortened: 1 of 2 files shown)


nboisteault:
@laurentj After more thinking and our discussion, I think there are several issues using the native color picker:

1. There is no way to set the transparency. Some people could want some transparency to see the map behind the map menu for example
2. There is no way to use 'red' or 'lightgray' as color for example
3. There is no way to go back to default value. This would  even be a bug for `headerBackgroundColor` as it overrides the default background image.

I suggest the easy solution to link to an online color picker. This one from Mozilla seems a good candidate : https://mdn.github.io/css-examples/tools/color-picker/

cc @rldhont @Gustry @mdouchin

Gustry:
> I suggest the easy solution to link to an online color picker. This one from Mozilla seems a good candidate : https://mdn.github.io/css-examples/tools/color-picker/

Ok sad..., but better than nothing, especially for people not knowing HTML.
Ok for it

nboisteault:
@laurentj it is ok for the template. What about placeholders?

nboisteault:
image

lizmap/modules/admin/locales/en_US/admin.UTF-8.properties
```diff
@@ -159,6 +159,7 @@ form.admin_section.message.accessControlAllowOrigin.bad.domain=Content is not a
 menu.lizmap.theme.label=Theme
 configuration.button.modify.theme.label=Modify
 form.admin_theme.h1=Theme
+form.admin_theme.helpcolor=You can select colors with this <a href="https://mdn.github.io/css-examples/tools/color-picker/">colorpicker</a> (e.g. <code>rgb(148 193 31 / 0.5)</code>).
```
nboisteault, line 162: @laurentj Is it ok to put HTML in strings? If not, how to handle that?

Gustry, line 162:
```suggestion
form.admin_theme.helpcolor=You can select colors with this <a href="https://mdn.github.io/css-examples/tools/color-picker/" target="_blank">colorpicker</a> (e.g. <code>rgb(148 193 31 / 0.5)</code>).
```

otherwise, it's not convenient to leave LWC ...

Gustry (commented):
Looks good, thanks for this help.

lizmap/modules/admin/locales/en_US/admin.UTF-8.properties
Gustry: It's already done quite few times in LWC.
Transifex manage well HTML in strings, it's a click and click translation process on HTML tags.

PR #3250, opened 2022-09-26
Editing - Move the child tables to the corresponding relation widget container

mdouchin:
Before this change, Lizmap Web Client displays the children related tables at the bottom of the form,
under the action buttons.
Now, the child tables will be moved inside the form if a relation widget has been inserted
with QGIS drag and drop designer.

The label of the relation widget is used and displayed above the table. If the label is empy,
the name of the relation is used instead.

Funded by Valabre https://www.valabre.com/

image


mdouchin:
@rldhont @nboisteault ready for review

nboisteault:
@mdouchin We are polishing 3.6 so I'd prefer to have this PR in 3.7. All cypress tests must be green.

mdouchin:
Je peux merger svp ?

mdouchin:
I would like to merge it :D Further improvements on editing will be done in the future

Gustry (approved):
Approved from my user point of view and because we talked about this PR and the future improvements coming.

PR #3254, opened 2022-09-28
Editing - Fix: the HTML form widget must use WYSIWYG editor

mdouchin:
LWC editing form must show a WYSIWYG editor when the map editor has configured
a text field widget with the option "Use HTML"

Funded by Parc naturel régional du Pilat https://www.parc-naturel-pilat.fr/

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -947,21 +947,6 @@ parameters:
 			path: lizmap/modules/view/classes/lizmapMapDockItem.class.php
 
 		-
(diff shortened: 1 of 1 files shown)


mdouchin:
cc @rldhont @Gustry

mdouchin:
@laurentj I created this PR to make this bug reproducible. Could you have a look and see why the HTML widget is not used (and instead a simple online text widget) ?

Gustry:
Cypress is ✅

PR #3245, opened 2022-09-23
[Fix] Feature is editable

rldhont:
Funded by 3Liz

PR #3256, opened 2022-09-29
fix(request): OGC GetCapabilities with tests

### LINKED ISSUE #3128: WMS/WMTS Urls have the wrong server urls.

rldhont:
Fixes #3174
Fixes #3128
Funded by Andromede Oceanologie https://www.andromede-ocean.com

### DIFF
--- a/lizmap/modules/lizmap/lib/Request/WFSRequest.php
+++ b/lizmap/modules/lizmap/lib/Request/WFSRequest.php
@@ -213,7 +213,7 @@ class WFSRequest extends OGCRequest
             array('repository' => $this->repository->getKey(), 'project' => $this->project->getKey())
         );
         $sUrl = str_replace('&', '&amp;', $sUrl).'&amp;';
-        preg_match('/<get>.*\n*.+xlink\:href="([^"]+)"/i', $data, $matches);
+        preg_match('/get[\W\s]*onlineresource.+xlink\:href="([^"]+)"/i', $data, $matches);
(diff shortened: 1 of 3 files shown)


nboisteault:
@rldhont Cypress NOK

rldhont:
I have found an issue in QGIS Server and provide a fix in lizmap server https://github.com/3liz/qgis-lizmap-server-plugin/pull/20

rldhont:
@nboisteault the failed test is about filter by polygon and has to failed since https://github.com/3liz/qgis-lizmap-server-plugin/pull/20 is not available and https://github.com/3liz/lizmap-web-client/pull/3245 is not been merged

PR #3268, opened 2022-10-03
Fix: a missing project broke the configuration page

laurentj:
When we display the configuration page, we retrieve the list of projects to display the label of the default project. But if this default project does not exists any more, the page turned into a 500 error page.

This is because, in a try/catch block, we catch the deprecated exception UnknownLizmapProjectException instead of the \Lizmap\Project\UnknownLizmapProjectException.

Many other try/catch have the same issue.

(UnknownLizmapProjectException is still there to not break external modules, but is not used anymore by our code).

Funded by 3liz

### DIFF
--- a/lizmap/modules/action/classes/actionConfig.class.php
+++ b/lizmap/modules/action/classes/actionConfig.class.php
@@ -29,7 +29,7 @@ class actionConfig
 
                 return;
             }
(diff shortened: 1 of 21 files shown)


mdouchin (approved):
Merci

PR #3277, opened 2022-10-12
Fix copy to clipboard

nboisteault:


Gustry (approved):
LGTM, I couldn't try...

PR #3291, opened 2022-10-26
Fix: Ip into logs is not the real ip

laurentj:
When Lizmap is behind a reverse proxy (this is often the case when running nginx + php-fpm), REMOTE_ADDR does not contain the user IP but the server IP.

We should use the method `jRequest::getIP()` which returns the real ip of the user.

Funded by 3Liz


mdouchin (approved):
merci

rldhont (approved):

rldhont:
Thnaks @laurentj

PR #3294, opened 2022-10-28
Playwright Github Action

nboisteault:
Funded by 3Liz

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -1,4 +1,4 @@
-name: "🎳 Cypress"
+name: "🎳 End2end"
 on:
   pull_request:
     types: [ labeled, opened, synchronize, reopened ]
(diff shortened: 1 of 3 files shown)


Gustry (approved):
LGTM for a first step.

However : 
1. I would switch cypress and playwright. You said playwright is quicker to run (there is also less tests), so you have the result quicker. And it will encourage us to switch to playwright I guess ? What do you think ?
2. For now, playwright will run **only** if cypress is OK for now. We should have both running in parallel so the first one stack doesn't prevent the second stack running in case of failure. But indeed, I see the issue in caching docker images to avoid pulling twice ... Let me think about it.

So maybe only check point 1 ?
Then, when it's mergerd, we can remove the "run cypress" label. I had to recreate it because you have "borken" the current "run cypress" workflow.

Gustry:
LGTM. Backport 3.5 ?

PR #3304, opened 2022-11-03
GH Action - Upgrade cypress action to 4.2.0

Gustry:
I guess to remove all warnings and deprecation notices on GH Action, CF https://github.com/3liz/lizmap-web-client/actions/runs/3386817243

I'm curious why dependabot didn't detect this action... We were on a old version.

@nboisteault I took the opportunity to look for cypress as well, ok for you ?


Gustry:
I reverted my changes about cypress 9.7.0 => [v10.11.0](https://github.com/cypress-io/cypress/releases/tag/v10.11.0)

PR #3309, opened 2022-11-04
Update jquery-ui to 1.13.2

nboisteault:
Funded by 3Liz


laurentj (approved):
Ok for me if you want Lizmap having its own jQueryUI lib. However I updated all JS libs (jQuery, jQueryUI, Datatables, CKEditor) into jelix and so into lizmap/www/assets/jelix/, so feel free to use them.

PR #3318, opened 2022-11-07
Docker, ini files: possibility to delete a section from target files

laurentj:
Into a drop-in INI file, we can now indicate a section to remove from the target INI file.

Funded by 3liz


dmarteau (approved):
LGTM

Issue #3326, opened 2022-11-14
[Bug]: 


arno974:
### What is the bug?

Activating the table attribute feature for a layer that comes from another QGIS project (embed layers) generates an error 500 with the error :
`[0]	count(): Argument #1 ($value) must be of type Countable|array, null given	/media/arnaud/perso/www/lizmap/lizmap/modules/lizmap/lib/Project/QgisProject.php	1033`

If I remove this layer from the list of available layers i

### Steps to reproduce the issue

1- In QGIS Lizmap Plugin activate attribute table for an embed layer
2- In QGIS Web reload the project -> error 500   

### Versions

QGIS : 3.16.16
Lizmap : 3.5.7

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### QGIS server version, only if the section above doesn't mention the QGIS Server version

3.16.6

### Operating system

Ubuntu 22.04

### Browsers

Firefox

### Browsers version

106.0.2

### Relevant log output

```shell
[0]	count(): Argument #1 ($value) must be of type Countable|array, null given	/media/arnaud/perso/www/lizmap/lizmap/modules/lizmap/lib/Project/QgisProject.php	1033
```

nworr:
is the php version > 8 and in this case can you test with lower version php  (7.4) ?

arno974:
PHP Version PHP 8.1.2-1ubuntu2.8 (cli)
Unfortunately I can't test with a 7.4 php version as it is not available in my packages.

nworr:
can reproduce with https://github.com/3liz/lizmap-web-client/pull/3334
end2end provided test fail (page in error while loading a project with embed layer)

PR #3334, opened 2022-11-17
Embed layer test

nworr:
Add new project with embed layer
and playwright test

Ticket : #3326  

Funded by 3liz

### DIFF
--- a/tests/end2end/playwright/startup.spec.js
+++ b/tests/end2end/playwright/startup.spec.js
@@ -40,4 +40,12 @@ test.describe('Startup', () => {
      
     await expect( page.locator('#layer-quartiers')).toHaveClass(/liz-layer/);
   });
+
+  test('Projects with embed layer can load', async ({page}) => {
+    const proj_url = '/index.php/view/map/?repository=testsrepository&project=embed_layer';
+    await page.goto(proj_url, { waitUntil: 'networkidle' });
+    await expect( page.locator('#map')).toBeVisible();
+
+
+  });
 });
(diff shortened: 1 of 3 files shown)


tests/end2end/cypress/integration/projects_embed_layer-ghaction.js
```diff
@@ -0,0 +1,14 @@
+// projects_filename.js created with Cypress
```
nboisteault, line 1: Please just put a test for playwright.

tests/end2end/playwright/startup.spec.js
```diff
@@ -40,4 +40,13 @@ test.describe('Startup', () => {
      
     await expect( page.locator('#layer-quartiers')).toHaveClass(/liz-layer/);
   });
+
+  test('Projects with embed layer can load', async ({page}) => {
+    const url_dots = '/index.php/view/map/?repository=testsrepository&project=embed_layer';
```
nboisteault, line 45: Why `url_dots` variable name?

tests/end2end/playwright/startup.spec.js
```diff
@@ -40,4 +40,13 @@ test.describe('Startup', () => {
...
+    await page.goto(url_dots, { waitUntil: 'networkidle' });
+    await page.$eval("#map, #map *", el => el.style.visibility = 'visible');
```
nboisteault, line 47: Why hiding other elements?

nboisteault (changes requested):

nboisteault:
@nworr should it be reopened?

nworr:
IMHO, the project file is way too simple to ensure embed layers actually works,  there may problematic use case that need to be checked ( user-filtered layers, mutiple projections, ...).

PR #3365, opened 2022-12-09
Bump min versions for LWC 3.7 : QGIS server and CFG files

Gustry:
Raise

* QGIS min version to 3.16
* CFG files to 3.4 (lower versions won't be displayed in the landing page)
* Latest Lizmap server plugin

@mdouchin @rldhont @nboisteault I guess it's more than OK.

I need to popup the docker stack and do updates if necessary in CFG files


nboisteault:
For Lizmap 3.7, I would make QGIS 3.22 the only allowed version.

Gustry:
Actually, it depends when we release 3.7 if it's just in few months or not :)

Let's merge like this and we will check later when we will be around the release.

Side note, it's not very nice for end users when they just update to see their project hidden if they have to update dozens/hundreds of projects. I was thinking about a flag in the administration interface to temporary disable this check.

Issue #2044, opened 2021-01-13
Add the version info to the Lizmap user-agent


dmarteau:
When sending request to Qgis server Lizmap set the user agent to 'Lizmap'. 

It would be better (for monitoring/debugging) that the user agent include the version number. A Qgis  server plugin could react to this  to support multiple Lizmap versions if necessary.

PR #3373, opened 2022-12-16
add lizmap version to proxy request

nworr:
Proxy Request currently use "Lizmap" as user-agent in request header, that can be usefull for OGC service

the PR add the current lizmap Version

Fix : #2044

Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmapServices.class.php
+++ b/lizmap/modules/lizmap/classes/lizmapServices.class.php
@@ -281,14 +281,15 @@ class lizmapServices
             $this->allowUserAccountRequests = false;
         }
 
-        // set it for external requests, needed for file_get_contents
-        $userAgent = 'Lizmap';
+        $appInfos = \Jelix\Core\Infos\AppInfos::load();
+        // set user_agent for external requests, needed for file_get_contents
+        $userAgent = 'Lizmap '.$appInfos->version;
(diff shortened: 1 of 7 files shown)


rldhont:
@nworr I think you have to rebase your branch to avoid conflict with phptan tests fixes.

lizmap/modules/lizmap/classes/lizmapServices.class.php
```diff
@@ -282,14 +282,15 @@ public function __construct($readConfigPath, $globalConfig, $ldapEnabled, $varPa
             $this->allowUserAccountRequests = false;
         }
 
-        // set it for external requests, needed for file_get_contents
-        $userAgent = 'Lizmap';
+        $appInfos = \Jelix\Core\Infos\AppInfos::load();
```
laurentj, line 285: Retrieve the userAgent in an `else` statement of the statement ` if (isset($readConfigPath['services']['userAgent'])) {`. 
`AppInfos` loads xml file, and so it takes some resources. We could avoid to load this file if we don't need it.

laurentj (changes requested):

rldhont:
@laurentj requested  changes done.

rldhont (approved):

laurentj (changes requested):

rldhont:
@nworr it is ready for merging?

PR #3369, opened 2022-12-13
JSON Lizmap Metadata - Add groups and repositories

mdouchin:
This addition will allow the Lizmap desktop plugin to retrieve the needed data to help the publisher

* the list of groups of users
* the list of Lizmap repositories with 
  * the groups allowed to view the projects and to edit the layers
  * the list of projects

Funded by Direction Départementale de la Terre et de la Mer de l'Hérault (DDTM 34): https://www.herault.gouv.fr/Services-de-l-Etat/Agriculture-environnement-amenagement-et-logement/Direction-Departementale-des-Territoires-et-de-la-Mer-de-l-Herault-DDTM-34

PR #3376, opened 2022-12-19
Dataviz - Allow to post a JSON plot config to get the plot

mdouchin:
When you configure a plot in QGIS Lizmap plugin, you cannot yet test how the plot will be rendered. 
This modification allows any client, like a Python script, to use the Lizmap Web Client `dataviz` service as an API.

This API is only available for authenticated users. You can at present only use **Basic authentication**.

The client is able to request Lizmap Web Client and get the plot data ready to be directly used with the Plotly JavaScript or Python library.

If LWC main URL is `http://lizmap.local:8130/` The **URL** of the dataviz API is : `http://lizmap.local:8130/index.php/dataviz/service?`

The preferred request method is **POST**, and the data should be sent in the body as a JSON object, 
with the keys `repository` (Lizmap repository code), `project` (the Lizmap project code) 
and the plot configuration in `plot_config`. Here is an example of JSON file `plot_config.json` :

```json
{
    "repository": "testsrepository",
    "project": "dataviz",
    "plot_config": {
        "type": "bar",
        "title": "Communes",
        "layerId": "polygons_1e227060_6105_4a4a_92f2_11863ebe65f1",
        "x_field": "name",
        "aggregation": "sum",
        "traces": [
            {
                "color": "cyan",
                "colorfield": "",
                "y_field": "id",
                "z_field": ""
            }
        ],
        "popup_display_child_plot": "False",
        "stacked": "False",
        "horizontal": "False",
        "only_show_child": "False",
        "display_legend": "True",
        "display_when_layer_visible": "False",
        "order": 0
    }
}
```

You can use `curl`, Python request or any other request library to send the data and get the plot as the response.
For example with curl

```bash
curl --verbose -X POST \
     -H 'Content-Type: application/json' \
     --user admin:admin \
     -d @plot_config.json \
     "http://lizmap.local:8130/index.php/dataviz/service?"
```

The API returns a JSON response, which can be used in a Web viewer with the necessary HTML/JavaScript code using the Plotly library:

```json
{
  "title": "Communes",
  "data": [
    {
      "type": "bar",
      "name": "id",
      "y": [
        0,
        1,
        2,
        3,
        4,
        5,
        6,
        7,
        8,
        9
      ],
      "x": [
        "Grabels",
        "Clapiers",
        "Montferrier-sur-Lez",
        "Saint-Jean-de-Védas",
        "Lattes",
        "Montpellier",
        "Lavérune",
        "Juvignac",
        "Le Crès",
        "Castelnau-le-Lez"
      ],
      "ids": null,
      "text": [],
      "marker": {
        "color": "cyan",
        "colorscale": null,
        "showscale": false,
        "reversescale": false,
        "colorbar": {
          "len": "0.8"
        },
        "line": {
          "color": null,
          "width": null
        }
      },
      "textinfo": "none",
      "orientation": "v",
      "transforms": [
        {
          "type": "aggregate",
          "groups": "x",
          "aggregations": [
            {
              "target": "y",
              "func": "sum",
              "enabled": true
            }
          ]
        }
      ]
    }
  ],
  "layout": {
    "showlegend": true,
    "legend": {
      "orientation": "h",
      "x": "-0.1",
      "y": "1.15"
    },
    "autosize": true,
    "plot_bgcolor": "rgba(0,0,0,0)",
    "paper_bgcolor": "rgba(0,0,0,0)",
    "margin": {
      "t": 10,
      "b": 30,
      "l": 30,
      "r": 30,
      "pad": 1
    },
    "xaxis": {
      "tickfont": {
        "size": 10
      },
      "automargin": true
    },
    "yaxis": {
      "tickfont": {
        "size": 10
      },
      "automargin": true
    }
  }
}
```

Here is a very simple Python script illustrating the possible usage 
(you need to have the variable `plot_json` set with the data coming from the API):

```python
# Plot data returned by the API
plot_json = {}

# plot configuration
plot_config = {
    "showLink": False,
    "scrollZoom": False,
    "locale": "en",
    "responsive": True,
}

# Compute the HTML content
html = f"""
    <head>
      <!-- Load plotly.js into the DOM -->
      <script src='https://cdn.plot.ly/plotly-2.16.1.min.js'></script>
    <script type="text/javascript">
        function plotLizmap() {{
            Plotly.newPlot(
                "lizmap-plot",
                {json.dumps(plot_json['data'])},
                {json.dumps(plot_json['layout'])},
                {json.dumps(plot_config)}
            );
        }}
        window.addEventListener('load', function () {{
          plotLizmap();
        }});
      </script>
    </head>

    <body>
      <div id='lizmap-plot' style="width:100%; height: 100%;">
        <!-- Plotly chart will be drawn inside this DIV -->
      </div>
    </body>
"""

```

### DIFF
--- a/lizmap/modules/dataviz/classes/datavizConfig.class.php
+++ b/lizmap/modules/dataviz/classes/datavizConfig.class.php
@@ -18,40 +18,48 @@ class datavizConfig
     public function __construct($repository, $project)
     {
         try {
-            $lproj = lizmap::getProject($repository.'~'.$project);
-            if (!$lproj) {
+            $lizmapProject = lizmap::getProject($repository.'~'.$project);
+            if (!$lizmapProject) {
                 $this->errors = array(
-                    'title' => 'Invalid Query Parameter',
-                    'detail' => 'The lizmap project '.strtoupper($project).' does not exist !',
+                    'code' => 404,
+                    'error_code' => 'project_not_found',
+                    'title' => jLocale::get('dataviz~dataviz.log.project_not_found.title'),
+                    'detail' => jLocale::get('dataviz~dataviz.log.project_not_found.detail', array($project)),
                 );
 
                 return;
             }
         } catch (\Lizmap\Project\UnknownLizmapProjectException $e) {
             $this->errors = array(
-                'title' => 'Invalid Query Parameter',
-                'detail' => 'The lizmap project '.strtoupper($project).' does not exist !',
+                'code' => 404,
+                'error_code' => 'project_not_found',
+                'title' => jLocale::get('dataviz~dataviz.log.project_not_found.title'),
+                'detail' => jLocale::get('dataviz~dataviz.log.project_not_found.detail', array($project)),
             );
 
             return;
         }
 
         // Check acl
-        if (!$lproj->checkAcl()) {
+        if (!$lizmapProject->checkAcl()) {
             $this->errors = array(
-                'title' => 'Access Denied',
-                'detail' => jLocale::get('view~default.repository.access.denied'),
+                'code' => 403,
+                'error_code' => 'access_denied',
+                'title' => jLocale::get('dataviz~dataviz.log.access_denied.title'),
+                'detail' => jLocale::get('dataviz~dataviz.log.access_denied.detail'),
             );
 
             return;
         }
 
         // Get config
-        $datavizConfig = $lproj->getDatavizLayersConfig();
+        $datavizConfig = $lizmapProject->getDatavizLayersConfig();
         if (!$datavizConfig) {
             $this->errors = array(
-                'title' => 'Dataviz Configuration not found',
-                'detail' => 'No dataviz configuration has been found for this project',
+                'code' => 404,
+                'error_code' => 'no_configuration',
+                'title' => jLocale::get('dataviz~dataviz.log.no_configuration.title'),
+                'detail' => jLocale::get('dataviz~dataviz.log.no_configuration.detail'),
             );
 
             return;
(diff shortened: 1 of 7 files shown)


Gustry:
Can you please add the URL to use as exemple ?
Some parameters are in the query param and the plotconfig in the post ?

mdouchin:
> Can you please add the URL to use as exemple ? Some parameters are in the query param and the plotconfig in the post ?

Done in the PR description

mdouchin:
I would like to backport this feature in 3.6, since it does not change the existing behaviour but will bring a huge improvement for end users

Gustry:
> <script src='https://cdn.plot.ly/plotly-2.16.1.min.js'></script>

Please add in the server metadata this URL, or in this plot response.
So it will avoid any desynchronisation between the plugin and different version of LWC.
Linked to https://github.com/3liz/lizmap-web-client/pull/3288 the version of Plotly will change according to the version of LWC, 3.6 versus 3.7 with this PR.

Gustry:
Can you add as well a flag in the **instance** metadata `api = [{'plotly' : True, 'version': 1.0.0}]`. It will avoid me to maintain a special `if version >= 3.6.1-pre.5432`

Gustry:
[Screencast from 21-12-2022 16:55:59.webm](https://user-images.githubusercontent.com/1609292/208948769-3cf2798c-379b-42fa-aa17-9b9b961b55ca.webm)

rldhont (approved):

Gustry (approved):

Gustry:
Side question, no right is associated with the dataviz API correct ?
Any "authenticated user" can make the POST request right ?
Or do you plan to make a change in #3369 ?

mdouchin:
> Side question, no right is associated with the dataviz API correct ? Any "authenticated user" can make the POST request right ? Or do you plan to make a change in #3369 ?

IMHO, no specific right is needed here. The behaviour is the same for the API as for opening the map in your browser.
The ressources will be availabre depending on the right on the repository and project. Therefore, the user may be unauthenticated, the API will still "work" if the project is public.

PR #3380, opened 2022-12-23
phpstan clean in project.php

nworr:
Cleaning some phpstan error messages in lib/Project.php

Funded by 3liz


Gustry:
Don't you want to backport, to 3.6 at least ? (even 3.5 I think).
It would be easier later if we have some bugfix later on these lines I think.

nworr:
not sure for 3.5 backport :) we'll see.

rldhont (approved):

PR #3397, opened 2023-01-06
Avoid downloading default project img multiple times

nboisteault:
250x250_mappemonde.jpg was downloaded for every project without custom thumbnail
Funded by 3Liz


nboisteault:
@mdouchin Arf I did not think about images with transparency :(

I'll go for the other solution I suggested.

image

PR #3410, opened 2023-01-13
Update the lizmap landing page title

### LINKED ISSUE #1073: Remove the hardcoded "Projects" from the title
Customers think it's confusing to have the projects in title, because in their working context projects are something completely different.
image
They would like to see only "Visualisierung (experimental)".

So what do you think about the possibility to have an option to remove the `jLocale::get("view~default.repository.list.title").' - '.` here?

https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/view/controllers/default.classic.php#L73

Or is this problematic?

Another approach would be, that customers can influence it by deleting the string from the translation file `lizmap/modules/admin/locales/de_DE/admin.UTF-8.properties` but then there is still a lost hyphen. So I would suggest to do something like:
```
if(  jLocale::get("view~default.repository.list.title") ) {
    $title = jLocale::get("view~default.repository.list.title").' - '.$services->appName;
} else {
    $title = $services->appName;
}
```

What does the community think about?

rldhont:
Fixes #1073

* Let the application name alone in the landing page
* Add a new string to translate: Landing page
* Put the page title in the second title in the header

Old landing page

lizmap-landing-page-old

New landing page

lizmap-landing-page-new

Funded by 3Liz

### DIFF
--- a/lizmap/modules/view/controllers/default.classic.php
+++ b/lizmap/modules/view/controllers/default.classic.php
@@ -81,15 +81,17 @@ class defaultCtrl extends jController
             }
         }
 
-        $title = jLocale::get('view~default.repository.list.title').' - '.$services->appName;
+        $title = $services->appName;
+        $subTitle = jLocale::get('view~default.home.title');
 
         if ($repository) {
             $lrep = lizmap::getRepository($repository);
-            $title = $lrep->getLabel().' - '.$title;
+            $subTitle = $lrep->getLabel().' - '.jLocale::get('view~default.repository.list.title');
         }
-        $rep->title = $title;
+        $rep->title = $subTitle.' - '.$title;
 
-        $rep->body->assign('repositoryLabel', $title);
+        $rep->body->assign('title', $title);
+        $rep->body->assign('subTitle', $subTitle);
 
         $auth_url_return = jUrl::get('view~default:index');
         if ($repository) {
--- a/lizmap/modules/view/locales/en_US/default.UTF-8.properties
+++ b/lizmap/modules/view/locales/en_US/default.UTF-8.properties
@@ -1,7 +1,8 @@
 repository.access.denied=You cannot access this repository.
-repository.list.title=Projects
+repository.list.title=Repository page
 
-home.title=Home
+home.title=Home page
+home.menu=Home
 
 server.information.error=This map cannot be displayed. Please contact the server administrator to check the server information panel.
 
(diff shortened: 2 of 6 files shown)


mdouchin (approved):

rldhont:
> refs #1073 ?

Yes, thanks @nworr to pointed it.

lizmap/modules/view/controllers/default.classic.php
```diff
@@ -81,15 +81,17 @@ public function index()
...
+        $title = $services->appName;
+        $subTitle = jLocale::get('view~default.home.title');
 
         if ($repository) {
             $lrep = lizmap::getRepository($repository);
-            $title = $lrep->getLabel().' - '.$title;
+            $subTitle = $lrep->getLabel().' - '.jLocale::get('view~default.repository.list.title');
         }
-        $rep->title = $title;
+        $rep->title = $title.' - '.$subTitle;
```
laurentj, line 91: In case of a project, you will then have "Lizmap - my project blabla". 
Because of how browsers show bookmarks or window title in tabs (as tabs can be tiny), I prefer to have "my project blabla - Lizmap". If a user open several project in his browser, and if he has already several tabs, he could see "Lizmap, m", "Lizmap, f", etc, so it is difficult to find the right tab.

laurentj: To be clear, the title of the window was better before this change. Probably you should not duplicate the title, and you should have a window title different from the page title.

laurentj (changes requested):

lizmap/modules/view/controllers/default.classic.php
rldhont: I have update the window title, it will be `Landing Page - Application name` or ` Repository title - Projects - Application name`

laurentj (approved):

lizmap/modules/view/locales/en_US/default.UTF-8.properties
```diff
@@ -1,7 +1,8 @@
 repository.access.denied=You cannot access this repository.
 repository.list.title=Projects
 
-home.title=Home
+home.title=Landing page
+home.menu=Home
```
mdouchin, line 5:
```suggestion
home.title=Home page
home.menu=Home
```

mdouchin (changes requested):

rldhont:
@mdouchin updating strings done

mdouchin (approved):

PR #3411, opened 2023-01-16
Print enhanced

nboisteault:
Funded by DDTM34

### DIFF
--- a/assets/src/components/Print.js
+++ b/assets/src/components/Print.js
@@ -31,18 +31,14 @@ export default class Print extends HTMLElement {
                 <input type="text" class="print-label"><br>
                 <textarea class="print-label"></textarea>
             </div>
-            <div class="row-fluid">
-                <div class="span4">
-                    <select id="print-format" title="${lizDict['print.toolbar.format']}" class="btn-print-format">
+            <div class="flex">
+                <select id="print-format" title="${lizDict['print.toolbar.format']}" class="btn-print-format">
                     <option value="pdf">PDF</option>
                     <option value="jpg">JPG</option>
                     <option value="png">PNG</option>
                     <option value="svg">SVG</option>
-                    </select>
-                </div>
-                <div class="span8">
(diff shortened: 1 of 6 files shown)


Gustry:
@nboisteault Can you add a screenshot and fix the conflict ?

Gustry:
I'm just making some UX review according to the screenshot.

* I think there should be some units, the rotation is degrees, the grid intervals is in meters, kilometers, degrees, project units ? Maybe use the same display for units as this [GIF](https://github.com/3liz/lizmap-javascript-scripts/tree/master/library/api/ign_web_services) for time&distance ?

* I would also "harmonize" the alignement of these inputs. Sometimes labels are on top, sometimes aside. For instance, the grid interval and then "rotation" brings a shift in the display. Maybe labels should always be on top ?

* I also tried the manual test "print in project projection" : 
  * http://lizmap.local:8130/index.php/view/map?repository=testsrepository&project=print_in_project_projection
  * https://demo.snap.lizmap.com/lizmap_3_6/index.php/view/map?repository=testsrepository&project=print_in_project_projection the red rectangle has a smaller scale visible as soon we click the "print" button.
  * With the PR, the zoom by default is now 50000 while the map is at 5000. So the user doesn't see the new "print extent". I think the print rectangle must be contained within the current view extent, otherwise, the scale must be decrease until we reach the last one. This should be fixed IMHO.
* the grid works only if we fill two values. It doesn't work if choose a single one. Not so important, but just a notice, if a message can be displayed if it's done ?

Gustry:
The new behavior of moving the map is nicer indeed, thanks !

However, I think every-time we scroll (in or out) on the map, it's was nice to have the "biggest" extent possible within the current map extent. (LWC <= 3.6)  It seems this behavior is not there anymore.

Gustry:
@nboisteault 

1. Visit the [demo project on the snap](https://demo.snap.lizmap.com/lizmap_3_7/index.php/view/map?repository=testsrepository&project=print_in_project_projection)
2. Do not scroll/pan
3. Click the print icon. Scale of the map stays at ~5000 but the scale of the print rectangle is at 50 000, so it's not visible

nboisteault:
@Gustry Fixed!

PR #3415, opened 2023-01-18
[Bugfix] Defined default repository rights

rldhont:
To help creating and defining lizmap repository, default rights are now provided to the user.

The default rights are :
* For Anonymous users
  * View projects in the repository
* For users (default group for authenticated users)
  * View projects in the repository
  * Display projects WMS links
  * Allow export of layers
* For admins
  * View projects in the repository
  * Display projects WMS links
  * Allow export of layers
  * Use the Edition tool

lizmap-default-repository-rights

Funded by 3Liz https://3Liz.com

### DIFF
--- a/tests/end2end/playwright.config.js
+++ b/tests/end2end/playwright.config.js
@@ -33,6 +33,7 @@ module.exports = defineConfig({
       height: 650
     },
     screenshot: 'only-on-failure',
+    locale: 'en-US',
     /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
     trace: 'on-first-retry',
   },
--- a/tests/end2end/playwright/maps-management.spec.js
+++ b/tests/end2end/playwright/maps-management.spec.js
@@ -5,7 +5,7 @@ test.describe('Maps management', () => {
 
   test.beforeEach(async ({ page }) => {
     // Go to admin.php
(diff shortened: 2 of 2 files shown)


mdouchin (approved):
👌

Gustry (changes requested):
For me, there is an UX issue if the user needs to scroll **at the bottom** if these checkbox are not stored by default.

There is a message at the top saying : 

> The repository data have been saved.
You can now configure the rights for this repository.

I know the second part of the message is explaining, but it's not clear for the end user, they might not know there is a button at the bottom. A lot of users might think "ok, it's enough, it looks alright, let's go out of this page."

laurentj (approved):
~~lgtm~~ I didn't see the comment of @Gustry

rldhont:
@Gustry with this fix, rights can be set at the creation!

Gustry (approved):
I gave it a quick try, whoua, confusing to have one step less in repository creation !

LGTM when the item below is fixed :

* [ ] Menu item issue

image


Then clicking "Create a repository"

image


For my self, **later**, we do that after the release : 

* [ ] Login/password in test must be stored outside by environment variable etc

rldhont:
Thanks @Gustry to pointed the menu item issue. I have fixed it.

PR #3383, opened 2022-12-30
Launch browser's print box for popups. Print with printer or as pdf.

nboisteault:
Funded by DDTM34

https://user-images.githubusercontent.com/2145040/210054330-985405a4-32d5-4002-8d39-5afc6283f583.mp4

PR #3421, opened 2023-01-20
Filter performance - Avoid to launch many useless GETFILTERTOKEN request

mdouchin:
Funded by Le Grand Narbonne https://www.legrandnarbonne.com/

### DIFF
--- a/assets/src/legacy/map.js
+++ b/assets/src/legacy/map.js
@@ -3502,14 +3502,34 @@ window.lizMap = function() {
      }
      lizMap.events.on({
         "layerFilterParamChanged": function( evt ) {
+            // Continue only if there is a popup displayed
+            // This would avoid useless GETFILTERTOKEN requests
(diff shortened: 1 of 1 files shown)


nboisteault:
@mdouchin I launch Cypress as there is an issue with playwright. If tests are ok, I'm ok.

mdouchin:
@nboisteault test has already been run, and it failed because of the new "print" button in the popup and the corresponding test which has not been actualized ;)

Gustry:
The regression is coming from https://github.com/3liz/lizmap-web-client/pull/3383

Gustry:
FYI the `run cypress` label has no effect on these branches. Only on 3.5 as there isn't any playright test.

Gustry:
Error cypress is on master as well : https://github.com/3liz/lizmap-web-client/actions/runs/3985749533/jobs/6833503093
I'm merging this PR.

PR #3422, opened 2023-01-20
Dataviz - Lazy loading of the plots: get data based on visibility

mdouchin:
At present, all the configured plots are rendered at map startup, just after the map is ready.
This can put a heavy load on the server, if many plots are configured for the map, especially
if the data underneath contain a lot of features.

This PR modifies the way Lizmap get the data of the plots: only the visible plots at startup
will be loaded, and the other will be loaded only when visible in the browser viewport, for example
when the user opens the dataviz panel, or scrolls down to see previously hidden plots.

Funded by 3liz, Grand Narbonne


rldhont (approved):

tests/end2end/cypress/integration/dataviz-ghaction.js
```diff
@@ -74,27 +69,6 @@ describe('Dataviz tests', function () {
...
-        cy.fixture(fixture_path).then((expected) => {
-            // newplot.png is maybe not the last one if the download folder is not empty ...
-            cy.readFile(path.join(downloadsFolder, "newplot.png"), 'base64').then((image) => {
-
-                // We can make a easier function to compare two images
-                const img1 = PNG.sync.read(Buffer.from(image, 'base64'));
-                const { width, height } = img1;
-                const img2 = PNG.sync.read(Buffer.from(expected, 'base64'));
-
-                expect(pixelmatch(img1.data, img2.data, null, width, height, { threshold: 0.1 }), 'expect plot').to.lessThan(expected_diff + 1)
```
nboisteault, line 95: Why removing this test?
For the rest there are so many changes that I think it needs a test to be sure we have the right behavior.

nboisteault (approved):

tests/end2end/cypress/integration/dataviz-ghaction.js
mdouchin: This feature is brought by Plotly, and adds unnecessary complexity (behaviour different from FF and other browser, etc). It failed locally, and this kind of pixel comparison is prone to failures. I discussed with @Gustry and @rldhont, authors of the addition of this test, and they agree to remove it.

mdouchin: it corresponds to the PNG export of a plot, which is not a Lizmap feature but a Plotly feature. I added some PR ago tests to get the JSON configuration sent by the backend and check the returned data is alright.

Gustry: Testing the JSON from plotly https://github.com/3liz/lizmap-web-client/blob/master/tests/end2end/cypress/integration/dataviz-ghaction.js

PR #3423, opened 2023-01-20
[Bugfix] WFS: No PostGIS features returned if map projection != layer.projection

### LINKED ISSUE #3194: [Bug]: Attribute table - No features returned if map projection != layer projection & "Limit fetched data" checked
### What is the bug?

No feature is returned if the attribute table of a layer if 
* the layer projection and the map projection differ
* the option "Limit fetched data to the current map extent and layer visibility" is checked

The [LWC code here](https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/lib/Request/WFSRequest.php#L573) does not take into account the parameter [SRSNAME](https://docs.qgis.org/3.22/en/docs/server_manual/services/wfs.html#wfs-getfeature-srsname) for the WFS GETFEATURE query

To reproduce, use 
* Create and publish a QGIS project in some projection different from `EPSG:3857`, for ex `EPSG:2154`
* Add a PostgreSQL layer with a projection different from `EPSG:3857`, for ex `EPSG:2154`
* Configure attribute table for this layer
* Add a external baselayer in the Baselayers tab of the lizmap plugin

The web map will be in `EPSG:3857`, and the attribute table will try to display the features with a WFS `GETFEATURE` query with a parameter SRSNAME

### Versions

**Versions :**

* Lizmap Web Client : 3.7.0-pre
* Lizmap plugin : master
* QGIS Desktop : 3.22.8
* QGIS Server : 3.22.10
* Py-QGIS-Server : 1.8.3
* QGIS Server plugin atlasprint : not found
* QGIS Server plugin lizmap_server : master
* QGIS Server plugin wfsOutputExtension : not found

### Check Lizmap plugin

_No response_

### Operating system

Ubuntu 20.04

### Browsers

Firefox

### Browsers version

Firefox 104

### Relevant log output

_No response_

rldhont:
The LWC WFSRequest class did not take into account the parameter [SRSNAME](https://docs.qgis.org/3.22/en/docs/server_manual/services/wfs.html#wfs-getfeature-srsname) for the WFS GETFEATURE query when it has to build the SQL request.

fixes #3194

### DIFF
--- a/lizmap/modules/lizmap/lib/Request/WFSRequest.php
+++ b/lizmap/modules/lizmap/lib/Request/WFSRequest.php
@@ -634,6 +634,21 @@ class WFSRequest extends OGCRequest
             }
         }
 
+        $srid = $this->qgisLayer->getSrid();
+        if (array_key_exists('srsname', $params)) {
+            $srsname = $params['srsname'];
+            if (!empty($srsname)) {
+                // SRSNAME parameter is not empty
+                // extracting srid
+                $srsname_id = array_pop(explode(':', $srsname));
+                if (ctype_digit($srsname_id)) {
+                    $srid = intval($srsname_id);
+                } else {
+                    return '';
+                }
+            }
+        }
+
         // Build the SQL
         $xmin = trim($bboxitem[0]);
         $ymin = trim($bboxitem[1]);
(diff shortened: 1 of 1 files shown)


lizmap/modules/lizmap/lib/Request/WFSRequest.php
```diff
@@ -634,14 +634,29 @@ protected function getBboxSql($params)
...
+        if (array_key_exists('srsname', $params)) {
+            $srsname = $params['srsname'];
+            if (!empty($srsname)) {
+                // SRSNAME parameter is not empty
+                // extracting srid
+                $srsname_id = array_pop(explode(':', $srsname));
+                if (ctype_digit($srsname_id)) {
+                    $srid = intval($srsname_id);
+                } else {
+                    return '';
```
mdouchin, line 647: I do not understand what happens in this case ? Why not letting $srid stay the QGIS layer srid instead ?

mdouchin (changes requested):

lizmap/modules/lizmap/lib/Request/WFSRequest.php
rldhont: Because the parameter srsname is false. So I do like false bbox.

mdouchin (approved):

PR #3428, opened 2023-01-23
add XML header to WMTS capabilities

nworr:
This PR add the mandatory xml head `<?xml .. ` to the generated WMTS Capabilities response  (assuming UTF-8 encoding) 

it prevent browser to detect error as above

Wrong XML

 

Funded by 3liz


Gustry:
Whaou, finally, thanks :)

PR #3434, opened 2023-01-24
[Bugfix] Form: A checkbox has to have a default equal to check or uncheck value

rldhont:
To be sure a checkbox is valid, even when it is readonly, the default value has to be equal to check or uncheck value.
https://github.com/jelix/jelix/blob/jelix-1.6.x/lib/jelix/forms/controls/jFormsControlCheckbox.class.php#L53


laurentj (approved):
It seems ok to me.

PR #3436, opened 2023-01-24
Attribute table - Fix no display of table with value relation fields with no layer name

mdouchin:
In some unknown contexts, the `layerName` option is not present any more in the QGIS XML project file in the definition of the `ValueRelation` configuration for a form field.
When Lizmap Web Client displays the form filter or an attribute table, it tries to get the labels for the columns using a Value relation form field. This causes Lizmap to fail to open the attribute table. 

This PR uses the layer ID instead of the `layerName`, which is always present and is by the way a more robust method to get the QGIS layer.

Funded by 3liz


rldhont (approved):

mdouchin:
@nboisteault gentle ping for review

nboisteault (approved):
LGTM

PR #3448, opened 2023-01-31
[Bug] The source key defined by QGIS for Postgres layer may not be a field

rldhont:
Funded by 3liz


nboisteault:
@rldhont What does it fix? Does it need a test?

rldhont:
@nboisteault it fixes error in log. If we add a test we have to check that nothing is wrote to the log if the layer has no primary key and qgis set key='tid' in the datasource when a WFS getFeature request is performed.

mdouchin (approved):

Issue #2590, opened 2021-11-15
[Bug]: Request GetPrintAtlas should provide the requested CRS


Gustry:
### What is the bug?

In order to use ATLAS_PK, the SRS must be in the URL:
https://docs.qgis.org/3.16/en/docs/server_manual/services.html#wms-atlaspk

This request is failing:
http://qgis.server/master?SERVICE=WMS&REQUEST=GetPrint&MAP=/home/etienne/dev/qgis/server_plugin_git/qgis-atlasprint/tests/data/atlas_simple.qgs&ATLAS_PK=2&TEMPLATE=layout1-atlas&FORMAT=png

>  The CRS is not valid. 

This request is working, only adding `SRS=EPSG:4326&`:
http://qgis.server/master?SERVICE=WMS&REQUEST=GetPrint&SRS=EPSG:4326&MAP=/home/etienne/dev/qgis/server_plugin_git/qgis-atlasprint/tests/data/atlas_simple.qgs&ATLAS_PK=2&TEMPLATE=layout1-atlas&FORMAT=png

It would be nice to add this parameter in the request so it would be easier to switch to atlaspk.

@rldhont What do you think ?

### Steps to reproduce the issue

Use the URL above on QGIS Server

### Lizmap version

3.5

### QGIS desktop version

3.22

### QGIS server version

3.22

### Operating system

Ubuntu

### Browsers

Firefox

### Browsers version

Latest firefox

Gustry:
Linked to request https://github.com/3liz/qgis-atlasprint/issues/14

PR #3460, opened 2023-02-03
Popup atlas buttons w/ icons

nboisteault:
- Display buttons for every atlas in popup with defined icon or default

Funded by DDTM 34

Fix https://github.com/3liz/qgis-atlasprint/issues/14


Gustry:
Removing AtlasPrint plugin ?
We would need to test some logic which were incorporated in AtlasPrint if so

Fixing #2590 as well ?

nboisteault:
I merge and I'll make tests in another PR.

PR #3466, opened 2023-02-08
Cypress - Fix tests for QGIS 3.28

Gustry:
Cypress - Fix tests for QGIS 3.28

* [x] Fix hack to get QGIS version as int. Before I was using the environment `QGIS_SERVER_INT` but I have string for now
* [ ] Stop checking the results from QGIS server, this needs to be done later, let's merge this PR first I think

The differences with QGIS 3.28 are : 
* Some text shifting in the PNG
* New `~` characters from https://github.com/qgis/QGIS/pull/50256
* Text in now polygonized in a SVG export

### DIFF
--- a/tests/end2end/cypress/integration/legend-ghaction.js
+++ b/tests/end2end/cypress/integration/legend-ghaction.js
@@ -1,4 +1,4 @@
-import {arrayBufferToBase64} from '../support/function.js'
+import {arrayBufferToBase64, serverMetadata} from '../support/function.js'
 
 describe('Legend tests', function () {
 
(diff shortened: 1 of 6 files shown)


Gustry:
I think I will merge later today if possible.

This could be improved to get the QGIS current version from the metadata and avoid some duplicated requests to the JSON. A simple function async would be better, but I will let this task for others if they want.

tests/end2end/cypress/integration/legend-ghaction.js
```diff
@@ -34,13 +34,20 @@ describe('Legend tests', function () {
...
 
                 let expected_path = 'images/treeview/' + check +'.png'
-                if (Cypress.env('QGIS_SERVER_INT') <= 31600 && check == 'layer_legend_categorized'){
-                    expected_path = 'images/treeview/' + check +'_316.png'
-                }
-
-                cy.fixture(expected_path).then((image) => {
-                    expect(image, 'expect legend to be displayed').to.equal(responseBodyAsBase64)
-                })
+                serverMetadata().then(metadataResponse => {
```
nworr, line 37: test the `check` before calling serverMetadata to prevent usesless call ? 
``` 
if (check == 'layer_legend_categorized') {
  serverMetadata().then(metadataResponse => {
        if (metadataResponse.body.qgis_server_info.metadata.version_int <= 31600 ) {
            expected_path = 'images/treeview/' + check +'_316.png';
        }
    });
 }
 ```
and maybe chaining ` then` will work  ?
```` 
 serverMetadata().then(metadataResponse => {
        if (metadataResponse.body.qgis_server_info.metadata.version_int <= 31600 ) {
            expected_path = 'images/treeview/' + check +'_316.png';
        }
    }).then(metadataResponse => {
         if (metadataResponse.body.qgis_server_info.metadata.version_int <= 32800) {
             // With QGIS 3.28 : https://github.com/qgis/QGIS/pull/50256
             cy.fixture(expected_path).then((image) => {
                  expect(image, 'expect legend to be displayed').to.equal(responseBodyAsBase64)
        })
 };

**Review by nworr — COMMENTED** [review]
some ideas to reduce serverMetadata call, not sure

File: tests/end2end/cypress/integration/legend-ghaction.js (hunk 1 above)
Comment (Gustry) [reply tests/end2end/cypress/integration/legend-ghaction.js:37]: Yes, i can have a look.

The best would be to have a function sync like :  

```javascript
it('should print title labels (PNG)', function () {
    let qgisVersion = serverMetadata();
    // blabla
    let diff = 0;
    if qgisVersion < 32800:
        diff = 230;
    /foo
```

Gustry: Fixed indeed, it's better, thanks

PR #3478, opened 2023-02-14
[Bugfix] Selection by Point - The bbox is empty

rldhont:
To speed up WFS getFeature request to get selected features, the geometry bounds is provided by the BBOX parameter to restrict the request to this bounding box. But if the bounding box is empty, the request returns no features.

So to get feature in intersection with a Point the bounding box has to be extend to not be empty.


rldhont:
@nboisteault tests added

PR #3480, opened 2023-02-14
All project WMS test commands in console.php

nworr:
`project:load jelix command` alternative for console.php using CliHelpers

use a output callback in helper as a simple solution
 
Funded by 3liz


rldhont (approved):

laurentj (approved):
LGTM, thank you

PR #3483, opened 2023-02-14
e2e test print and fixes

nboisteault:


Gustry:
Do you want to take my commit from #3484 if it's linked ?
`print-ghaction.js` is failing.

PR #3485, opened 2023-02-14
Migration script to ease migration from 3.5 to 3.6

laurentj:
The script migrate lizmap 3.5 from 3.6.

* the user should install sources of lizmap 3.6
* he then lauches `lizmap/install/migrate_from_3_5.sh path/to/lizmap3.5/directory/` 
   * it copies configuration and data files from a lizmap 3.5
   * it updates packages into lizmap/my-packages/composer.json and install modules from lizmap-modules as composer package if there is a known package for these modules
   * it launches the configurator and the installer
* the user launches `sudo lizmap/install/clean_vartmp.sh` and `sudo lizmap/install/set_rights.sh www-data www-data`
* the migration is finished

Funded by 3Liz


laurentj:
It has to be tested. We should see if the script should launch the installer or not.
And we should think about the migration from inside the docker container.

rldhont (approved):

Issue #3474, opened 2023-02-10
[Bug]: zoom to feature (from the attribute table) goes to far away


josemvm:
### What is the bug?

my all lizmap project are in epsg 3763 and my layers too. i have basemaps in epsg 3857

when i try to zoom a feture from table, the map goes to far far away!...

### Steps to reproduce the issue

open one table and try zoom to feature

i can provide an url to check the problem

### Versions

**Versions :**

* Lizmap Web Client : 3.6.2-pre.5961
* Lizmap Web Client target version : 3.6
* Lizmap plugin : 3.10.1
* QGIS Desktop : 3.16.16
* QGIS Server : 3.16.16
* Py-QGIS-Server : not used
* QGIS Server plugin atlasprint : 3.3.2
* QGIS Server plugin lizmap_server : 1.3.1
* QGIS Server plugin wfsOutputExtension : 1.7.1

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### QGIS server version, only if the section above doesn't mention the QGIS Server version

3.16.16

### Operating system

Ubuntu 20.04

### Browsers

Firefox

### Browsers version

Firefox 108.0.2

### Relevant log output

```shell
no logs
```

Gustry:
I think I got the issue as well on 3.6.0 a few weeks ago on a local project.

josemvm:
hi @Gustry and @rldhont please can you take a look to this problem? due to this problem i was forced to downgrade again  to 3.5...
thanks in advance

rldhont:
I found the bug

josemvm:
@rldhont sorry, but the problem remains!...

Gustry:
Did you download the zip from packages ? Did you apply the diff manually ?

josemvm:
Lizmap Team thanks a lot for your work!
👍

PR #3489, opened 2023-02-16
[Bugfix] zoom to feature (from the attribute table) goes to far away

rldhont:
Fixes https://github.com/3liz/lizmap-web-client/issues/3474

If the layers are not in EPSG:4326 to zoom button in the attribute table goes the map to far away.

And `getDataAndFillAttributeTable` and `getFeatureData` as iso

Funded by 3liz https://3liz.com/

### DIFF
--- a/assets/src/legacy/attributeTable.js
+++ b/assets/src/legacy/attributeTable.js
@@ -268,7 +268,8 @@ var lizAttributeTable = function() {
 
             function getDataAndFillAttributeTable(layerName, filter, tableSelector, callBack){
 
-                const typeName = lizMap.config.layers[layerName].typename;
+                let layerConfig = lizMap.config.layers[layerName];
+                const typeName = layerConfig.typename;
 
                 const wfsParams = {
                     TYPENAME: typeName,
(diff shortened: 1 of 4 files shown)


assets/src/legacy/attributeTable.js
```diff
@@ -335,8 +343,14 @@ var lizAttributeTable = function() {
...
                     }
-                    lizMap.config.layers[layerName]['columns'] = responses[1].columns;
-                    buildLayerAttributeDatatable(layerName, tableSelector, responses[0].features, responses[1].aliases, responses[1].types, allColumnsKeyValues, callBack);
+                    layerConfig['featureCrs'] = 'EPSG:4326';
+                    if (namedRequests?.['describeFeatureType']) {
+                        const describeFeatureTypeResponse = responses[namedRequests['describeFeatureType']];
+                        layerConfig['aliases'] = describeFeatureTypeResponse.aliases;
+                        layerConfig['types'] = describeFeatureTypeResponse.types;
+                        layerConfig['columns'] = describeFeatureTypeResponse.columns;
+                    }
```
mdouchin, line 352: Shouldn't we test that `describeFeatureTypeResponse` contains `aliases`, `types` and `columns`

mdouchin (changes requested):

assets/src/legacy/attributeTable.js
rldhont: This request with json output format has been created for that - it is a dedicated lizmap request.

rldhont: I have added a test, to check each properties from the request.

mdouchin (approved):

rldhont:
I have added tests for https://github.com/3liz/lizmap-web-client/pull/3495

PR #3502, opened 2023-02-19
Admin: allow to edit the name for the sender email

laurentj:
This property should not be considered as a sensitive parameter. We show also the email used to send emails, even if we keep as a non editable value in sensitive mode.

Funded by 3liz


mdouchin (approved):

rldhont:
@laurentj php unit tests failed

PR #3511, opened 2023-02-22
Tests with some QGIS 3.28 versions

Gustry:
Testing purpose only...

All tests files have been opened with a updated version of [audit qgis project](https://github.com/3liz/QGIS-Resources/blob/master/collections/processing-scripts/processing/AuditQGISProjects.py) script.


Gustry:
These 2 tests are also failing on my local.

PR #3527, opened 2023-02-27
[Backport release_3_5] [Bugfix] Defined default repository rights

rldhont:
Backport https://github.com/3liz/lizmap-web-client/pull/3415
Authored by: @rldhont


Gustry (approved):
Go for it, I will check for the cron job with playwright.

PR #3534, opened 2023-03-03
[Bugfix] WMS GetFeatureInfo Filter: the layer name has to be the same

rldhont:
The QGIS Server WMS Filter parameter has to start with the wms layername to identify the layer to apply filter. The layername has to be the same the one for LAYERS and QUERY_LAYERS parameters.


nboisteault (approved):
LGTM

Issue #3544, opened 2023-03-13
[Question]: Problem with Lizmap and Qgis server on Ubuntu 22.04


abrousseau89:
### What is the question?

Good morning,

I'm trying to deploy an Ubuntu 22.04 server with Lizmap 3.6 and Qgis Server. After a lot of hassle I now access the Lizmap interface.

When I check in the administration of Lizmap I have this message:

`QGIS Server
We cannot have the details on the installation of your QGIS server (version, extensions etc.).`

and in the Lizmap Logs I have this:

`023-03-13 08:31:42    192.168.20.117    error    QGIS Serveur doit être mise à jour au moins avec la version 3.10pour cette version de Lizmap Web Client.
2023-03-13 08:31:42    [2]    Undefined array key "plugins"    /var/www/lizmap-web-client-3.6.2/lizmap/modules/lizmap/lib/Server/Server.php    54

2023-03-13 08:31:42    [2]    Trying to access array offset on value of type null    /var/www/lizmap-web-client-3.6.2/lizmap/modules/lizmap/lib/Server/Server.php    54

2023-03-13 08:31:42    [2]    Trying to access array offset on value of type null    /var/www/lizmap-web-client-3.6.2/lizmap/modules/lizmap/lib/Server/Server.php    54

2023-03-13 08:31:42    192.168.20.117    error    L'extension lizmap_serverdoit être mise à jour.`

after checking on Ubuntu, I am however using the latest versions and my configuration seems correct... Maybe an oversight somewhere?

thank you for your information

### Versions

Versions :

 Lizmap Web Client : 3.6.2
 Lizmap Web Client target version : 3.6
 Lizmap plugin : 3.10.1
 QGIS Desktop : 3.22.10
 QGIS Server and plugins unknown status because of the settings in QGIS Server, please review your server settings in the Lizmap Web Client administration interface, then in the "Server Information" panel.

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 22.04

### Browsers

Firefox

### Browsers version

Firefox 110

nboisteault:
Thanks @abrousseau89 for your report. We reproduce the bug and we'll fix it soon.

rldhont:
@abrousseau89 to be sure your QGIS Server install is complete for Lizmap Web Client 3.6, did you find in QGIS Server logs this line:
[...]

PR #3548, opened 2023-03-14
Fix Server information admin panel

rldhont:
Fix closing tr element and not displayed not named plugin

Ticket : # 

Funded by

### DIFF
--- a/lizmap/modules/admin/templates/server_information.tpl
+++ b/lizmap/modules/admin/templates/server_information.tpl
@@ -123,8 +123,9 @@
             {if $displayPluginActionColumn }
                 <th>{@admin.server.information.qgis.plugin.action@}</th>
             {/if}
-        <tr/>
+        </tr>
         {foreach $data['qgis_server_info']['plugins'] as $name=>$version}
+        {if $version['name']}
         <tr>
             <th style="width:20%;">{$version['name']}</th>
             <td style="width:20%;">
@@ -147,6 +148,7 @@
                 {/if}
             {/if}
         </tr>
+        {/if}
         {/foreach}
     </table>
     {hook 'QgisServerPlugins', $data['qgis_server_info']['plugins']}


lizmap/modules/admin/templates/server_information.tpl
```diff
@@ -123,8 +123,9 @@
             {if $displayPluginActionColumn }
                 <th>{@admin.server.information.qgis.plugin.action@}</th>
             {/if}
-        <tr/>
+        </tr>
         {foreach $data['qgis_server_info']['plugins'] as $name=>$version}
+        {if $version['name']}
```
Gustry, line 128: The idea was to show that plugins **are** missing for WfsOutputExtension and AtlasPrint.
With this PR, we are not advertising possible plugins to enhance LWC anymore.
Either we let the server providing the name (next version available), or only L130, if instead `$version['name']` is not defined you should switch to the JSON key.

A fix has been made in the server side to always add the name.

Gustry (changes requested):

Gustry:
Linked to #3544

lizmap/modules/admin/templates/server_information.tpl
rldhont: I add this because I have nothing in the first colonne and not found in the second one. So my fix could be: 
```
{if $version['name']}
<th style="width:20%;">{$version['name']}</th>
{else}
<th style="width:20%;">{$name</th>
{/if}
```

Gustry: > I add this because I have nothing in the first colonne

Yes, this is fixed in the next version of the plugin

> and not found in the second one

Yes, this is on purpose. So I can control from the plugin, easier to release than a LWC version

Your proposed fix looks good. But anyway, it's already fixed in the next version of the plugin. And the minimum lizmap server version required is hard-coded and upgraded for each LWC release now. So your `else` should never be called I think.

Let's close this PR, I will commit straight about the `<tr/>`. I should be alright.

rldhont: Do not close this PR, I'd like to add tests to this page.

Gustry:
Both issues has been fixed : cbe7454b6e8d248b6ccd390bdbdf6a26ca23e689 and https://github.com/3liz/qgis-lizmap-server-plugin/commit/eb6a773ba035f877e9fa91db5ef87911a2648ee1

PR #3572, opened 2023-03-30
[Bugfix] Regression in view/media/getCssFile

rldhont:
Reset the fileName jResponseBinary property because Lizmap update the CSS content to replace url link.

Funded by AUDELOR


laurentj (approved):
r+

PR #3589, opened 2023-04-04
Raise QGIS minimum version to 3.22

Gustry:
Raise QGIS minimum version to 3.22 for the server.

It doesn't impact versions in projects for being visible.

It means we jumped from QGIS 3.10 to 3.22 for LWC 3.7


Gustry:
I guess now, it's more than OK.

PR #3608, opened 2023-04-12
[Feature] Display feature geometries defined in map parameter

rldhont:
If the parameter `layer` and `filter` is defined, the selected feature geometries has to be displayed.

Ticket: https://github.com/3liz/lizmap-web-client/pull/3074

Funded by natur&ëmwelt https://www.naturemwelt.lu/


mdouchin (approved):
Thanks

PR #3613, opened 2023-04-13
[Bugfix] Use shortname in selection token if provided

rldhont:
The GetSelectionToken method does not check if the provided typename is the right wms name. The wms name of a layer is the shortname or the name of the layer.

Funded by SMICA https://www.smica.fr/


Gustry:
Just to keep reference, starting from release of [3.11.2](https://github.com/3liz/lizmap-plugin/releases/tag/3.11.2) of the QGIS desktop Lizmap plugin, **and** if LWC ≥ 3.6, there is a shortname.

https://github.com/3liz/lizmap-plugin/pull/431

PR #3634, opened 2023-04-17
UX - Add reminder to check the QGIS server URL when the table is not correct

Gustry:


lizmap/modules/admin/locales/en_US/admin.UTF-8.properties
```diff
@@ -267,6 +267,7 @@ server.information.qgis.error.fetching.information=We cannot get the details abo
 server.information.qgis.error.fetching.information.description=You must check that the followings steps are done :
 server.information.qgis.error.fetching.information.qgis.version.html=<strong>QGIS Server</strong> is above or equal to <strong>%s</strong>
 server.information.qgis.error.fetching.information.plugin.version.html=The <strong>Lizmap QGIS server plugin</strong> is installed on the server with a version above or equal to <strong>%s</strong>
+server.information.qgis.error.fetching.information.qgis.url=The URL to reach QGIS server is correct in Lizmap Web Client
```
mdouchin, line 270: "correct" or "not correct"

Gustry: It's used in a checklist, things that must be done in order to solve the "error".

PR #3629, opened 2023-04-14
Add webdav informations into Lizmap metadata

laurentj:
Funded by 3Liz

PR #3635, opened 2023-04-17
UX - Display the URL of webdav if possible

Gustry:
Follow up from https://github.com/3liz/lizmap-web-client/pull/3629 and #3642

I guess we want to make our users aware later.


mdouchin (approved):

Gustry:
Question, shouldn't we use the right `lizmap.webdav.access` ? (I think it must have been done in #3629 from @laurentj). It's specified in the module source code.

PR #3638, opened 2023-04-18
Refactor layers treeview

nboisteault:
Funded by [Département du Calvados](https://www.calvados.fr/accueil.html)


rldhont:
@nboisteault can be closed ?

Issue #1510, opened 2020-02-10
base layer - order by name


josemvm:
hi

![imagem](https://user-images.githubusercontent.com/10053874/74173919-e7d42980-4c2a-11ea-8282-84c153ef321d.png)

the base layers have a random order
it's good to have the layers order given by the plugin

thanks

Gustry:
Hi @josemvm 

We really appreciate your early testing and feedbacks on LWC. You are spotting bugs in the early stage of development and this is appreciated to have beta testing.

I invite you to read the blogpost from Nyall Dawson, the main contributor in core QGIS : https://nyalldawson.net/2016/08/how-to-effectively-get-things-changed-in-qgis/
Lizmap and QGIS is slightly different, because QGIS is not owned by a company but we are trying to make Lizmap an opensource project with a community on github.

Please read the latest changelog about the 3.5 release : https://www.3liz.com/news/lizmap-web-client-3-5.html
You will notice that new features are :  
- either coming from the GitHub opensource community (contribution from Florian, Igor...)
- or paid by customers. Their logo is showed and paid features are listed.

At 3Liz, we do spend times on non-paid features in Lizmap Web Client, refactoring code, doing our manual testing which is invisible for end users like you. You can read [our blog post](https://www.3liz.com/blog/rldhont/index.php?post/2021/03/18/On-our-way-to-quality) how we are trying to improve testing in Lizmap. That's why I'm adding the label "sponsor or PR needed".

josemvm:
hi @Gustry 

thank you very much to point me to the excellent writeup from Nyall Dawson!

i really understand your point of view and i clearly assume my natural tendency to try to shape lizmap a little bit according to my expectations (as in everything in life) - sorry for that...

i really appreciate the lizmap project in particular and all other work and technical skills of the 3Liz team!- thank you.

it' s a great pleasure to me to help the project and community, doing some tests and reporting the results :smiley:

nboisteault:
Now w/ Lizmap 3.7, you define a `baselayers` group and put your layers in it in the order you want.

PR #3656, opened 2023-04-27
Baselayers ol7

nboisteault:
Let's finish https://github.com/3liz/lizmap-web-client/pull/2935
Funded by 3Liz


assets/src/modules/BaseLayersMap.js
```diff
@@ -0,0 +1,232 @@
...
+            }
+            const minResolution = Utils.getResolutionFromScale(params.minScale);
+            const maxResolution = Utils.getResolutionFromScale(params.maxScale);
+
+            if (params.cached === "False") {
+                this.addLayer(new ImageLayer({
+                    extent: extent,
+                    minResolution: minResolution,
+                    maxResolution: maxResolution,
+                    visible: params.toggled === "True",
```
Gustry, line 179: Can you make a global function to parse either a string or bool and return its boolean value ? 
Similar to https://github.com/3liz/lizmap-plugin/blob/master/lizmap/tools.py#L180...L191

It would be nice to close your own ticket :)
https://github.com/3liz/lizmap-plugin/issues/176
It's one line of code in lizmap plugin to switch (I can make the switch when the server is targeting > LWC 3.7

Side note, the comparaison L179 looks impossible after L174 ?

nboisteault: We talked with @mdouchin having the correct cfg sent by the server in Lizmap 3.7. I'll then have to replace all "True"/"False" tests.

Gustry: I mean calling a global function like the existing one https://github.com/3liz/lizmap-web-client/blob/78b57ac29f67880a12d5072b00bf422b59f67e58/lizmap/www/assets/js/dataviz/dataviz.js#L85

josemvm:
hi @nboisteault can you take it into account?
https://github.com/3liz/lizmap-web-client/issues/1510

assets/src/modules/BaseLayersMap.js
```diff
@@ -0,0 +1,399 @@
...
+    }
+
+    /**
+     * Return overlay layer if `typeName` matches
+     */
+    getLayerByTypeName(typeName){
+        return this.overlayLayers.find(
+            layer => layer.getSource().getParams?.()?.LAYERS === typeName
+        );
+    }
```
rldhont, line 398: The `LAYERS` is not the typeName, it is a wmsName.

nboisteault: I'll fix it

assets/src/modules/BaseLayersMap.js
```diff
@@ -0,0 +1,399 @@
...
+                });
+
+                if (node.name !== 'root') {
+                    layerGroup.setVisible(layerCfg?.toggled === "True");
+                    layerGroup.setProperties({
+                        name: node.name,
+                        parentName: parentName,
+                        mutuallyExclusive: layerCfg?.mutuallyExclusive === "True",
+                        groupAsLayer: layerCfg?.groupAsLayer === "True"
+                    });
```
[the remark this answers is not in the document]
nboisteault: Yes, I'll change it when your work will be merged.

assets/src/modules/config/Layer.js
```diff
@@ -27,6 +27,7 @@ const requiredProperties = {
 };
 
 const optionalProperties = {
+    'noLegendImage': {type: 'boolean'},
```
rldhont, line 30: Duplicated key.

assets/src/modules/state/MapLayer.js
```diff
@@ -178,6 +178,7 @@ export class MapItemState extends EventDispatcher {
         }
         // Calculate visibility
         this.calculateVisibility();
+        mainEventDispatcher.dispatch({ type: 'overlayLayer.visibility.changed', name: this.name, checked: this._checked});
```
rldhont, line 181: This setter has to be 
```
    /**
     * Set layer tree item is checked
     *
     * @type {Boolean}
     **/
    set checked(val) {
        this._layerItemState.checked = val;
    }
```

assets/src/modules/state/MapLayer.js
```diff
@@ -605,4 +610,34 @@ export class MapLayerState extends MapItemState {
...
+        const newVal = convertBoolean(loading);
+        // No changes
+        if (this._loading == newVal) {
+            return;
+        }
+        // Set new value
+        this._loading = newVal;
+
+        this.dispatch({
+            type: 'overlayLayer.loading.changed',
```
rldhont, line 638:
```suggestion
            type: 'layer.loading.changed',
```

tests/end2end/playwright/filter-layer-by-user.spec.js-snapshots/map-not-connected-chromium-linux.png
Gustry, line 1: Is-it just a test output for debugging or is-it used for image comparaison ?
IMHO, we should do only QGIS server GET request comparaison.

nboisteault: We should only QGIS Server request comparison when there is just QGIS Server involved but here there is also some Lizmap backend logic.

rldhont:
@nboisteault ready to merge, isn't it ?

PR #3658, opened 2023-04-28
Remove deprecated call to the Lizmap server plugin

Gustry:
Since LWC 3.6, we know that the plugin is installed and updated before we open a map, so we can remove some code...

The idea is to remove the GET request to QGIS server : `SERVICE=LIZMAP&REQUEST=GetServerSettings`
It's handled by the JSON metadata.


Gustry:
Hum, end2end is happy, another PR with only some lines removed :)

PR #3684, opened 2023-05-11
Data filtering - Fix the filter not applied to the children layers

mdouchin:
The parameters sent to QGIS Server were not well formated which prevented the filter from beeing applied on the related layers

* Add spaces between the values in the SQL "IN" clause
* Remove the unneeded layer name as a prefix for the EXP_FILTER parameter

Ticket : # 

Funded by 3Liz


mdouchin:
@rldhont @nboisteault ready for review. I fixed the end2end tests

PR #3726, opened 2023-06-13
Config: Add layerType and extract xyz and WMTS 3857 datasource parameters

rldhont:
The external access property of layer config can contain XYZ and WMTS 3857 parameters, extracted dynamically from QGIS XML.

layerType is also added to layer config information.

### DIFF
--- a/assets/src/modules/config/BaseLayer.js
+++ b/assets/src/modules/config/BaseLayer.js
@@ -49,10 +49,14 @@ export class BaseLayerConfig extends BaseObjectConfig {
 
         this._hasAttribution = false;
         this._attribution = null;
-        if (cfg.hasOwnProperty('attribution')
-            && Object.getOwnPropertyNames(cfg['attribution']).length != 0) {
-            this._attribution = new AttributionConfig(cfg['attribution']);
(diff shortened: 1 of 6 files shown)


mdouchin (approved):

assets/src/modules/config/BaseLayer.js
```diff
@@ -353,6 +357,23 @@ export class WmtsBaseLayerConfig extends BaseLayerConfig {
         }
 
         super('wmts', name, cfg, wmtsProperties, wmtsOptionalProperties);
+
+        // Remove unnecessary parameters
+        let wmtsUrl = new URL(this._url);
+        let keysToRemove = []
+        for (const [key, ] of wmtsUrl.searchParams) {
```
nboisteault, line 364:
```suggestion
        for (const key of Object.keys(wmtsUrl.searchParams)) {
```

nboisteault (approved):

assets/src/modules/config/BaseLayer.js
rldhont: It does not work, searchParams does not contain keys. It is an array of key values.

PR #3727, opened 2023-06-13
[Bugfix] Fix visibility preset error

### LINKED ISSUE #3519: [Bug]: Theme/VisibilityPreset error (confusion between expanded and visible)
### What is the bug?

When selecting a theme in a project with layer groups, sometimes the visible layer are not correct.

-

### Versions

3.5.8

### Check Lizmap plugin

- [ ] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

3.28

### Operating system

Ubuntu 20.4

### Browsers

Chrome

### Browsers version

Chrome 102

### Relevant log output

_No response_

rldhont:
Fixes #3519 the right way and added tests

Funded by 3Liz

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/QgisProject.php
+++ b/lizmap/modules/lizmap/lib/Project/QgisProject.php
@@ -1265,6 +1265,12 @@ class QgisProject
                 // Copy layers and their attributes
                 foreach ($theme->layer as $layer) {
                     $layerObj = $layer->attributes();
+                    // Since QGIS 3.26, theme contains every layers with visible attributes
+                    // before only visible layers are in theme
+                    // So do not keep layer with visible != '1' if it is defined
+                    if (isset($layerObj->visible) && (string) $layerObj->visible != '1') {
+                        continue;
+                    }
                     $themes[(string) $themeObj->name]['layers'][(string) $layerObj->id] = array(
                         'style' => (string) $layerObj->style,
                         'expanded' => (string) $layerObj->expanded,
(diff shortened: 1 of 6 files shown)


rldhont:
@nboisteault cleanning tmp is needed before testing theme.

@TeoGoddet here the way to fully fix #3519

lizmap/modules/lizmap/lib/Project/QgisProject.php
```diff
@@ -1268,6 +1268,7 @@ protected function readThemes($xml)
                     $themes[(string) $themeObj->name]['layers'][(string) $layerObj->id] = array(
                         'style' => (string) $layerObj->style,
                         'expanded' => (string) $layerObj->expanded,
+                        'visible' => isset($layerObj->visible) ? (string) $layerObj->visible : '1',
```
rldhont, line 1271: @nboisteault  To avoid getting all layers, we probably can do something like that:
```php
                     // do not keep layer with visible != '1' if it is defined
                     if ( isset($layerObj->visible) && (string) $layerObj->visible != '1')
                        conitnue;
                    $themes[(string) $themeObj->name]['layers'][(string) $layerObj->id] = array(
                        'style' => (string) $layerObj->style,
                        'expanded' => (string) $layerObj->expanded,
                    );
```
And revert the JS changes.

rldhont:
@nboisteault I have added tests

nboisteault (approved):
As `lizmap/modules/lizmap/classes/qgisProject.class.php` is deprecated, why doing changes also in it?

TeoGoddet:
@rldhont huge thanks for fixing it, i'm sorry for the errors

rldhont:
> As `lizmap/modules/lizmap/classes/qgisProject.class.php` is deprecated, why doing changes also in it?

Even if it is deprecated, it is still here and need to be maintained. If we want to removed it, we have to check every modules that could use it.

nboisteault:
> @rldhont huge thanks for fixing it, i'm sorry for the errors

@TeoGoddet don't be sorry. Your contribution is really appreciated.

PR #3730, opened 2023-06-14
[JavaScript] Extract layers order from config and WMS Capabilities

rldhont:
Continuing to build the JavaScript API by defining a way to access layers order.

The `Config.layersOrder` provides the layer names displaying order like in QGIS:
* The first one in this list is the top one in the map
* The last one in this list is the bottom one in the map


Gustry:
Does it work with the "custom layer order" ?

rldhont:
@Gustry yes it is

PR #3736, opened 2023-06-16
[Bugfix] WMS GetLegendGraphic is the right request name not GetLegendGraphics

rldhont:
In OGC WMS standard the request `GetLegendGraphics` and `GetLegendGraphics` are the same, but the right name is `GetLegendGraphic` without an `s`.

### DIFF
--- a/lizmap/modules/lizmap/lib/Request/WMSRequest.php
+++ b/lizmap/modules/lizmap/lib/Request/WMSRequest.php
@@ -265,12 +265,15 @@ class WMSRequest extends OGCRequest
      *
      * @see https://en.wikipedia.org/wiki/Web_Map_Service#Requests.
      */
-    protected function process_getlegendgraphic()
+    protected function process_getlegendgraphics()
     {
-        return $this->process_getlegendgraphics();
+        // The right request name is GetLegendGraphic not GetLegendGraphics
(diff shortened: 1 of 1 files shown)


lizmap/modules/lizmap/lib/Request/WMSRequest.php
```diff
@@ -265,13 +265,21 @@ protected function checkMaximumWidthHeight()
...
 
-    protected function process_getlegendgraphics()
+    protected function process_getlegendgraphic()
     {
+        // The right request name is GetLegendGraphic not GetLegendGraphics
+        $wmsRequest = strtolower($this->param('request'));
+        if ($wmsRequest == 'getlegendgraphic') {
+            $this->params = array_merge(
+                $this->params,
+                Proxy::normalizeParams(array('request' => 'getlegendgraphic'))
```
nworr, line 280: Params were normalized by OGCRequest constructor

https://github.com/3liz/lizmap-web-client/blob/74dee7ddfd979f40d331305b3caebd026603f97e/lizmap/modules/lizmap/lib/Request/OGCRequest.php#L77

why not
```php
$this->params['request'] = 'getlegendgraphic';
```

rldhont: you are right @nworr

PR #3813, opened 2023-07-28
check project scales and replace scale 0 with scale 1

nworr:
When the user set the projects min scale to 0, it leads to a fatal "Division by zero" error : 

see https://github.com/3liz/lizmap-web-client/blob/68ab4107642a9cb7baf106891faf108e9e2220e6/lizmap/modules/lizmap/classes/lizmapTiler.class.php#L256
scaledenomiator will be divided  until having out of bound value : 0

Fix is done on the plugin side https://github.com/3liz/lizmap-plugin/commit/392299aa54875c4e409ef5b7e31f39371d4f6292, but old projects conf may still have the 0 value

Funded by 3Liz

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmapTiler.class.php
+++ b/lizmap/modules/lizmap/classes/lizmapTiler.class.php
@@ -219,9 +219,7 @@ class lizmapTiler
             $rootExtent[3] = $geoExtent[3];
         }
 
-        $scales = array_merge(array(), $project->getOption('mapScales'));
-        rsort($scales);
-
+        $scales = self::normalizeProjectScales($project);
         $projection = $project->getOption('projection');
 
         $tileMatrixSetList = array();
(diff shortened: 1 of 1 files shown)


lizmap/modules/lizmap/classes/lizmapTiler.class.php
```diff
@@ -602,4 +599,16 @@ public static function getTileBbox($tileMatrixSet, $tileMatrixId, $tileRow, $til
 
         return (string) $minx.','.(string) $miny.','.(string) $maxx.','.(string) $maxy;
     }
+
+    private static function normalizeProjectScales($project)
+    {
+        $scales = array_merge(array(), $project->getOption('mapScales'));
+        if (reset($scales) == 0) {
```
mdouchin, line 606: If we are sure that the array contains values ordered from minimum value to maximum values, why not just use 
```suggestion
        if ($scales[0] == 0) {
```

nworr: We're sure :  
https://github.com/3liz/lizmap-plugin/blob/392299aa54875c4e409ef5b7e31f39371d4f6292/lizmap/plugin.py#L1371

mdouchin (approved):

PR #3814, opened 2023-07-28
Mailer: set the replyTo parameter

laurentj:
Set it with the adminContactEmail value only if the new parameter `setAdminContactEmailAsReplyTo` is true.

It allows for jCommunity and Jelix to send emails with a default `Reply-To` header that could be different from the "webmaster" email.

Funded by 3Liz


laurentj:
the backport into 3.5 need a little change into the upgrade script..

PR #3831, opened 2023-08-04
New lizmap log file to show into lizmap admin.

laurentj:
The admin page that shows log, now show only logs we want, that is usefull for the user.

Funded by 3Liz


mdouchin (approved):
🚀

Issue #3829, opened 2023-08-02
[Bug]: UI for mobile. Activate Hamburger button on Chrome


Crapaud04:
### What is the bug?

When using lizmap on a tablet I realized that the group menu does not appear. Is desktop UI.

![menu_mobile](https://github.com/3liz/lizmap-web-client/assets/37582771/0706ab46-d842-4af7-9fff-8435c16c71d7)

The problem is on Chrome and not Firefox.

### Steps to reproduce the issue

1. Open project on Chrome on tablet

### Versions

**Versions :**

* Lizmap Web Client : 3.6.4
* Lizmap plugin : 3.15.1
* QGIS Desktop : 3.22.16
* Lizmap.com : Yes
* QGIS Server : 3.22.16
* Py-QGIS-Server : 1.8.8
* QGIS Server plugin atlasprint : 3.3.2
* QGIS Server plugin cadastre : 1.17.3
* QGIS Server plugin lizmap_server : 2.8.0
* QGIS Server plugin wfsOutputExtension : 1.7.1

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### QGIS server version, only if the section above doesn't mention the QGIS Server version

3.22.16

### Operating system

Serveur Valabre

### Browsers

Chrome

### Browsers version

Chrome 115

nworr:
quick analyse : it seems linked to css rule for detecting non-desktop device `pointer :coarse`

see 
https://github.com/3liz/lizmap-web-client/blob/9196bcc0d3a4a53ab303de30ff178be6bc1d5eba/lizmap/www/assets/css/media.css#L177
and  https://github.com/3liz/lizmap-web-client/blob/9196bcc0d3a4a53ab303de30ff178be6bc1d5eba/lizmap/www/assets/css/media.css#L431

Crapaud04:
Thank you for your message.

I read this blog post following your message.
https://www.ctrl.blog/entry/css-media-hover-samsung.html

Do you think we can change it css just on samsungs?

nworr:
Thansk @Crapaud04  for the link, in comments there's another link 
https://andreasburg.de/browser-hover-check.html

For a confirmation, can you test this page with your tablet and see if pointing on the middle button works ?

[...]

Issue #2895, opened 2022-04-05
[Bug]: Hamburger menu not working with embedded project on mobile


u-cav:
### What is the bug?

Hi! It seems like the hamburger menu is not working correctly when viewing an embedded LWC project on mobile.

You can see a quick demo here:

https://www.faunalia.eu/en/lizmap-embed-test

Everything works correctly on the desktop.

On mobile, the hamburger menu does not work when embedding the map with the 'view/embed' link.

Embedding the full version ('view/map'), works correctly (see last example on the webpage).

Opening the menu, the icon changes and the underlying box moves slightly, but no content is displayed:

![Screenshot_20220405-091310_Bromite](https://user-images.githubusercontent.com/4831528/161684008-59ce5591-4a31-4888-a9c3-725f1bba9393.png)

Tested with Bromite 100.0.4896.57 (Chromium based) and Firefox 98.3.0

### Steps to reproduce the issue

1. On mobile, go to https://www.faunalia.eu/en/lizmap-embed-test
2. Click on the hamburger menu of one of the first three examples, no menu is shown

### Versions

Lizmap 3.5.2

### QGIS server version, only if the section above doesn't mention the QGIS Server version

QGIS Server 3.22

### Operating system

Android 10 on mobile, Debian 11 on the server

### Browsers

Firefox, Chrome

### Browsers version

Bromite 100.0.4896.57 (Chromium based), Firefox 98.3.0

nworr:
it seems like a conflict between the embed style and the responsive style, using onlye the embed style may be a easy solution ?

PR #3836, opened 2023-08-10
 remove hover:none rule (untrustable) 

nworr:
On some tablet browser (samsung) the `hover:none` css rule won't be match, so the left menu will be display in "desktop mode"

the PR remove the rule so the css une only `pointer: coarse` 

Ticket : #3829

Funded by 3Liz


mdouchin (approved):

Gustry:
Is there a link with #2895 ?

PR #3837, opened 2023-08-11
Fix address search when results are empty

laurentj:
Sometimes api-adresse.data.gouv.fr doesn't return a "features" array and it causes PHP errors. The JSON content should be checked before processing list of features.

Funded by 3Liz


mdouchin (approved):
Thanks

PR #3853, opened 2023-08-21
Fix export layer with accent

### LINKED ISSUE #3660: [Bug]: Export dropdown do not show in pop up if layer name in QGIS project has an accent
### What is the bug?

Scenario: QGIS project with some layer published as WFS and with the pop up dialog active. If permissions are granted is expected to see the export layer dropdown in the pop up dialog.

image

But...

if the layername in the QGIS project has an accent, i.e.

image

then it won't show. Actually I just noticed that if the layer name (in QGIS project) even start with an uppercase latter, also the export layer dropdown is a no show.

At some point I was even convinced that having two layers named in a similar way was enough to trigger this, but I struggled to replicate and after several hours of troubleshooting I'm fed with this as it seems a very finicky problem.

To note that this only affect the pop up dialog, as the export drodown shows as expected in the attribute table, even in the above cases.

See above.

### Versions

QGIS LTR on Desktop and Server. Latest version of LMWC, Limap plugin and Lizmap Server plugin.

### Check Lizmap plugin

3.28

### Operating system

Ubuntu 22.04, desktop and server

### Browsers

Firefox, Chrome

### Browsers version

Latest

### Relevant log output

_No response_

nboisteault:
Fix #3660 but just for next Lizmap 3.7. @mdouchin @rldhont Any other idea to fix it for previous versions?

Funded by 3Liz

### DIFF
--- a/assets/src/legacy/switcher-layers-actions.js
+++ b/assets/src/legacy/switcher-layers-actions.js
@@ -161,26 +161,12 @@ var lizLayerActionButtons = function() {
                 && featureTypes.length != 0 ) {
                 var exportFormats = lizMap.getVectorLayerResultFormat();
                 var options = '';
-                for ( var i=0, len=exportFormats.length; i<len; i++ ) {
-                    var format = exportFormats[i];
+                for ( const format of exportFormats ) {
                     options += '<option value="'+format+'">'+format+'</option>';
                 }
                 // Export layer
                 // Only if layer is in attribute table
-                var showExport = false;
-                if( options != '' ) {
-                    for (const featureType of featureTypes) {
-                        var typeName = featureType.getElementsByTagName('Name')[0].textContent;
-                        if ( typeName == aName ) {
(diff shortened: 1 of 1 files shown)


assets/src/legacy/switcher-layers-actions.js
```diff
@@ -161,26 +161,12 @@ var lizLayerActionButtons = function() {
...
-                            showExport = true;
-                            continue;
-                        } else if (typeName == aName.split(' ').join('_') ) {
-                            showExport = true;
-                            continue;
-                        }
-                    }
-                }
-                if( showExport ) {
+                if( options != '' && lizMap.mainLizmap.initialConfig.attributeLayers.layerNames.includes(aName) ) {
```
rldhont, line 169:
```suggestion
                if( options != '' && layerConfig.typename != undefined) {
```

rldhont (changes requested):

PR #3856, opened 2023-08-24
New permalink not OL2 dependant

nboisteault:
New permalink as we move from OL2

Funded by 3Liz


nboisteault:
TODO: 
- handle filter
- remove old OL2 code

PR #3858, opened 2023-08-25
Editing - Add a new public method to check the form validity

mdouchin:
This modification is a first step to improve the overall editing workflow, especially when creating child features for a not-yet existing parent object.

The future workflow could be

* open the creation form for a new parent object
* allow opening a creation form for 1 to many children
* validate the form data for each child to create, without having to save first the parent object (and the children). Do it also when modifying the child form data (we could keep the N forms in the DOM)
* Display the table of children "to be added or modified", with a look & feel like

|                  UID                 |    Feature ID   | Label (automatically generated based on QGIS display expression) | Edit   | Remove |
|:------------------------------------:|:---------------:|------------------------------------------------------------------|--------|--------|
| 96944c88-b947-4b69-a538-d9605193bd57 | to be generated | Lampadaire EEOOFG tungstène                                      | [edit] | [x]    |
| 96944c88-b947-4b69-a538-d96051923456 | 38              | Lampadaire AABBCC diode                                          | [edit] | [x]    |

* save the children form data in a single hidden text field, called for example `children_data_layer_XXX` which will store a table of JSON encoded form data
* when the user will Save the parent, in the PHP backend:
  * validate the new parent object
  * re-validate each child
  * save the parent, then save the children in batch. If a database error occurred, rollback all.

Funded by Commune d'Avignon

### DIFF
--- a/lizmap/modules/lizmap/controllers/edition.classic.php
+++ b/lizmap/modules/lizmap/controllers/edition.classic.php
@@ -319,6 +319,22 @@ class editionCtrl extends jController
     }
 
     /**
+     * Generate the jForms form identifier
+     * depending on the action (creation or modification)
+     * and the source layer and optional existing feature.
+     *
+     * @return string the form ID to be used when creating/getting the form
+     */
+    protected function formId()
+    {
+        if ($this->featureId) {
+            return $this->layerId.'||'.$this->featureId;
+        }
+
+        return $this->layerId.'||____new__feature___';
+    }
+
+    /**
      * @param jFormsBase $form
      * @param mixed      $forCreation
      */
@@ -398,9 +414,11 @@ class editionCtrl extends jController
             return $this->serviceAnswer();
         }
 
-        jForms::destroy('view~edition', $this->featureId);
+        // Destroy existing form
+        jForms::destroy('view~edition', $this->formId());
+
         // Create form instance
-        $form = jForms::create('view~edition', $this->featureId);
+        $form = jForms::create('view~edition', $this->formId());
 
         if (!$this->initializeForm($form)) {
             return $this->serviceAnswer();
@@ -455,7 +473,7 @@ class editionCtrl extends jController
         }
 
         // Create form instance
-        $form = jForms::create('view~edition', $this->featureId);
+        $form = jForms::create('view~edition', $this->formId());
         if (!$form) {
             jMessage::add(jLocale::get('view~edition.message.error.form.get'), 'formNotDefined');
 
@@ -514,7 +532,7 @@ class editionCtrl extends jController
         }
 
         // Get the form instance
-        $form = jForms::get('view~edition', $this->featureId);
+        $form = jForms::get('view~edition', $this->formId());
         if (!$form) {
             jMessage::add(jLocale::get('view~edition.message.error.form.get'), 'formNotDefined');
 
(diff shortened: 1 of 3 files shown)


lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -398,9 +398,9 @@ public function createFeature()
             return $this->serviceAnswer();
         }
 
-        jForms::destroy('view~edition', $this->featureId);
+        jForms::destroy('view~edition', $this->layerId.'||'.$this->featureId);
```
rldhont, line 401:
```suggestion
        jForms::destroy('view~edition', $this->formId());
```

Après avoir défini le méthode `formId`
```php
protected function formId() {
    if ($this->featureId) {
        return $this->layerId.'||'.'||'.$this->featureId;
    }
    return $this->layerId; // ou $this->layerId.'||____new__feature___';
}
```

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -398,9 +398,9 @@ public function createFeature()
...
         // Create form instance
-        $form = jForms::create('view~edition', $this->featureId);
+        $form = jForms::create('view~edition', $this->layerId.'||'.$this->featureId);
```
rldhont, line 403:
```suggestion
        $form = jForms::create('view~edition', $this->formId());
```

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -455,7 +455,7 @@ public function modifyFeature()
         }
 
         // Create form instance
-        $form = jForms::create('view~edition', $this->featureId);
+        $form = jForms::create('view~edition', $this->layerId.'||'.$this->featureId);
```
rldhont, line 458:
```suggestion
        $form = jForms::create('view~edition', $this->formId());
```

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -514,7 +514,7 @@ public function editFeature()
         }
 
         // Get the form instance
-        $form = jForms::get('view~edition', $this->featureId);
+        $form = jForms::get('view~edition', $this->layerId.'||'.$this->featureId);
```
rldhont, line 517:
```suggestion
        $form = jForms::get('view~edition', $this->formId());
```

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -602,6 +602,110 @@ public function editFeature()
...
+        // Get repository, project data and do some right checking
+        $save = true;
+        if (!$this->getEditionParameters($save)) {
+            $rep->data['message'] = 'An error occurred while initializing the context (repository, project, etc.)';
+
+            return $rep;
+        }
+
+        // Get the form instance
+        $form = jForms::get('view~edition', $this->layerId.'||'.$this->featureId);
```
rldhont, line 642:
```suggestion
        $form = jForms::get('view~edition', $this->formId());
```

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -621,7 +725,7 @@ public function saveFeature()
         }
 
         // Get the form instance
-        $form = jForms::get('view~edition', $this->featureId);
+        $form = jForms::get('view~edition', $this->layerId.'||'.$this->featureId);
```
rldhont, line 728:
```suggestion
        $form = jForms::get('view~edition', $this->formId());
```

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -709,7 +813,7 @@ public function saveFeature()
 
         // Redirect to the edition form or to the validate message
         $next_action = $form->getData('liz_future_action');
-        jForms::destroy('view~edition', $this->featureId);
+        jForms::destroy('view~edition', $this->layerId.'||'.$this->featureId);
```
rldhont, line 816:
```suggestion
        jForms::destroy('view~edition', $this->formId());
```

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -793,7 +897,7 @@ public function closeFeature()
         }
 
         // Destroy the form
-        jForms::destroy('view~edition', $this->featureId);
+        jForms::destroy('view~edition', $this->layerId.'||'.$this->featureId);
```
rldhont, line 900:
```suggestion
        jForms::destroy('view~edition', $this->formId());
```

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -849,7 +953,7 @@ public function deleteFeature()
         }
 
         // Create form instance to get uploads file
-        $form = jForms::create('view~edition', $this->featureId);
+        $form = jForms::create('view~edition', $this->layerId.'||'.$this->featureId);
```
rldhont, line 956:
```suggestion
        $form = jForms::create('view~edition', $this->formId());
```

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -934,7 +1038,7 @@ public function saveNewFeature()
         }
 
         // Get the form instance
-        $form = jForms::create('view~edition', '____new__feature___');
+        $form = jForms::create('view~edition', $this->layerId.'||____new__feature___');
```
rldhont, line 1041:
```suggestion
        $form = jForms::create('view~edition', $this->formId());
```

lizmap/modules/lizmap/controllers/edition.classic.php
```diff
@@ -1003,7 +1107,7 @@ public function saveNewFeature()
         }
         $pkvals = $qgisForm->saveToDb($feature);
 
-        jForms::destroy('view~edition', '____new__feature___');
+        jForms::destroy('view~edition', $this->layerId.'||____new__feature___');
```
rldhont, line 1110:
```suggestion
        jForms::destroy('view~edition', $this->formId());
```

rldhont (changes requested):

lizmap/modules/lizmap/controllers/edition.classic.php
mdouchin: Good catch. I also thought about creating a unique form ID for each "new feature". So that each form would be unique. We would also need to have different form node ID property for each form.

mdouchin: @laurentj In the context of object creation (parent and optional children) what about using a parameter UID filled by the JS code when calling the `edition` methods and use it to generate the form ID ? I would use a UUID such as `96944c88-b947-4b69-a538-d9605193bd57` (or a simpler hash).

The JS code would then keep the form's uniques IDs and be able to manage validations for each form (and hide/show the forms, for example if we show a table of children data "to be added" with buttons to modify each child form data.

We could have a IRL conversation about it ;)

laurentj: Unfortunately, in the case of there multiple forms of the same type in the same HTML page, even if they have a different form ID, the HTML and JS code cannot work correctly. All HTML field have the same html id across all forms, so a `document.getElementById()` won't work. 

I must study how we can solve this issue. A solution is to change how the html id are built, but I'm afraid that it will brake all existing custom JS scripts that rely on these html id... And this change cannot be made in the current stable version of Jelix, so it will be made into Jelix 1.9, so for Lizmap 3.8 or 4.0..

I'll try to find another solution...

rldhont (approved):

Issue #3799, opened 2023-07-21
[Bug]: Dataviz: unhandled 404 error on function getPlot (dataviz.js)


mind84:
### What is the bug?

Hi!
We're working with @tethysgeco on Lizmap 3.6 with a simple test project set up to display some scattered charts on parent layer popup.

If the parent layer has no child data associated with it, I would expect to see a blank graph like in Lizmap 3.5. Instead the chart section in the popup remains stuck on the loading bar.
![stuck_bar](https://github.com/3liz/lizmap-web-client/assets/20710470/ece52f21-0194-4081-8a6c-464b19d54a9a)

Since the server side logic of dataviz module has been improved from the 3.5 version, I think that the problem is related to the function [getPlot](https://github.com/3liz/lizmap-web-client/blob/994860713f84008daaa2a0de813d7f9defd6b1ac/lizmap/www/assets/js/dataviz/dataviz.js#L117) in dataviz.js, which now cant'handle the 404 response that the server send to client when no child data has been found.

@u-cav

Thanks

### Steps to reproduce the issue

Here's the test project made by @tethysgeco [test_lizmap.zip](https://github.com/3liz/lizmap-web-client/files/12129793/test_lizmap.zip) to reproduce the issue.

### Versions

**Versions :**

* Lizmap Web Client : 3.6.4
* Lizmap plugin : 3.14.1
* QGIS Desktop : 3.28.7
* QGIS Server : 3.28.5
* Py-QGIS-Server : 1.8.8
* QGIS Server plugin DataPlotly : 4.0.0
* QGIS Server plugin atlasprint : 3.3.2
* QGIS Server plugin lizmap_server : 2.7.1
* QGIS Server plugin wfsOutputExtension : 1.7.1

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Debian

### Browsers

Firefox, Chrome

### Browsers version

latest

Gustry:
Do you have any error in the web browser console ? (opening before clicking on the dataviz panel), F12

u-cav:
We are getting a 404 on the plot with missing data, with Lizmap 3.6 (the same error is also in the console):
![image](https://github.com/3liz/lizmap-web-client/assets/4831528/b00056ad-d6d5-4c13-87a6-156188723954)


The same project on 3.5 gets a 200 empty response:
[...]

PR #3861, opened 2023-08-29
[Backport release_3_6] Dataviz - Do not display child plot in Popup when there is no data

mind84:
Backport to release 3_6 https://github.com/3liz/lizmap-web-client/pull/3854

When displaying the popup of an object containing child plots, if the layer has no child data associated with a specific plot, the loading bar must not be displayed but the plot should be hidden.

Ticket : fix https://github.com/3liz/lizmap-web-client/issues/3799

Funded by Faunalia (https://www.faunalia.eu/)

### DIFF
--- a/lizmap/www/assets/js/dataviz/dataviz.js
+++ b/lizmap/www/assets/js/dataviz/dataviz.js
@@ -114,8 +114,16 @@ var lizDataviz = function() {
         }
     }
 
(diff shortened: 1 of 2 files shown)


Gustry:
Thanks for fixing the tests and doing the backport

rldhont (approved):

mdouchin (approved):
Thanks a lot

PR #3868, opened 2023-09-07
Use `wmsName` for WMS requests

### LINKED ISSUE #3675: [Bug]: Filtered features disappear from map
### What is the bug?

When activating the filter on selected features/records, the features disappear from the map.
This happens from both the attribute table and the "select" dialog.

image

image

image

### Versions

QGIS LTR on Desktop and Server, latest version of LMWC, LM plugin and LM server plugin.

### Check Lizmap plugin

3.28

### Operating system

Ubuntu 22.04, desktop and server

### Browsers

Firefox, Chrome

### Browsers version

Latest

### Relevant log output

_No response_

nboisteault:
Fix #3675

Funded by 3Liz

### DIFF
--- a/assets/src/legacy/attributeTable.js
+++ b/assets/src/legacy/attributeTable.js
@@ -467,12 +467,12 @@ var lizAttributeTable = function() {
                 // Unselect button
                 html+= '    <button class="btn-unselect-attributeTable btn btn-mini' + selClass + '" value="' + cleanName + '" title="'+lizDict['attributeLayers.toolbar.btn.data.unselect.title']+'"><i class="icon-star-empty"></i></button>';
 
-                // Move selected to Top button
+                // 'Move selected to top' button
                 html+= '    <button class="btn-moveselectedtotop-attributeTable btn btn-mini' + selClass + '" value="' + cleanName + '" title="'+lizDict['attributeLayers.toolbar.btn.data.moveselectedtotop.title']+'"><i class="icon-arrow-up"></i></button>';
 
                 // Filter button : only if no filter applied at startup
                 if( !startupFilter
-                    && ( !lizMap.lizmapLayerFilterActive || lizMap.lizmapLayerFilterActive == cleanName )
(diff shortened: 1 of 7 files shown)


tests/qgis-projects/tests/attribute_table.qgs.cfg
```diff
@@ -64,26 +69,24 @@
...
             "imageFormat": "image/png",
             "cached": "False",
-            "serverFrame": null,
             "clientCacheExpiration": 300
         },
-        "quartiers": {
+        "Les quartiers \u00e0 Montpellier": {
             "id": "quartiers_5fe55662_2cbf_48f4_a505_498c61fe978c",
-            "name": "quartiers",
+            "name": "Les quartiers \u00e0 Montpellier",
```
Gustry, line 89: Is-it an encoding issue on the file ?
Should I do something ?

Gustry (commented):
Just a question

assets/src/legacy/attributeTable.js
```diff
@@ -2207,6 +2207,7 @@ var lizAttributeTable = function() {
                 var layerConfig = config.layers[typeName];
                 layerConfig['features'] = {};
                 var foundFeatures = {};
+                const wmsName = lizMap.mainLizmap.state.layersAndGroupsCollection.getLayerByName(layerConfig.name).wmsName;
```
rldhont, line 2210:
```suggestion
                const wmsName = layerConfig?.shortname || layerConfig.name;
```
To keep compatibility with legacy code and to be sure to backport automatically.

rldhont (changes requested):

tests/qgis-projects/tests/attribute_table.qgs.cfg
nboisteault: I don't know. It is cleanly represented in the developer tool. I guess it is to avoid some issue w/ JSON format.
@rldhont an idea?

image

nboisteault:
CI fails because we have this issue: https://github.com/cypress-io/cypress/issues/19122
But I've tested locally and it is ok so I merge.

PR #3879, opened 2023-09-18
Fix nullable bool value map

nboisteault:
Funded by 3Liz


lizmap/modules/lizmap/lib/Form/QgisFormControl.php
```diff
@@ -514,14 +514,33 @@ protected function fillControlDatasource()
...
+                        if (!$this->required
+                            && isset($valueMap[self::QGIS_NULL_VALUE])
+                        ) {
+                            unset($valueMap[self::QGIS_NULL_VALUE]);
+                        }
+                        // transform values from QGIS ValueMap to Postgres Boolean
+                        $booleanValues = array();
+                        foreach ($valueMap as $v => $label) {
+                            $strV = strtolower($v);
+                            if ($strV === 'true' || $strV === 't'
```
rldhont, line 531:
```suggestion
                            $strV = strtolower($v);
                            if ($strV === self::QGIS_NULL_VALUE) {
                                $booleanValues[$strV] = $label;
                            } elseif ($strV === 'true' || $strV === 't'
```

Gustry:
Very side question, but it does it fix the custom label for the no value ? (which can be set in QGIS, it's already done in the test project for instance)

If you are OK, it's fine for me about the fix it self.

nboisteault:
> Very side question, but it does it fix the custom label for the no value ? (which can be set in QGIS, it's already done in the test project for instance)
> 
> If you are OK, it's fine for me about the fix it self.

The custom label is visible in the popup. Should it also be visible in list in edition mode?

Gustry:
> Should it also be visible in list in edition mode?

I'm talking about edition only, not the popup.
In QGIS, it works for edition, to set a custom value for "True", "False" and NULL.

nboisteault:
In your example `One way` and `Two way` are visible, not `I don't know` for NULL value.

PR #3895, opened 2023-09-28
[PHP] Get cached Polygon filter for project, layer, login and editing context

rldhont:
The Polygon filter is provided by QGIS Server. The request could take a long time so the response as to be cached.

Funded by [Conseil Départemental du Calvados](https://www.calvados.fr/accueil.html)

### DIFF
--- a/lizmap/modules/lizmap/classes/qgisVectorLayer.class.php
+++ b/lizmap/modules/lizmap/classes/qgisVectorLayer.class.php
@@ -1416,10 +1416,11 @@ class qgisVectorLayer extends qgisMapLayer
      * @param bool $editing_context If we are in a editing context or no. Default false
      * @param int  $ttl             Cache TTL in seconds. Default 60. Use -1 to deactivate the cache.
      * @param bool $get_expression  If we need the expression and not the SQL
+     * @param bool $use_cache       If we need to not use cache
      *
      * @return array associative array containing the keys 'expression' and 'polygon'
      */
-    protected function requestPolygonFilter($editing_context = false, $ttl = 60, $get_expression = false)
+    protected function requestPolygonFilter($editing_context = false, $ttl = 60, $get_expression = false, $use_cache = true)
     {
         // No filter response
         $no_filter_array = array(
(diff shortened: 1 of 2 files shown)


lizmap/modules/lizmap/classes/qgisVectorLayer.class.php
```diff
@@ -1522,8 +1534,7 @@ protected function requestPolygonFilter($editing_context = false, $ttl = 60, $ge
                 );
 
                 if ($use_cache) {
-                    // Todo: use the project cache tools
-                    // jCache::set($cache_key, json_encode($polygon_filter_data), $ttl);
+                    $cached = $this->project->getCacheHandler()->setProjectRelatedDataCache($cache_key, $polygon_filter_data, 3600);
```
nworr, line 1537: cacheHandler may be unavailable ?

catch Exception  ?

 ```
       try {
            $cached = $this->project->getCacheHandler()->setProjectRelatedDataCache($cache_key, $polygon_filter_data, 3600);
        } catch (\Exception $e) {
            // if qgisprojects profile does not exist, or if there is an
            // other error about the cache, let's log it
            \jLog::logEx($e, 'error');
        }
```

or set $use_cache=false above (when getProjectRelatedDataCache() throws Exception ?

**Review by nworr — CHANGES_REQUESTED** [review]

**Comment by Gustry** [comment]
Unrelated test failure about metadata. Fixed in master, you can rebase if you want or skip this failing test

File: lizmap/modules/lizmap/classes/qgisVectorLayer.class.php (hunk 1 above)
Comment (rldhont) [reply lizmap/modules/lizmap/classes/qgisVectorLayer.class.php:1540]: Keep it before with `getProjectRelatedDataCache`

**Review by nworr — APPROVED** [review]

**Comment by nworr** [comment]
jcache doesn't provide a cache dependencies behaviour, so the TTL must be explicitly set for projetRelatedData , there's no (simple) way to make a smarter ttl.

When the projet cache will be flushed the items stored with projetRelatedDataCache will still be there.

**Comment by rldhont** [comment]
@nworr the `projectCache` is enough smart for me
```php
   /**
     * Read some data from the project cache, related to the project.
     *
     * It checks if the cache is belong with the current project cache.
     *
     * @param string $key
     *
     * @return false|mixed false if there is no cache, else the data
     */
    public function getProjectRelatedDataCache($key)
    {
        $data = false;

        try {
            $cacheKey = $this->fileKey.'.'.$key;
            $cacheContent = $this->appContext->getCache($cacheKey, $this->profile);
            // check if the cache correspond to the current project
            if ($cacheContent === false
                || $cacheContent['qgsmtime'] < $this->qgsMtime
                || $cacheContent['qgscfgmtime'] < $this->qgsCfgMtime
                || $cacheContent['format_version'] != self::CACHE_FORMAT_VERSION
            ) {
                return false;
            }
            $data = $cacheContent['data'];
        } catch (\Exception $e) {
            // if qgisprojects profile does not exist, or if there is an
            // other error about the cache, let's log it
            $this->appContext->logException($e, 'error');
        }

        return $data;
    }
```

PR #3900, opened 2023-09-29
e2e: migrate Cypress tests to playwright

nboisteault:
Funded by 3Liz


Gustry:
I guess there is a wrong rebase @nboisteault ?

PR #3903, opened 2023-09-29
[Bugfix] The qgisVectorLayer::editableFeatures did not check edition capabilities

rldhont:
The qgisVectorLayer::editableFeatures was able to return every feature of the layer when the layer is not editable.

Funded by [Conseil Départemental du Calvados](https://www.calvados.fr)


nboisteault:
@rldhont could you create/change a project and create a test?

rldhont:
If the e2e tests still be ok, my changes does not break anything.

PR #3927, opened 2023-10-13
Landing page: possibility to add contents at the bottom of the page

laurentj:
In the administration, we can now edit content which will be inserted at the bottom of the landing page.

Improve also the UX on this administration page.

Funded by 3Liz


laurentj:
I don't know if the possibility to add template tags was a feature or not. If this is a case, we should set the template as "untrusted".

Gustry (commented):
LGTM, good idea to be able to add a footer

Note, there are two errors from phpstan run

lizmap/modules/admin/locales/en_US/admin.UTF-8.properties
```diff
@@ -362,3 +362,10 @@ project.list.column.show.line.hidden.columns=Show/Hide the hidden columns
 project.modal.title=Table legend
 project.modal.button.close=Close
 project.list.no.hidden.column.content=No extra content for this project in the following hidden table columns
+
+
+landingPageContent.top=Content at the top of the landing page
+landingPageContent.bottom=Content at the bottom of the landing page
+landingPageContent.error.save=Error during the save of the landing page contents
+landingPageContent.error.submit=An error occurred during the save, your session has been terminated, please retry.
```
mdouchin, line 370:
```suggestion
landingPageContent.error.save=Error while saving the landing page contents
landingPageContent.error.submit=An error occurred while saving the content. Your session has been terminated, please retry.
```

mdouchin (changes requested):
Nice addition !
I just proposed a small syntax modification

laurentj:
Locales have been updated. 

Errors of phpstan is about a bad comment into jelix. I don't want to release a new version of jelix just for a comment. So let's ignore PhpStan for this PR.

gioman:
Nice addition. Would be possible to enable the html view in the editor?

PR #3928, opened 2023-10-16
Permalink finish

nboisteault:
Funded by 3Liz


Gustry:
Is-it possible to have a look to ticket gitlab LWC 224 ?

Issue #3693, opened 2023-05-17
[Bug]: Snapping do not works on vertexes (works on segments)


gioman:
### What is the bug?

Snapping do not works on vertexes (works on segments)

### Steps to reproduce the issue

Configure your project with a PostGIS layer and set it as editable in LM plugin, vertexes only. > not working

Se snappjng to segments > work (but still no snapping on vertexes)

### Versions

LMWC 3.6.3 Latest version of LM plugin and LM server plugin. 

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### QGIS server version, only if the section above doesn't mention the QGIS Server version

3.28

### Operating system

Ubuntu 22.04, desktop and server

### Browsers

Firefox, Chrome

### Browsers version

Latest

gioman:
Can anyone confirm?

gioman:
Bumping my tickets before they go "quite",

josemvm:
hi @gioman on LWC 3.6.5 everything works fine here

josemvm:
qgis server and desktop 3.22.16

gioman:
> qgis server and desktop 3.22.16

@josemvm thanks for the feedback. I'm on 3.28 and cannot downgrade to test.

gioman:
I tried on a different machine still with QGIS Desktop and Server 3.28(.11) and with LMWC 3.6.6 and still can't see snapping to vertexes work. I tried on several layers, some complex. some very simple and yet I can only see snap on segments work.

Gustry:
What kind of geometry are you editing ?
What kind of geometry are you snapping to ?

Which checkbox did you use for snapping ? Did you use "Intersection" ?

gioman:
@Gustry 

> What kind of geometry are you editing ? 

polygons


> What kind of geometry are you snapping to ?

I'm trying to snap to the geometries of the layer I'm editing, not the ones of other layers 

> Which checkbox did you use for snapping ? 

only vertices > does not work
check also segments > works on segments, still not on vertices


>Did you use "Intersection" ?

no

Gustry:
In PR #3929, it works well with the checkbox "Intersection".
I'm not able to check exactly what this checkbox is doing in LWC source code. But it's worth to have a look into label, tooltip etc. (`Snapping on inner vertices` like in QGIS ?)

gioman:
[...]

PR #3929, opened 2023-10-18
Tests - Add more data for testing snapping

Gustry:
#3693


rldhont:
@Gustry can you rebase ?

Issue #3930, opened 2023-10-18
Advanced form, filter: expression like "intersects($geometry, centroid(@current_geometry))" does not work


gioman:
### What is the bug?

Following the example here https://docs.lizmap.com/current/en/publish/layer_properties/attributes_form.html for a value relation/filter, and when the layer being edited is a point layer the example provided works.

When trying with a polygon and its centroid it doesn't work, the combobox remains unpopulated (works in QGIS):

`intersects($geometry, centroid(@current_geometry))`

### Steps to reproduce the issue

Follow the example for a value relation/filter as described here https://docs.lizmap.com/current/en/publish/layer_properties/attributes_form.html but use a polygon layer instead of a point one, and this expression

`intersects($geometry, centroid(@current_geometry))`

### Versions

LMWC 3.6.6

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### QGIS server version, only if the section above doesn't mention the QGIS Server version

3.28..1

### Operating system

Ubuntu 22.04

### Browsers

Firefox, Chrome

### Browsers version

Latest updates

Gustry:
It works well on the test project `form_edition_value_relation_field.qgs`

```
2023-10-18 13:52:19,680 DEBUG   [57]    Qgis: Lizmap: Fetching /srv/projects/tests/form_edition_value_relation_field.qgs.cfg cfg file with last modified timestamp : 1697636723.2394717
2023-10-18 13:52:19,680 DEBUG   [57]    Qgis: Lizmap: No acl defined in Lizmap config
2023-10-18 13:52:19,680 DEBUG   [57]    Qgis: Server: FIELDS:code,label
2023-10-18 13:52:19,680 DEBUG   [57]    Qgis: Server: FILTER:intersects($geometry, centroid(@current_geometry))
2023-10-18 13:52:19,680 DEBUG   [57]    Qgis: Server: FORM_FEATURE:{"type":"Feature","geometry":{"type":"Polygon","coordinates":[[[0.305070750024655,47.89601575850115],[1.5291213472474636,47.53621582892768],[0.9111540554456543,46.776610723827396],[0.305070750024655,47.89601575850115]]]},"properties":[]}
```

**Versions :**

* Lizmap Web Client : 3.7.0-beta.2
* Lizmap plugin : master
* QGIS Desktop : 3.28.11
[...]

PR #3931, opened 2023-10-18
Tests - polygon layer for Value Relation with centroid expression

Gustry:
Related to #3930

In this dataset, it works @gioman


rldhont:
@Gustry can you rebase ?

PR #3934, opened 2023-10-20
Finish text draw

nboisteault:


Gustry:
Thanks @nboisteault it's indeed better now
Let's merge it and we check the related ticket.

PR #3936, opened 2023-10-20
Fix phpdoc / update phpstan config to ignore Dao related method/properties

nworr:
about 30 errors dropped
 

Funded by 3liz


Gustry (approved):
LGTM because CI is green and fix the CI in master

rldhont (approved):

PR #3937, opened 2023-10-20
[Feature] Use remote WebDAV server to store and retreive files

mind84:
Provide on LWC the functionality of the `Attachment Widget`'s _WebDav Storage_ property, according to the widget [documentation](https://docs.qgis.org/3.28/en/docs/user_manual/working_with_vector/attribute_table.html?highlight=webdav#configuring-an-external-storage) 

Funded by Faunalia


mdouchin:
Thanks a lot for this contribution. We will take time to audit the code. Not sure if it can land in 3.7, as we would like to release in a very short term

mind84:
> Thanks a lot for this contribution. We will take time to audit the code. Not sure if it can land in 3.7, as we would like to release in a very short term

Thanks to you, this is an early stage development and contains only the basic structure of the desired feature. It can be a start point

I'm available for any questions

Cheers!

pcav:
Any plans to push this in the next LM release? We can patch it, but it would be nice (and I believe useful for others) to have in stable version.

lizmap/modules/lizmap/lib/Request/RemoteStorageRequest.php
```diff
@@ -0,0 +1,361 @@
...
+     * @return null|string
+     */
+    public static function getRemoteUrl($storageUrl, $filename = null)
+    {
+        $storeLen = strlen($storageUrl);
+        $fileNameLen = strlen(self::$fileNameExpression);
+        if ($fileNameLen < $storeLen && substr_compare($storageUrl, self::$fileNameExpression, $storeLen - $fileNameLen, $fileNameLen) === 0) {
+            if ($filename) {
+                // TODO @selected_file_path property is not evaluated, for now replace the expression with the file name
+                return str_replace('file_name(@selected_file_path)', "'".$filename."'", $storageUrl);
```
nworr, line 353: `self::$fileNameExpression` is made for that, no ?
```suggestion
                return str_replace(self::$fileNameExpression, "'".$filename."'", $storageUrl);
```

lizmap/modules/lizmap/lib/Request/RemoteStorageRequest.php
```diff
@@ -0,0 +1,361 @@
...
+            }
+
+            return str_replace('file_name(@selected_file_path)', "''", $storageUrl);
```
nworr, line 356: idem 
```suggestion
            return str_replace(self::$fileNameExpression, "''", $storageUrl);
```

lizmap/modules/lizmap/lib/Request/RemoteStorageRequest.php
```diff
@@ -0,0 +1,361 @@
...
+                $opt = array();
+                $headers = array();
+                $headers['User-Agent'] = 'lizmap-user-agent';
+
+                if (array_key_exists('user', $profile) && array_key_exists('password', $profile)) {
+                    $opt['auth'] = array();
+                    array_push($opt['auth'], $profile['user']);
+                    array_push($opt['auth'], $profile['password']);
+                }
+                $opt['headers'] = $headers;
```
nworr, line 67: auth and header are always set, and guzzle allow to add request option in the constructor, you can create a dedicated function (or a dedicated class) like 
```php
protected static function buildClient($profile) {
        $opt = array();
        $headers = array();
        $headers['User-Agent'] = 'lizmap-user-agent';

        if (array_key_exists('user', $profile) && array_key_exists('password', $profile)) {
             $opt['auth'] = array();
             array_push($opt['auth'], $profile['user']);
             array_push($opt['auth'], $profile['password']);
        }
        $opt['headers'] = $headers;
        return new new Client($opt);
}
```

and call it in  `uploadToWebDAVStorage `,`deleteFromWebDAVStorage`, ...

lizmap/modules/lizmap/lib/Request/RemoteStorageRequest.php
```diff
@@ -0,0 +1,361 @@
...
+                    $message = 'Error on file upload '.$e->getMessage();
+                }
+            } else {
+                $message = 'Invalid path '.$storageUrl;
+            }
+        } else {
+            $message = 'WebDAV configuration not found';
+        }
+        if ($message) {
+            \jLog::log($message, 'error');
```
nworr, line 94: you can use `$this->appContext->logMessage`

lizmap/modules/view/controllers/media.classic.php
```diff
@@ -119,86 +120,121 @@ public function getMedia()
...
+
+                $tempFileUrl = RemoteStorageRequest::getRemoteFile($url, $fileName);
+                if (!is_file($tempFileUrl)) {
+                    $content = 'No media file in the specified path: '.$path;
+
+                    return $this->error404($content);
+                }
+                $isWebDavResource = true;
+                $finalPath = $tempFileUrl;
+            }
```
nworr, line 146: maybe return an error if no profile ?
```suggestion
            } else {
                 return $this->error404('Unable to get media');
            }
```

nworr (commented):
Look good to me, some small code improvements can be done. (see suggestion)

Gustry:
Regarding the last comment on the PR, in the timeline

> Thanks to you, this is an early stage development and contains only the basic structure of the desired feature. It can be a start point

So I think now, it's more than "early stage development" ? Can you just confirm that it's ready for a production use now ? Or any known bullet list to check ?

I will checkout the PR, thanks for this work.

mind84:
> So I think now, it's more than "early stage development" ? Can you just confirm that it's ready for a production use now ? Or any known bullet list to check ?
> 
> I will checkout the PR, thanks for this work.

There are still a couple of things that need to be done:

- **hide webdav url for getMedia** 
if the layer is a "postgre layer", Lizmap collects the information directly from the db so it is possible via php to modify the returned url and prepare it for viewing by the client. Otherwise, the feature data is collected with the `getfeatureQgis` method. If I'm not mistaken, this function forwards the request to `qgis-lizmap-server-plugin` so I left out this part thinking that the best way to hide the URL is to modify the server plugin

- **displaying url in `Popup`**
basically the same for the previous point, HTML of the popup is generated server-side and I left out this part for the same reasons

Do you think that resolving this two points by intercepting the `qgis-lizmap-server-plugin` response via php could be fine? 

Initially I had discarded this option because I thought it was more correct to make the changes on the server plugin.

Gustry (changes requested):
-- Note, I started the review and screenshots before your previous comment about known bottlenecks. I would be nice to write explicitly what is still a work in progress, before we start a review. I didn't edit my comment below, that I started before your comment ---

I just give it a try : 

* Minor note, maybe in the your new test tables, I would add some data with a webdav data in the attribute table ? I would make it easier for testing ? (for attribute table and popup tests) I have noticed you always start playwright test with addition of new features ? Just an idea
* The popup doesn't have a proper link to the new GetMedia, it's leading to `http://webdav/ksnip_20231017-111339.png` which obvously giving an error in web-browser when clicking on the link in the popup

it's the first think I tried, create a new feature with a webdav picture, open on it's popup... This is a blocker I think.

image

The data from WFS has the raw URL, I think all data from LWC should be a proxy to hide the webdav server
image

If I understood correctly, data can only be stored in the root directory of the web dav server ? Because of the QGIS expression used for now ?

Sorry, I might other tests coming

mind84:
> -- Note, I started the review and screenshots before your previous comment about known bottlenecks. I would be nice to write explicitly what is still a work in progress, before we start a review. I didn't edit my comment below, that I started before your comment ---
> 
Right, sorry for the lack of information

> I just give it a try :
> 
>     * Minor note, maybe in the your new test tables, I would add some data with a webdav data in the attribute table ? I would make it easier for testing ? (for attribute table and popup tests) I have noticed you always start playwright test with addition of new features ? Just an idea

Yes, but in this case during application building some files should be moved in webdav folder right? In this case can you point me the files I should update?
> 
>     * The popup doesn't have a proper link to the new GetMedia, it's leading to `http://webdav/ksnip_20231017-111339.png` which obvously giving an error in web-browser when clicking on the link in the popup
> 
> 
> it's the first think I tried, create a new feature with a webdav picture, open on it's popup... This is a blocker I think.
> 
Yes, as pointed out, this is a work in progress to discuss together

> 
> The data from WFS has the raw URL, I think all data from LWC should be a proxy to hide the webdav server 
> 

Yes, as pointed out, this is true for non postgre layers. Work in progress same as the previous point

> If I understood correctly, data can only be stored in the root directory of the web dav server ? Because of the QGIS expression used for now ?
> 

No, you can store data also in subfolders. 
If you check out the test project `tests/qgis-projects/tests/form_upload_webdav.qgs` the layer "_geom" stores the files in the root directory, the other layer stores the files in the `remoteData` subfolder.
The `storageUrl` must include the expression `file_name(@selected_file_path)`

> Sorry, I might other tests coming

Thank you

Gustry:
> Yes, but in this case during application building some files should be moved in webdav folder right? In this case can you point me the files I should update?

Why not committing just a text file and a very small picture in `./qgis-projects/webdav/test` ? (just add it in the gitignore to be able to commit) and fill the data in the SQL table like you did with other fields.

> Yes, as pointed out, this is a work in progress to discuss together

I think, popup is blocker (whatever the mode : automatic, QGIS HTML maptip, QGIS form) to have a proper replacement, the popup feature is a common feature, after creation of a new feature in the table. Users are expecting to see the data they just added.

> Yes, as pointed out, this is true for non postgre layers.

I will let my colleague answer, I don't know exactly the path in the PHP source code, but in PHP, there should be a place where data is replaced. (not JS, so it work in QGIS with WFS stream, and making a replacement in JS side is not secure) for all kind of requests (WFS attribute table, WMS GetFeatureInfo for popup...) whatever the backend.

Indeed with PG layers, lizmap indeed tries **if possible** to make a straight SQL query, but if for any reason it fails, it will still use the WFS service from QGIS server (because we know for sure QGIS server can access the data). And QGIS server is used for all other backend indeed. And Lizmap QGIS server plugin is not involved in this part. (only for the GetFeatureInfo, to have the "form" popup about Drag&Drop design).

Side note, for playwright, I would avoid pixel comparaison, even if you are using the same picture as upload, so it should work ? But it seems plawyright tests are failing on that ? I would just check HTTP code 200, maybe just mime type. Pixel comparaison is very fragile

I hope I will have some insights from @mdouchin

lizmap/modules/lizmap/lib/Request/WFSRequest.php
```diff
@@ -849,7 +849,36 @@ protected function getfeaturePostgres()
...
+                                if ($pathInfo['filename'] !== '') {
+                                    $geoJson['properties'][$key] = $pathInfo['filename'].'.'.$pathInfo['extension'];
+                                } else {
+                                    $geoJson['properties'][$key] = $pathInfo['basename'];
+                                }
+                            }
+                        }
+                    }
+                }
+                $d->geojson = json_encode($geoJson);
```
mdouchin, line 880: This code is only used for PostgreSQL layers, since you wrote it in the `getfeaturePostgres()` method. Perhaps it could also be great to support other types of layers used in Lizmap (users can have a FGB or GeoPackage edited from QGIS Desktop and containing WebDav fields pointing to the same WebDav server configured in `profiles.ini.php`. Not a strong opinion about it, as it could be added in the future.

Be aware that if there are some issues with the built SQL query (PostgreSQL errors), or if the QGIS layer contains joined fields from another layer, the function `getfeaturePostgres()` fallbacks to `return $this->getfeatureQgis();`, so you are not 100% sure that the Webdav fields will be correctly parsed, even for PostgreSQL layers.

I think this code must be extracted in a new function `processFeatureGeoJSON` that we could use in both methods `getfeaturePostgres()` and return `getfeatureQgis()`

You should not do anything if there are no webdav fields. Please move 
`$webDavConfiguration = $this->qgisLayer->getWebDavFieldConfiguration();` before calling the new function `processGeoJSONFeature` and if possible do not check if there are webdav fields for each feature but before (I think you should have all information before entering the foreach). Something like : 

```php
// Not sure there must be a new method $this->hasSomeWebdavFields()
// if the test is simple
foreach($q as $d) {
    if ($this->hasSomeWebdavFields()) {
        // Use the function processFeatureGeoJSON containing your code (without getWebDavFieldConfiguration)
        $d->geojson = $this->processFeatureGeoJSON($d->geojson);
    }
    yield $virg.$d->geojson;
    $virg = ',';
}
```

If we would like to compute WebDav fields only for PostgreSQL, we could also transfer the logic in the `buildQueryBase` method ( https://github.com/3liz/lizmap-web-client/blob/ca669772529ca59249babff35f9c306e14b8d294/lizmap/modules/lizmap/lib/Request/WFSRequest.php#L505 ) which could use PostgreSQL functions to directly change the values. Not a strong opinion on this : this would be more performant (no need to json_decode/json_encode) but would prevent from using it for non PostgreSQL layers

mdouchin (changes requested):

lizmap/modules/lizmap/lib/Request/RemoteStorageRequest.php
mind84: OK thank you. Actually in most cases the method returns the error message so I will remove the logs inside the method and let the caller log the error or not

lizmap/modules/lizmap/lib/Request/WFSRequest.php
mind84: > Be aware that if there are some issues with the built SQL query (PostgreSQL errors), or if the QGIS layer contains joined fields from another layer, the function getfeaturePostgres() fallbacks to return $this->getfeatureQgis();, so you are not 100% sure that the Webdav fields will be correctly parsed, even for PostgreSQL layers.

Right. Thanks for pointing it out to me.

> I think this code must be extracted in a new function processFeatureGeoJSON that we could use in both methods getfeaturePostgres() and return getfeatureQgis()

I agree. For me the goal is to support WebDav fileds for all datasources available, non only PostgreSQL. 

Thanks

mind84: @mdouchin,
`getfeatureQgis` returns directly `OGCResponse` object. In order to process the response body I would store it  in a variable, process it and then return new OGCResponse like in the `getfeaturePostgres` function. Follow this idea, I would create a new function that return OGCResponse that can be called by both getFeatureQgis and getFeaturePostgres, something like that:
```php
protected function sendOGCResponse($httpCode,$contentType,$data){
     //.....
     return new OGCResponse($httpCode,$contentType,$data);
}
```
Do you think this could be the right approach?

I'm wondering if I should call `$this->processFeatureGeoJSON($geoJson)` anyway and move inside it the webdav logics. What do you think? Might it be useful in the future to have it for purposes other than webdav?

mdouchin: You should not store de response in a variable, because it can be very heavy. This is why we know stream the response from QGIS and send it back to the brower in chunk by also streaming the response.
If you want to modify the DAV url you need to use tools such as json-machine 
My colleague @rldhont will give more details

rldhont: As @mdouchin do not stored response in a variable. You have to work as a stream.

First of all, move this code in its own function `processFeatureProperties`

Then you have to use `json-machine` to update the stream provided in `getFeatureQgis`
```php

        // Else pass query to QGIS Server
        // Get remote data as stream
        $wfsResult = $this->request(true, true);
        if ($wfsResult->code >= 400) {
            return $rep;
        }
        if ($wfsResult->mime != 'application/vnd.geo+json; charset=utf-8') {
            return $rep;
        }
        if (!$this->hasSomeWebdavFields()) {
            return $wfsResult;
        }
        if ($wfsResult->code >= 400) {
            return $rep;
        }
        if ($wfsResult->mime != 'application/vnd.geo+json; charset=utf-8') {
            // Do not process other format ?
            return $rep;
        }

        $featureStream = \Psr7\StreamWrapper::getResource($wfsResult->getBodyAsStream());
        $features = \JsonMachine\Items::fromStream($featureStream, ['pointer' => "/features"]);
        return new OGCResponse(200, 'application/vnd.geo+json; charset=utf-8', (function () use ($q) {
            yield '{"type": "FeatureCollection", "features": [';
            $virg = '';
            foreach ($features as $feat) {
                if (property_exists($feat, 'properties')) {
                    $feat->properties = $this->processFeatureproperties($feat->properties);
                }
                yield $virg.json_encode($feat);
                $virg = ',';
            }
      );
```

mind84:
@mdouchin,

I didn't mark as resolved the last conversation since I'm not sure if the changes are what you are expected. 

A couple of question:
 - the WFS XML request is forwarded to the function `getfeatureQgis` before `$this->qgisVectorLayer` is set. I ended up not parsing the WFS response in this case, but I'm not sure if this is the right solution. Alternatively, is it safe to get a local instance of `qgisVectorLayer` to get its webdav configuration?
 
 - as suggested by @laurentj, it might be useful to set a maximum upload file size. In this case, this parameter should be configurable to provide some flexibility to users. Do you think it could be set in `profiles.ini.php` in the `webdav` configuration? If `maxUploadFileSize` is not set, then fall back to a default value (2 MB?)

Thanks

laurentj:
>it might be useful to set a maximum upload file size. this parameter should be configurable to provide some flexibility to users

It can be, however, keep in mind that there is a also a such limit into the PHP configuration (upload_max_filesize and post_max_size), and so this parameter cannot be higher than this limit, and the PHP limits cannot be changed by Lizmap.

lizmap/modules/lizmap/lib/Request/WFSRequest.php
mind84: marked as resolved since this part is no longer needed.

rldhont (approved):

rldhont:
Thanks @mind84 !

assets/src/legacy/attributeTable.js
```diff
@@ -1511,17 +1510,31 @@ var lizAttributeTable = function() {
...
                         // Check if we need to replace url or media by link
+                        let davConf = lizUrls['webDavUrl'] && config.layers[aName]?.webDavFields && Array.isArray(config.layers[aName].webDavFields) && config.layers[aName].webDavFields.includes(columnName) || false;
                         colConf['render'] = function (data, type, row, meta) {
                             // Replace media and URL with links
                             if (!data || !(typeof data === 'string'))
                                 return data;
+                            if (davConf) {
+                                // replace the root of the url
+                                if(data.startsWith(lizUrls['webDavUrl'])){
+                                    data = data.replace(lizUrls['webDavUrl'],'dav/')
```
Gustry, line 1521: I'm not a PHP dev, but is-it possible to have a static variable for `dav/` ?
It's used in many places : L1521, L1532, lizmap/modules/lizmap/lib/Request/WMSRequest.php lizmap/modules/view/classes/popup.class.php, lizmap/modules/view/controllers/media.classic.php ? (or where it's appropriate if any)

Because this is a kind of "overpass" specific in Lizmap.
It's look easier to debug in IDE : right click, search usages in case of a refactoring etc
Thanks if it's possible, otherwise, skip :)

@mdouchin Please confirm if possible.

tests/docker-conf/phpfpm/profiles.ini.php
```diff
@@ -162,4 +162,8 @@
 ;db=2
 ;ttl=0
 
-
+[webdav:default]
```
Gustry, line 165: Should the distributed `profiles.ini.php.dist` be updated with a commented example ? (at least to advertise a little bit these new variables)
lizmap/var/config/profiles.ini.php.dist

If yes, I would add a quick doc and warning about webdav URL being exposed in the WFS service, so we can prevent any users complaining about this security "issue".

Adding the warning `profiles.ini.php.dist` will warn sysadmin about this.

lizmap/modules/lizmap/lib/Request/RemoteStorageRequest.php
```diff
@@ -0,0 +1,316 @@
...
+namespace Lizmap\Request;
+
+use GuzzleHttp\Client;
+use GuzzleHttp\Psr7;
+
+class RemoteStorageRequest
+{
+    protected static $webDAVNamespace = 'DAV:';
+
+    protected static $fileNameExpression = 'file_name(@selected_file_path)';
```
Gustry, line 22: This will need to be written in the documentation.
Side question,does `file_name( @selected_file_path   )` work ? or it needs to very explicit in the doc :)

mind84: Agree, this should be documented. As you pointed out, the syntax of `file_name(@selected_file_path)` is specific and the feature doesn't work if there is trail spaces or other chars. Actually, `file_name(@selected_file_path)` is mandatory

Gustry: I just tried in QGIS desktop, using the UI to build an expression with click and click.

When double clicking items from the center to add the function on the left panel, it adds automatically two spaces ` file_name( ` before and after the inserted function, so I would expect a lot of users not being careful, even if it's clearly written in the documentation.

image

It would be way better to use a regepx. Tested very quickly `file_name\(\s*@selected_file_path\s*\)`

I guess trailing spaces works ? ` file_name(@selected_file_path)`

In the doc, it would still needed to say no multiline expression or comment.

assets/src/legacy/attributeTable.js
mind84: @Gustry
**js**: I've not a strong opinion on this. If `'dav/'` is used only in `attributeTable.js ` for now I would leave it in plain text, but if it could be used in other contexts could you point me out the right place to eventually set this variable (`lizMap` object or `lizMap.config` object maybe? ). In this case, also the `media/` string could be stored in a variable?

**php**: you can have an utility class that expose public static properties, in this case I would use `RemoteStorageRequest` class directly

Gustry: Oh yes, one was in JS indeed.
I saw you did in PHP, thanks. I find it easier to debug later, to have something centralized, or helpful in IDE.

Yes, I think `media/` might like 10 years old in Lizmap... At that time, the source code was way smaller, so that's why it wasn't done, always a good idea to start new good practice ;-)

mind84: @Gustry,
ok, I'm going to add a property in `lizUrls`, seems to me that could be the right place to store stuff about... urls.
I'm not going to add a property for `/media` in this PR because I would keep things distinct

Thanks

Gustry: Of course, I didn't ask about `media`, it was just to explain the context, 10 years ago.

assets/src/legacy/attributeTable.js
```diff
@@ -1511,17 +1510,31 @@ var lizAttributeTable = function() {
                         }
                     } else {
                         // Check if we need to replace url or media by link
+                        let davConf = lizUrls['webDavUrl'] && config.layers[aName]?.webDavFields && Array.isArray(config.layers[aName].webDavFields) && config.layers[aName].webDavFields.includes(columnName) || false;
```
mdouchin, line 1513: It seems the` || false` for debugging purpose and should be removed ?

mdouchin (changes requested):

mdouchin:
@mind84 do you prefer that we merge your commits "as is" (i.e. add them all to the master branch with an additional merge commit), or you would prefer to reorganise/merge/rename your commits ?

mind84:
>do you prefer that we merge your commits "as is" (i.e. add them all to the master branch with an additional merge commit), or you would prefer to reorganise/merge/rename your commits ?

@mdouchin,
Makes no difference to me, I'll follow your indications.

Thanks

Gustry (approved):

mdouchin:
Thanks a lot

PR #3947, opened 2023-10-30
[Feature] Load layers as single WMS layer

mind84:
Allows to load project layers as single WMS Layer.

### Feature description:
This development is intended to increase LZM performance when large projects containing many layers are loaded.
This option is presented as alternative to classic mode, and users should be able to switch from one mode to another by acting on the Lizmap plugin configuration.

Despite the performance improvements, some functionality are no longer available when a single layer is loaded, i.e.:
- acting on the opacity of a single layer
- define the image type of the single layer
- turning off a layer is no longer immediate, but the system must reload the entire image
- excluding cached map layers from the single WMS image breaks the layer tree order

### Logics in detail

- **General** 
I used the same classic layer map instance for loading the single WMS image. The advantage is that you can inherit the general configuration of the main map without duplicating it. If the main map `useTileWms` property is `true` , then the single WMS layer is loaded as `TileWMS`.

- **Baselayers**
All `external` baselayers and `cached` ones (`WMTS`) comes independently. Only `Lizmap` baselayers (i.e. loaded by QGIS server) are loaded in the singleWMS layer together with map layers. 

- **Cached map layers**
All `cached` map layers (i.e. tiled `WMTS` layers) comes independently 

- **Filtering features/ filtering by legend**
These two functionalities (alone or in combination) works also with single WMS layer 

### Consideration about layers order
As said, if some layer are `tiled` then they will be not included in the single WMS. This unfortunately has a very high chance of breaking the layers map order and there is no way (at least IMHO) to avoid this. I'll explain by an example the current layer order logic when single WMS is turned on. 

Take for example the following `layertree`:

```
Layer tree
 Map layers
   Layer A -> classic layer, included in single WMS
   Layer B -> classic layer,  included in single WMS
   Layer C -> cached WMTS layer, excluded from single WMS
   Layer D -> classic layer,  included in single WMS
   Layer E -> cached WMTS layer, excluded from single WMS
 Baselayers
   OpenStreetMap (or Bing or whatever) -> excluded from WMS
   BaseLayer E -> Lizmap layer, included in single WMS
```
On the map, the real order is:

**CASE WHEN** you have selected the `OpensStreetMap` baselayer

```
Map order:
   --------- Cached layers block, ordered 
   Layer C
   Layer E
   --------- single WMS layers block, ordered
   Layer A
   Layer B
   Layer D
   --------- Baselayer
   OpenStreetMap (or Bing or whatever)
```

**CASE WHEN** you have selected the `Lizmap` `BaseLayer E` as baselayer

```
Map order:
   --------- Cached layers block, ordered 
   Layer C
   Layer E
   --------- single WMS layers block, ordered
   Layer A
   Layer B
   Layer D
   BaseLayer E
  ---------------
```
The base layers are obviously always placed at the bottom, but for the other two block I had to choose (arbitrarily, I would say) which one goes on top . IMHO there is no general right solution for this problem, but maybe you have a some suggestions about that.

Funded by Faunalia


mind84:
### Test notes:
Waiting for Lizmap plugin "single layer" option development, you can switch from a mode to another by adding the attached js file as custom js in media folder. 

[switch_to_single_layer.js.zip](https://github.com/3liz/lizmap-web-client/files/13207252/switch_to_single_layer.js.zip)

For now, I've only added a test project `single_wms_image.qgs`. 

**e2e**
I think it's best to wait for a first review before writing tests

Known failure on e2e tests when single Layer mode is turned on:
 - Tests related to `opacity` fails, since the opacity option is hidden and users can't act on it
 - some tests when comparison between screenshots is
 required (due to the fact that the image is loaded into a separate map)

rldhont:
Before doing some reviews, we need to decide where and how to store in the config this new option.

@Gustry  @mdouchin Do we add `wmsSingleLayer` or `wmsSingleMapLayer` in options ?

mdouchin:
> Before doing some reviews, we need to decide where and how to store in the config this new option.
> 
> @Gustry @mdouchin Do we add `wmsSingleLayer` or `wmsSingleMapLayer` in options ?

I would use `wmsSingleRequestForAllLayers`

Side idea: I think we should remove the active base layer (and obviously the externally requested WMS layers such as TMS) from the "single" image request. This would avoid having an empty map when zooming or panning (the base layer would come independently)

mind84:
> Side idea: I think we should remove the active base layer (and obviously the externally requested WMS layers such as TMS) from the "single" image request. This would avoid having an empty map when zooming or panning (the base layer would come independently)

@mdouchin ,
For now baseLayer(s) are excluded from the single request. At least that's what I tried to do :)

assets/src/modules/SingleWMSLayer.js
```diff
@@ -0,0 +1,331 @@
...
+        return this._layer;
+    }
+    /**
+     * creates the layer instance and get the startup single image 
+     * @memberof SingleWMSLayer
+     */
+    initializeLayer(){
+        let minResolution =  Utils.getResolutionFromScale(this._minScale, this._metersPerUnit);
+        let maxResolution =  Utils.getResolutionFromScale(this._maxScale, this._metersPerUnit); 
+
```
rldhont, line 185: You need to take into account the `wmsMaxWidth` and `wmsMaxHeight` https://github.com/3liz/lizmap-web-client/blob/master/assets/src/modules/map.js#L115 and the `pixelRatio` https://github.com/3liz/lizmap-web-client/blob/master/assets/src/modules/map.js#L121

mind84: Agree. In addition, based on these info, if `useTileWms` is true, then I should load the whole wms image as `TileWMS`

mind84: @rldhont 
for single layers `minResolution` and `maxResolution` are calculated for each node. Since with singleWMS these informations are lost, is it ok to left these parameters `undefined` for the singleWMS?

tests/js-units/node/config.test.js
```diff
@@ -130,4 +130,17 @@ describe('Config', function () {
...
+        expect(capabilities.Capability).to.not.be.undefined
+        const config = JSON.parse(readFileSync('./data/single_wms_image-config.json', 'utf8'));
+        expect(config).to.not.be.undefined
+
+        const initialConfig = new Config(config, capabilities);
+
+        expect(initialConfig.options).to.be.instanceOf(OptionsConfig)
+        expect(initialConfig.options.wms_single_request_for_all_layers).to.be.eq("True")
+    })
+
```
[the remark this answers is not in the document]
mind84: @rldhont 

> And add expect(opt.wms_single_request_for_all_layers).to.be.eq(false) to Valid test of optons.test.js

This one is perfectly clear to me, I've added this part.

> Can you rewrite this test to add to js-units/config/options.test.js with expect(opt.wms_single_request_for_all_layers).to.be.eq(false) ?

About this one I'm not sure I understand what you mean, maybe there is a typo on `to.be.eq(false)`? I want to test **singleWMS = true** since singleWMS = false is the default. Should I remove this test from `js-units/node/config.test.js` and add it to` js-units/config/options.test.js` with `expect(opt.wms_single_request_for_all_layers).to.be.eq(true)`?

Thanks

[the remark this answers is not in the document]
rldhont: @mind84 Sorry, What I meant was to add a test in `js-units/config/options.test.js` with `expect(opt.wms_single_request_for_all_layers).to.be.eq(true)` based on this config.

assets/src/modules/config/Options.js
```diff
@@ -43,25 +44,26 @@ export class OptionsConfig  extends BaseObjectConfig {
...
+     * @param {string}   cfg.datavizLocation                      - the popup location in the User interface: dock, bottom-dock, right-dock
+     * @param {boolean}  [cfg.hideProject]                        - is the project hidden in user interface ? Only services are available.
+     * @param {number}   [cfg.wmsMaxHeight]                       - the image max height for WMS GetMap request
+     * @param {number}   [cfg.wmsMaxWidth]                        - the image max width for WMS GetMap request
+     * @param {boolean}  [cfg.fixed_scale_overview_map]           - does the Overview map have fixed scale ?
+     * @param {boolean}  [cfg.use_native_zoom_levels]             - does the map use native zoom levels ?
+     * @param {boolean}  [cfg.hide_numeric_scale_value]           - does the scale line hide numeric scale value ?
+     * @param {boolean}  [cfg.hideGroupCheckbox]                  - are groups checkbox hidden ?
+     * @param {boolean}   [cfg.wms_single_request_for_all_layers] - are layers loaded as single WMS image ?
```
rldhont, line 66:
```suggestion
     * @param {boolean}  [cfg.wms_single_request_for_all_layers]  - are layers loaded as single WMS image ?  
```
Just a bit of formatting.

assets/src/modules/map.js
```diff
@@ -730,6 +759,21 @@ export default class map extends olMap {
...
+        return this._useTileWms;
+    }
+
+    get customTileGrid(){
+        return this._customTileGrid;
+    }
+
+    get hidpi(){
+        return this._hidpi;
+    }
```
rldhont, line 776: DocStrings needed, thanks.

assets/src/modules/state/BaseLayer.js
```diff
@@ -31,6 +31,21 @@ export class BaseLayerState extends EventDispatcher {
...
+
+    /**
+     * set if the base layer is loaded in a single layer ImageLayer or not
+     * @param {boolean} val
+     */
+    set singleWMSLayer(val){
+        this._singleWMSLayer = val;
+    }
+    /**
+     * @type {boolean}
```
rldhont, line 45: Description, thanks

rldhont (changes requested):

rldhont (approved):

mdouchin:
@mind84 What about the WMS layers when the user has checked the checkbox `Get images directly from WMS Server` ?

I would not put them in the "single layer" group of layers, but I have no strong opinion about it. Any hint @Gustry @rldhont ?

mind84:
Hi @mdouchin!

> What about the WMS layers when the user has checked the checkbox `Get images directly from WMS Server` ?

Right, I didn't consider that option.

In this case excluding it means that the layer will go on top.

rldhont:
> @mind84 What about the WMS layers when the user has checked the checkbox `Get images directly from WMS Server` ?
> 
> I would not put them in the "single layer" group of layers, but I have no strong opinion about it. Any hint @Gustry @rldhont ?

I think, if `single layer` is checked, then `Get images directly from WMS Server` is disabled.

Gustry:
> I think, if single layer is checked, then Get images directly from WMS Server is disabled.

This is not done for now on the plugin side.

Gustry (approved):
From the UX point of view

Gustry:
> I think, if single layer is checked, then Get images directly from WMS Server is disabled.

Can you make a ticket in plugin ?

@nboisteault Is this PR OK for you ?

nboisteault (changes requested):
@mind84 Many thanks for this work. There is just those messages in the console to look at please:


https://github.com/3liz/lizmap-web-client/assets/2145040/ac222c3e-4696-40ab-ab08-1636cad7eee2

mind84:
> @mind84 Many thanks for this work. There is just those messages in the console to look at please:

Yep, thanks, I'll fix it

nboisteault (approved):

PR #3969, opened 2023-11-16
[Bugfix] fixed typo qsgmtime

mind84:
Fixed a typo that prevented projects with embedded layers from being read correctly from the `ProjectCache`

Funded by Faunalia


mdouchin (approved):
Thanks !

nworr (approved):

Issue #3939, opened 2023-10-23
Support for Vector Tiles


chrimohr:
In my QGIS Project, I'm using a Vector Tile layer. 
URL: https://sgx.geodatenzentrum.de/gdz_basemapde_vektor/tiles/v1/bm_web_de_3857/{z}/{x}/{y}.pbf
Style URL: https://sgx.geodatenzentrum.de/gdz_basemapde_vektor/styles/bm_web_gry.json

Support for Vector Tiles would be a great addition.
It may be already possible but I could not figure it out and found no resources about it.

geographit2:
@nboisteault , Would it be possible to get an idea of the cost to fund this feature?

Also, more generally, I think it would be great to have a ballpark estimate of the cost for features tagged "Sponsor or PR needed."

Thanks

Gustry:
> I've made a quick draft to use your vector tiles layer and its style : https://github.com/3liz/lizmap-web-client/pull/3985

Link to the branch instead https://github.com/nboisteault/lizmap-web-client/tree/vector_tiles

Gustry:
There is a QGIS Server plugin to have vector tiles out from QGIS Server : https://github.com/3liz/qgis-server-tiles-plugin

CC @vvmruder as you commented on the PR #3985

PR #3985, opened 2023-11-24
Add a vector tiles layer as a base layer

nboisteault:
Related to #3939


vvmruder:
@nboisteault I appreciate this enhencement a lot. Still I dont get what is the idea here. Currently its possible to add VectorTiles in the QGIS project and configure it as a base layer. Once done we can use these in our lizmap webclient. However, the result is rendered in the underlying QGIS Server always. Thats really slow.

Is the idea of this PR to make vector tiles being rendered by the Lizmap WebClient directly without  passing it through the QGIS Server?

nboisteault:
> Is the idea of this PR to make vector tiles being rendered by the Lizmap WebClient directly without passing it through the QGIS Server?

Yes it is

Gustry:
Wrong rebase @nboisteault ?
Can we close this PR ? as it didn't get activity since 2023.
The ticket is still open #3939

PR #4001, opened 2023-12-04
Extendable map parameters for redirection after authentication

rldhont:
Funded by [SDEC énergie](https://www.sdec-energie.fr/)


nboisteault:
@rldhont what is the context? What does it fix?

rldhont:
> @rldhont what is the context? What does it fix?

@nboisteault it fixes IDU parameter for NetADS module, that will be lost in case of redirection to authentication.

PR #4023, opened 2023-12-14
init FormControl property only if exists

nworr:
To prevent deprecated notice on php8.2

Funded by 3Liz


nworr:
check ok with php8.2, should we use php8.2 only for master ?

laurentj (approved):

Gustry:
Any reason why removing the 3.7 label ?
Can it be merged ? I will rebase #3983, CI was failing because of this as well.

nworr:
the commit dd0361384632801bb106933a3b1655611c248b24  update the php version used in test to php8.2 , great for master branch, but we might want to make keep lizmap 3.7 on php 8.1 ?

commit fa4faae68f2a05c1c4465d1fd5eb2af252e75a09 will be cherry-picked to release_3_7 branch

Gustry:
Ok ! I could it in #3983 if needed. OK, go for it.

Issue #3558, opened 2023-03-17
Shows less decimals for project extent in project info


gioman:
Is not really useful to show as many decimals

![image](https://user-images.githubusercontent.com/1951107/226024220-29529897-177e-40c5-ac5c-616e35c627d9.png)

and also makes things harder to show in a sensible (small) width/space.

gioman:
The same apply here:

![image](https://user-images.githubusercontent.com/1951107/226024951-81f13c66-c1d8-4c15-9e55-297bdc41a053.png)

In this case a further improvement would be to show the extent with a semi transparent rectangle on top of a base layer, i.e. OSM like QGIS does in its CRS selector

![image](https://user-images.githubusercontent.com/1951107/226026621-45124c45-257e-4d3e-846b-b99f95058beb.png)

PR #4032, opened 2023-12-18
Less decimals for bbox info on project view page

nworr:
Round value for project bounds in the project view page.

I also use a css transition instead of jquery slide
image

Ticket : #3558

Funded by 3Liz

### DIFF
--- a/lizmap/modules/view/classes/lizmapMainViewItem.class.php
+++ b/lizmap/modules/view/classes/lizmapMainViewItem.class.php
@@ -35,7 +35,7 @@ class lizmapMainViewItem
         $this->abstract = $abstract;
         $this->keywordList = $keywordList;
         $this->proj = $proj;
-        $this->bbox = $bbox;
+        $this->bbox = preg_replace('/([0-9]+\.[0-9]{2})[0-9]+/', '$1', $bbox);
         $this->url = $url;
         $this->img = $img;
         $this->order = $order;
--- a/assets/src/legacy/view.js
+++ b/assets/src/legacy/view.js
@@ -1,16 +1,3 @@
-var addDescriptionSlider = function(){
-    $('.liz-project-img').parent().mouseenter(function(){
-        var self = $(this);
-        self.find('.liz-project-desc').slideDown();
-    }).mouseleave(function(){
-        var self = $(this);
-        self.find('.liz-project-desc').hide();
-    }).click(function(){
-        var self = $(this);
-        window.location = self.parent().find('a.liz-project-view').attr('href');
-        return false;
-    });
-}
 
 var searchProjects = function(){
     // Hide search if there are no projects
@@ -287,6 +274,5 @@ var searchProjects = function(){
 }
 
 window.addEventListener('load', function () {
-    addDescriptionSlider();
     searchProjects();
 });
--- a/lizmap/modules/view/templates/view.tpl
+++ b/lizmap/modules/view/templates/view.tpl
@@ -14,8 +14,9 @@
     <a name="link-projet-{$idm}"></a>
     <div class="thumbnail">
       <div class="liz-project">
-        <img width="250" height="250" loading="lazy" src="{$p->img}" alt="project image" class="liz-project-img">
-        <p class="liz-project-desc" style="display:none;">
+        <a href="{$p->url}{if $hide_header}&h=0{/if}">
+        <img width="250" height="250" loading="lazy" src="{$p->img}" alt="project image" class="_liz-project-img">
+        <p class="liz-project-desc" >
           <b class="title">{$p->title}</b>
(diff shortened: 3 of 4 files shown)


gioman:
Fixes https://github.com/3liz/lizmap-web-client/issues/3558

Gustry (approved):
From a UX point of view, good.

Do you want to backport ?

nworr:
only 3.7 ?

nworr:
precision : it doesn't round values in the information panel

gioman:
> precision : it doesn't round values in the information panel

@nworr it would be highly appreciated if it would include also the info panel.

Also seems to me that

https://github.com/3liz/lizmap-web-client/pull/4032/files#diff-2c0e8b758097ca053a90b477e70751986f294f1d80fd8b85e0096e46dd21b4b0R38

just takes the first two decimal places from the bbox values, I think it would be better to round the values instead, and add some logic because for gepographical CRSs 2 decimal places is hardly enough to describe with precision an extent that has values in degrees (while is ok for projected CRSs).

lizmap/modules/view/classes/lizmapMainViewItem.class.php
```diff
@@ -35,7 +35,7 @@ public function __construct($id, $title, $abstract = '', $keywordList = '', $pro
         $this->abstract = $abstract;
         $this->keywordList = $keywordList;
         $this->proj = $proj;
-        $this->bbox = $bbox;
+        $this->bbox = preg_replace('/([0-9]+\.[0-9]{2})[0-9]+/', '$1', $bbox);
```
rldhont, line 38:
```suggestion
        $this->bbox = preg_replace('/([0-9]+\.[0-9]{6})[0-9]+/', '$1', $bbox);
```

Can you provide 6 decimals ?

rldhont (changes requested):

lizmap/modules/view/classes/lizmapMainViewItem.class.php
gioman: Not sure if I get the code at all... From a UX it would be nice to show 2 decimal places (rounded) for projected CRS, 6 I guess would be ok for geographical ones (degrees). Six decimals for CRSs in meters or feet do not give a good UX.

Gustry:
It's just my opinion, but these kind of information about extent, is not really human readible, and doesn't need to be displayed on the main landing page of Lizmap Web Client (which is the "page vitrine" of the LWC project, to give a first preview of the project).

If a GIS user wants to have it's extent, IMHO he should open the project and go in the information tab. It's really targeting precise GIS technicians be able to ready text representing coordinates.

I really like the work from @jersig on his instance to make something "neat" and "slick" about the landing page of LWC : https://mapeo.calvados.fr/carto/index.php hiding quite a few complex information for most users.

Just my opinion (I got inspired from Spotify or Netflix apps, where just title, description and thumbnail are displayed nicely). All "secondary information" are coming in the second step (actors, producters,number of episodes). Sorry for the comparaison :)

gioman:
Not stale, actually a very welcome UX improvement (this would also need to be applied to the BBOX values as shown in the project infos after having opened the map, not only in the LMWC landing page).

nworr:
This PR is now outdated, and have conflicts with master branch (and release_3_8 for sure).
The logic of simple rounding value is quite weak, as you noticed.
It need to use a smarter way to round value depending on projection unit, and it will be performed in another way (dedicated class/ method, maybe only on browser side).
Thus IMHO, another PR will be more relevant, and the original ticket keep track of this one which is usefull (even closed)

Gustry:
IMHO, displaying coordinates in a web interface is very GIS technicians oriented. I don't think that reading coordinates is done by a lot of humans, in a web interface :)

I really would like to see something more user-friendly, similar to what Calvados has done for instance, slick, very oriented for non-GIS technicians : 

Peek 2024-07-08 17-47

gioman:
@Gustry what I can say... I agree! but since that bbox values are shown since... always, I assumed that you wanted to keep them. If in reality you want to get rid of them, please go on!

Gustry:
Ah great, as my previous comment just above last April didn't get attraction, and as you commented "not stale", I thought you were interested into this PR :D
I don't need these coordinates either ;-)

It looks IMHO easier to close this PR if we agree and to create a new one removing codes ;-)

This interface needs a more modern UI, which started in LWC 3.7 with this new theme. Hands are welcome, if you have some design, CSS skills etc.

@nworr, it's up to you :)

Gustry:
This branch didn't get any activity since end of 2023, can we close this PR ?
Your branch can remain in your fork, and you should create a ticket to reference it.

Moreover, it seems that the discussion here goes to "remove these coordinates", no to display them.

Issue #4046, opened 2023-12-29
[Bug]: Group as layer option breaks layer order


mind84:
### What is the bug?

When you group multiple layers as a single layer in your Lizmap configuration (`Group as layer` option) in some cases the order of the layers displayed on the client does not match the order of the layer tree.

In detail, the last layers in the tree before the `baselayer group` (if any) are displayed below the active `baselayer`. The number of layers involved in this behavior depends on how many layers are grouped as single layer and also on the position of the single layer group in the layer tree. Given that, the entire behavior is not completely predictable.

However, the issue appears to be related to the `zIndex` layer property set during map initialization as the [total layers count](https://github.com/3liz/lizmap-web-client/blob/1ecbe1a942e70bc1cb2b844727d6a302ff73fe91/assets/src/modules/BaseLayersMap.js#L245) is always lesser than the highest `node.layerOrder` value. Therefore, as result of [this](https://github.com/3liz/lizmap-web-client/blob/1ecbe1a942e70bc1cb2b844727d6a302ff73fe91/assets/src/modules/BaseLayersMap.js#L368) calculation, some layers may have `zIndex` lesser than 0.

I think this happens because the total layer count is calculated by counting the `Group as layer` layer as one (no matter how many layers are placed in it), while the layer order is calculated for [each](https://github.com/3liz/lizmap-web-client/blob/1ecbe1a942e70bc1cb2b844727d6a302ff73fe91/assets/src/modules/state/Layer.js#L544) single layer indipendently.
 

### Steps to reproduce the issue

Use this test project to see the bug 
[group_as_layer_layertree.zip](https://github.com/3liz/lizmap-web-client/files/13793952/group_as_layer_layertree.zip)

### Versions

**Versions :**

* Lizmap Web Client : 3.7.1-pre
* Lizmap plugin : 4.0.1
* QGIS Desktop : 3.28.7
* QGIS Server : 3.22.16
* Py-QGIS-Server : 1.8.8
* QGIS Server plugin atlasprint : not found
* QGIS Server plugin lizmap_server : 2.8.2-alpha
* QGIS Server plugin wfsOutputExtension : 1.8.1-alpha

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Debian

### Browsers

Firefox, Chrome

### Browsers version

latest

PR #4053, opened 2024-01-04
[Fix] Fix the display order of layers when the 'Group as layer' option is used

mind84:
Ticket : #4046

Funded by Faunalia


rldhont (changes requested):
@mind84 can you add js-unit tests ?

mind84:
Hi @rldhont,
> @mind84 can you add js-unit tests ?

I added a simple equality comparison between the output of the new function and the expected number of layers.
If you have any suggestions for improving this, please let me know

Thanks

tests/js-units/node/state/maplayer.test.js
```diff
@@ -10,6 +10,7 @@ import { LayerIconSymbology, LayerSymbolsSymbology, SymbolIconSymbology } from '
 import { LayerGroupState, LayerVectorState, LayersAndGroupsCollection } from '../../../../assets/src/modules/state/Layer.js';
 
 import { MapLayerLoadStatus, MapGroupState, MapLayerState } from '../../../../assets/src/modules/state/MapLayer.js';
+import exp from 'constants';
```
rldhont, line 13: Seems not be used

rldhont (approved):
Thanks for the test.
It seems 1 line has been added and not used.

tests/js-units/node/state/maplayer.test.js
mind84: Ouch, I didn't notice, sorry. It seems to have been added automatically by my text editor.

mind84:
closes #4046

PR #4058, opened 2024-01-08
Editing form - Upload fields: allow to use an expression to set up the storage path

mdouchin:
In QGIS editing form configuration, a file-based field (image or generic file) can be set up with an expression to calculate the target directory where the file must be stored. For example:

```sql
-- Here the target directory is calculated based on the value of the field "feature_code" in the form
concat(
	'media/some_target_folder/',
	CASE
		WHEN lower(trim(to_string(current_value('feature_code')))) IN ('', 'null') OR current_value('feature_code') IS NULL 
			THEN 'no_code'
		ELSE trim(to_string(current_value('feature_code')))
	END,
	'/'
)
```

This piece of code adds the same capability in Lizmap Web Client.

Funded by Cartophyl https://www.cartophyl.com/

### DIFF
--- a/lizmap/modules/lizmap/lib/Form/QgisForm.php
+++ b/lizmap/modules/lizmap/lib/Form/QgisForm.php
@@ -171,12 +171,16 @@ class QgisForm implements QgisFormControlsInterface
         // Hide when no modify capabilities, only for UPDATE cases (  when $this->featureId control exists )
         if (!empty($featureId) && strtolower($capabilities->modifyAttribute) == 'false') {
             foreach ($toDeactivate as $de) {
-                if ($form->getControl($de)) {
+                /** @var null|\jFormsControl $getCtrl */
+                $getCtrl = $form->getControl($de);
+                if ($getCtrl) {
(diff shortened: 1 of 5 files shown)


mdouchin:
This is a draf. Tests must be added, PHP Stan must be fixed

lizmap/modules/lizmap/lib/Form/QgisForm.php
```diff
@@ -1030,25 +1146,46 @@ protected function getParsedValue($ref, $geometryColumn)
...
+
+        // We must check if the saveFile action has worked
+        if (
+            !$saveFile
+            && !is_file($targetFullPath.$newFilename)
+            && !is_file($targetFullPath.$filename)
+        ) {
+            // TODO : raise an exception or an error
+            // to let the user know the file has not been saved ?
+            // Or just a message and go back to the
```
laurentj, line 1186: to the ? :-)

laurentj (approved):
LGTM

rldhont (approved):

PR #4069, opened 2024-01-16
Increase `uploadedImageMaxWidthHeight` value for 4K screens

nboisteault:
Many people now have a 4K screen of smartphones w/ big resolution In Lizmap 3.7 the WMS getMap requests has a ratio of 1.1 so we need to set `4224` as the default value.
Many admin don't know `uploadedImageMaxWidthHeight` exists so they don't increase its value, so I prefer to put a high value as default

Funded by 3Liz


nboisteault:
@rldhont I forgot about https://github.com/3liz/lizmap-web-client/pull/4002

PR #4078, opened 2024-01-19
method to check if projet is compatible with target version

nworr:
Projet written for a  version N are supported until N+2 version

the method can be usefull for testing projets before upgrade

Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/Project.php
+++ b/lizmap/modules/lizmap/lib/Project/Project.php
@@ -2314,6 +2314,19 @@ class Project
     }
 
     /**
+     * test if provided version (int format 30520).
+     *
+     * @param int $targetVersion the desired version (maj.min) as int (<Maj><Min as 2 digit>00)
+     */
+    public function isCompatibleWithVersion(int $targetVersion): bool
+    {
+        // the projet version written for N are supported until N+2
+        $maxSupportedVersion = $this->getLizmapWebClientTargetVersion() + 200;
+
+        return $maxSupportedVersion >= $targetVersion;
+    }
+
+    /**
      * Check acl rights on the project.
      *
      * @return bool true if the current user as rights on the project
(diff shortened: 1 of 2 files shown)


rldhont (approved):

lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -2313,6 +2313,19 @@ public function needsUpdateWarning()
         return false;
     }
 
+    /**
+     * test if provided version (int format 30520).
```
Gustry, line 2317: I guess this docstring needs to be rephrased ?

lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -2313,6 +2313,19 @@ public function needsUpdateWarning()
...
+     *
+     * @param int $targetVersion the desired version (maj.min) as int (<Maj><Min as 2 digit>00)
+     */
+    public function isCompatibleWithVersion(int $targetVersion): bool
+    {
+        // the projet version written for N are supported until N+2
```
Gustry, line 2323:
```suggestion
        // the project version written for version N are supported until version N+2
```

Gustry (changes requested):
I totally understand the PR, it works these days because we have followed this rule until now.

With one release per year, this was giving 2 years of time for a "sleeping project" to be displayed.

* 3.4 : 2020
* 3.5 : 2021
* 3.6 : 2022
* 3.7 : 2023

The variable `lizmapWebClientTargetVersion` in the [conf file](https://github.com/3liz/lizmap-web-client/blob/4f4eb755ec4f2d09e7d87fc332a31c5c4f183369/lizmap/app/system/mainconfig.ini.php#L39) was somehow edited at the time of the release of **X.Y.0** more or less to let the user enough time with these projects and the variable `lizmapWebClientTargetVersion` could be adjusted accordingly.

With the planned new release schedule with 2 releases per year, this is considerably reducing the timeframe (see internal issues/232#note_67164). Increasing the release schedule.

With this PR, it will be now hardcoded N+2. If we still go this way, I would rephrase the maybe as well in the conf file. Or I would even remove the configuration file and the function `getLizmapWebClientTargetVersion` because this would be only using the current LWC version +2 or -2 (depending of the context)

Issue #4076, opened 2024-01-18
[Bug]: Error when repository defined in conf does not exists


nworr:
### What is the bug?

When a repository is defined and the folder doesn't exists, error 500 and you can see in log
```
2024-01-18 12:22:02 185.243.122.225 [0] scandir(): Argument #1 ($directory) cannot be empty
 /srv/lizmap/lib/lizmap-web-client/lizmap_web_client-3.7.1.7084/lizmap/modules/view/zones/main_view.zone.php   61

```

a fix can be made by adding is_dir() before `scan_dir() `[there](https://github.com/3liz/lizmap-web-client/blob/40b8a39fc285c4e9733e2e94220ef7bb1666f702/lizmap/modules/view/zones/main_view.zone.php#L61)

### Steps to reproduce the issue

1. Create a folder 
2. Define a repository for this folder
3. Delete the folder
4. Go to index.php

### Versions

LWC 3.7.1

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

whatever

### Browsers

Firefox

### Browsers version

whatever

### Relevant log output

```shell
2024-01-18 12:22:02 185.243.122.225 [0] scandir(): Argument #1 ($directory) cannot be empty
 /srv/lizmap/lib/lizmap-web-client/lizmap_web_client-3.7.1.7084/lizmap/modules/view/zones/main_view.zone.php   61
```

PR #4080, opened 2024-01-19
check if repository has a valid path

nworr:
Add a method for poject to check is path is valid

in view check path is valid before performing scan_dir
in admin view : add a badge for bad path

Ticket : #4076
Screenshot 2024-01-19 at 12-09-00 Administration

Funded by 3Liz


Gustry (approved):
From a UX point of view, LGTM

I just got the issue on my local instance, by removing a bunch of "Lizmap directories" from CLI

rldhont (approved):

PR #4085, opened 2024-01-19
[Bugfix] JS: Get typeName from config

rldhont:
Funded by 3Liz


nboisteault (approved):
LGTM but I'd like to have a project to reproduce and test the issue.

PR #4055, opened 2024-01-05
use css transitions instead of js for project description

nworr:
commit cherry-picked from https://github.com/3liz/lizmap-web-client/pull/4032

Funded by 3Liz

PR #4099, opened 2024-01-26
[end2end]fix DOM search in cypress test

nworr:
Ticket : https://github.com/3liz/lizmap-web-client/pull/4055#issuecomment-1908260909

Funded by 3Liz


nworr:
cypress test in now playwright, not sure if cold cache/hot cache make sense anymore

nboisteault:
> cypress test in now playwright, not sure if cold cache/hot cache make sense anymore

This was due to a bug but I think it is to much tests for a non important feature.

PR #4104, opened 2024-01-29
Tooltips with OL9

nboisteault:
Funded by [PNR Ballon des Vosges](https://www.parc-ballons-vosges.fr/)


assets/src/modules/Tooltip.js
```diff
@@ -0,0 +1,197 @@
...
+                style: vectorStyle
+            });
+
+            // Load tooltip layer
+            this._activeTooltipLayer.once('sourceready', () => {
+                mainEventDispatcher.dispatch('tooltip.loaded');
+            });
+
+            this._activeTooltipLayer.on('error', () => {
+                console.log(`Tooltip layer '${layerName}' could not be loaded.`);
```
mdouchin, line 118: Could be great to display a message to the user with `addMessage` . Better if it can be translated. 
`An error prevented to load the tooltip data`

In the console, try to also display the real error message if possible

mdouchin (changes requested):

assets/src/modules/Tooltip.js
nboisteault: Done. The real error message from OL don't say anything more.

PR #4114, opened 2024-01-30
Fix regrssion : keep atlasprint plugin

rldhont:
QGIS Server does not work atlas print does.

### DIFF
--- a/assets/src/components/FeatureToolbar.js
+++ b/assets/src/components/FeatureToolbar.js
@@ -512,38 +512,18 @@ export default class FeatureToolbar extends HTMLElement {
     }
 
     printAtlas(templateName) {
-        const projectProjection = mainLizmap.config.options.qgisProjectProjection.ref;
         const wmsParams = {
             SERVICE: 'WMS',
-            REQUEST: 'GetPrint',
+            REQUEST: 'GetPrintAtlas',
             VERSION: '1.3.0',
             FORMAT: 'pdf',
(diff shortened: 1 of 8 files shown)


nboisteault (approved):

Gustry (changes requested):
* The file given to the user is a random string `ts_gkX7a` and the output format extension is missing. Extension is required, using a proper name
* The print on the "Montpellier" project http://localhost:8130/index.php/view/map?repository=montpellier&project=montpellier is not printing base layer, is-that correct ?

rldhont:
Changes done

Gustry:
The AtlasPrint is not required to run LWC. Pressing the button will lead to a 404 from QGIS server and the button will not work. Possible to use the PHP class `lizmap/modules/lizmap/lib/Server/Server.php` ?

Gustry:
Changes in the documentation : https://github.com/3liz/lizmap-documentation/pull/246

Your PR is partially reverting 25ee656a54609fa3c7fe921f52fb98c43a798196, just to keep track for later

Gustry (approved):
Looks better to me

Gustry:
Another related commit in the plugin [2c504c0faa86972d1136da16f4dc4a78f6c4a7ee](https://github.com/3liz/lizmap-plugin/commit/2c504c0faa86972d1136da16f4dc4a78f6c4a7ee)

PR #4127, opened 2024-01-31
JS: Created the map and modules after toolbar created

rldhont:
To be sure, the map and modules are ready at the `uicreated` event, the map and modules are activated at the `toolbarcreated` event.


assets/src/modules/Lizmap.js
```diff
@@ -61,14 +61,15 @@ export default class Lizmap {
 
                 register(proj4);
 
+                console.log(lizMap.map.getProjection());
```
nboisteault, line 64: is it wanted?

rldhont: no

PR #4137, opened 2024-02-01
[Bugfix] Build WMTS options return null

rldhont:
The WMTS options could be null if the layer has not be found in WMTS Capabilities

Funded by Agence de l'eau Rhône Méditerranée Corse

### DIFF
--- a/assets/src/modules/map.js
+++ b/assets/src/modules/map.js
@@ -180,20 +180,34 @@ export default class map extends olMap {
                 let minResolution = node.wmsMinScaleDenominator <= 1 ? undefined : Utils.getResolutionFromScale(node.layerConfig.minScale, metersPerUnit);
                 let maxResolution = node.wmsMaxScaleDenominator <= 1 ? undefined : Utils.getResolutionFromScale(node.layerConfig.maxScale, metersPerUnit);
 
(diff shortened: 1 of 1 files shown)


assets/src/modules/map.js
```diff
@@ -180,20 +180,30 @@ export default class map extends olMap {
...
+                // The layer is configured to be cached
                 if (node.layerConfig.cached) {
+                    // Using WMTS
                     const parser = new WMTSCapabilities();
                     const result = parser.read(lizMap.wmtsCapabilities);
+                    // Build WMTS options
                     const options = optionsFromCapabilities(result, {
                         layer: node.wmsName,
                         matrixSet: mainLizmap.projection,
                     });
```
nboisteault, line 192:
```suggestion
                    let options;

                    // Build WMTS options
                    if (result['Contents']['Layer']) {
                        options = optionsFromCapabilities(result, {
                            layer: node.wmsName,
                            matrixSet: mainLizmap.projection,
                        });
                    }
```

We need this because of this fix in progress: https://github.com/openlayers/openlayers/pull/15525

nboisteault (changes requested):

PR #4149, opened 2024-02-02
Bugfix: WMTS for layers with shortname 

rldhont:
* Fixing WMTS capabilities for cached layers with shortname defined
* Adding WMTS end2end tests with a cached layer which has a different shortname from name.

Funding by [Agence de l'eau Rhône Méditerranée Corse](https://www.eaurmc.fr/)


nboisteault:
@rldhont I've made the wmts_test project in a quick and dirty manner without thinking to integrate it to tests projects. Would it be better to add a layer in the `cache` project and complete requests-getcapabilities-ghaction.js test?

Issue #4030, opened 2023-12-16
LMWC 3.7.0: background layer added with QuickMapServices do not show


gioman:
### What is the bug?

As far as I understand, tell me if I'm wrong, there baselayers we used to add within the baselayers tab in LM plugin (that are not printable and not re-projectable) are not supported anymore, right?

Instead anything in the "baselayers" group become a baselayer and it seems to me that in the LM plugin is suggested to use QuickMapservices, right?

I tried to add a base layer (Bing satellite) there with QMS and it is not showing in the project when loaded.

This new paradigm for baselayers make sense in a way, but leave me with a few questions:

* the legacy un-printable and un-reprojectable base layers where much faster to load compared to the ones loaded with QMS or other remote services (WMS  et al), in a scenario where printing was not an issue they were very useful. 

* when using the latest version of the LM plugin the layer properties for anything inside the "baselayers" group is completely disabled, why? How is controlled how those layers are shown (i.e. title, rendering)

![image](https://github.com/3liz/lizmap-web-client/assets/1951107/795d7e53-7119-493b-b96d-a0bcb42dd1b0)

### Steps to reproduce the issue

Test project here:

[dorset.zip](https://github.com/3liz/lizmap-web-client/files/13692637/dorset.zip)

### Versions

LMWC 3.7.0

On Ubuntu 22.04 with QGIS Server 3.28.12, Project saved on QGIS 3.28.12 ans LM plugin 4.1.1

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 22.04

### Browsers

Firefox, Chrome

### Browsers version

Latest updates

gioman:
Also:

how is controlled which is the default baselayer

![image](https://github.com/3liz/lizmap-web-client/assets/1951107/bdd8fed4-69ad-420b-ba56-fc9b2fa9f9be)

if the control we used for is not not populated?

![image](https://github.com/3liz/lizmap-web-client/assets/1951107/4efe4dfd-0792-4f93-bda8-097bf4079ec0)

Changing the order in the group makes no difference.

Gustry:
> when using the latest version of the LM plugin 

> LM plugin 3.18.1

We are at version 4.1.1, not latest in your case.
https://plugins.qgis.org/plugins/lizmap/#plugin-versions

> unclear of tho choose the default baselayer 

[...]

PR #4150, opened 2024-02-02
[Feature] Add QMS Bing Maps

mind84:
**TL;DR**

Enabling Bing Maps to be loaded as baselayers in LWC through OpenLayers dedicated implementation by importing Bing Maps in QGIS project from [QuickMapServices QGIS Plugin](https://plugins.qgis.org/plugins/quick_map_services/)

**DETAILS**

In previous versions of LWC (i.e. 3.6) it was possible to add a defined set of external layers (including Bing) by flagging the corresponding option in the plugin's `base-layer` panel and without adding any layer to the project. Then the plugin updates the `cfg` accordingly by setting the corresponding properties in the option section of the cfg:

```json
"options": {


        "osmMapnik": "True",
        "openTopoMap": true,
        "bingKey": "some___key",
        "bingStreets": "True",
        "bingSatellite": "True",

}
```
 
Now, since the direction is to rely on the QGIS project, in newer versions this panel has become deprecated and the above properties that defines these external layers are no longer written in `cfg` file.
Furthermore, following the update of the OpenLayers library, some of this external layers are not longer supported (e.g. Google Maps) by the lib itself.

In the "Bing" specific case, the current version of OpenLayers offers supports for rendering Bing tiles.

This implementation allows LWC to recognize the `QMS Bing Maps` from QGIS project, read the [corresponding](https://github.com/3liz/lizmap-plugin/pull/550) `Bing Api Key` from the `cfg` and instantiate with these options the Bing OpenLayers [class](https://openlayers.org/en/latest/apidoc/module-ol_source_BingMaps-BingMaps.html) 

**NOTES**

- This functionality relies for now on the external `url` assigned to the layer by QMS when it is added on QGIS project. This url is set in the `url` property of the `externalAccess` object in the `cfg` response.

- I guess [this object](https://github.com/3liz/lizmap-web-client/blob/2f8d01ce0b67bf1cb24cb8f4446fed9088b05c76/assets/src/modules/config/BaseLayer.js#L751) was only kept for backwards compatibility since the properties listed are no longer written to the `cfg` file. 

- I've extended [this default configuration](https://github.com/3liz/lizmap-web-client/blob/2f8d01ce0b67bf1cb24cb8f4446fed9088b05c76/assets/src/modules/config/BaseLayer.js#L542) with a new `QMSExternalLayer object`, to keep things clearly separated (and hopefully maintainable). 

-  urls of Bing maps currently supported:
   * `Bing Map`, alias `bing-road` in the LWC < 3.7 configuration. url = `https://ecn.dynamic.t0.tiles.virtualearth.net/comp/CompositionHandler/{q}?mkt=en-us&it=G,VE,BX,L,LA&shading=hill`
   * `Bing Satellite`, alias `bing-aerial` in the LWC < 3.7 configuration. url = `https://ecn.t3.tiles.virtualearth.net/tiles/a{q}.jpeg?g=0&dir=dir_n`. 
 
You can check this functionality in the `bing_basemap.qgs` project included in this PR.

Thanks

Related to https://github.com/3liz/lizmap-web-client/issues/4030

Funded by Faunalia

### DIFF
--- a/tests/js-units/data/bing_basemap-config.json
+++ b/tests/js-units/data/bing_basemap-config.json
@@ -1,0 +1,204 @@
+{
+    "locateByLayer": {},
+    "formFilterLayers": {},
+    "attributeLayers": {},
+    "layers": {
+        "baselayers": {
+            "id": "baselayers",
+            "name": "baselayers",
+            "type": "group",
+            "title": "baselayers",
+            "abstract": "",
+            "link": "",
+            "minScale": 1,
+            "maxScale": 1000000000000,
+            "toggled": "True",
+            "popup": "False",
+            "popupSource": "auto",
+            "popupTemplate": "",
+            "popupMaxFeatures": 10,
+            "popupDisplayChildren": "False",
+            "popup_allow_download": true,
+            "legend_image_option": "hide_at_startup",
+            "groupAsLayer": "False",
+            "baseLayer": "False",
+            "displayInLegend": "True",
+            "singleTile": "True",
+            "imageFormat": "image\/png",
+            "cached": "False",
+            "clientCacheExpiration": 300,
+            "shortname": "baselayers",
+            "mutuallyExclusive": "True"
+        },
+        "bing roads": {
+            "id": "Bing_Map_d0e5fe42_3d9b_4ca8_9e1d_7312d05b08e1",
+            "name": "bing roads",
+            "type": "layer",
+            "extent": [
+                -20037508.342789244,
+                -20037508.342789248,
+                20037508.342789244,
+                20037508.342789248
+            ],
+            "crs": "EPSG:3857",
+            "title": "bing roads",
+            "abstract": "",
+            "link": "",
+            "minScale": 1,
+            "maxScale": 1000000000000,
+            "toggled": "False",
+            "popup": "False",
+            "popupSource": "auto",
+            "popupTemplate": "",
+            "popupMaxFeatures": 10,
+            "popupDisplayChildren": "False",
+            "popup_allow_download": true,
+            "legend_image_option": "hide_at_startup",
+            "groupAsLayer": "False",
+            "baseLayer": "False",
+            "displayInLegend": "True",
+            "singleTile": "True",
+            "imageFormat": "image\/png",
+            "cached": "False",
+            "clientCacheExpiration": 300,
+            "shortname": "bing_roads",
+            "layerType": "raster",
+            "externalWmsToggle": "True",
+            "externalAccess": {
+                "type": "xyz",
+                "zmin": "1",
+                "zmax": "19",
+                "url": "https:\/\/ecn.dynamic.t0.tiles.virtualearth.net\/comp\/CompositionHandler\/{q}?mkt=en-us&it=G,VE,BX,L,LA&shading=hill",
+                "crs": "EPSG:3857"
+            }
+        },
+        "bing aerial": {
+            "id": "Bing_Satellite_5a4c914b_cf62_481a_944b_ff5c1645b49b",
+            "name": "bing aerial",
+            "type": "layer",
+            "extent": [
+                -20037508.342789244,
+                -20037508.342789248,
+                20037508.342789244,
+                20037508.342789248
+            ],
+            "crs": "EPSG:3857",
+            "title": "bing aerial",
+            "abstract": "",
+            "link": "",
+            "minScale": 1,
+            "maxScale": 1000000000000,
+            "toggled": "False",
+            "popup": "False",
+            "popupSource": "auto",
+            "popupTemplate": "",
+            "popupMaxFeatures": 10,
+            "popupDisplayChildren": "False",
+            "popup_allow_download": true,
+            "legend_image_option": "hide_at_startup",
+            "groupAsLayer": "False",
+            "baseLayer": "False",
+            "displayInLegend": "True",
+            "singleTile": "True",
+            "imageFormat": "image\/png",
+            "cached": "False",
+            "clientCacheExpiration": 300,
+            "shortname": "bing_aerial",
+            "layerType": "raster",
+            "externalWmsToggle": "True",
+            "externalAccess": {
+                "type": "xyz",
+                "zmin": "1",
+                "zmax": "18",
+                "url": "https:\/\/ecn.t3.tiles.virtualearth.net\/tiles\/a{q}.jpeg?g=0&dir=dir_n'",
+                "crs": "EPSG:3857"
+            }
+        }
+    },
+    "timemanagerLayers": {},
+    "atlas": {
+        "layers": []
+    },
+    "tooltipLayers": {},
+    "layouts": {
+        "config": {
+            "default_popup_print": false
+        },
+        "list": []
+    },
+    "loginFilteredLayers": {},
+    "filter_by_polygon": {
+        "config": {
+            "polygon_layer_id": null,
+            "group_field": "",
+            "filter_by_user": false
+        },
+        "layers": []
+    },
+    "datavizLayers": {
+        "layers": [],
+        "dataviz": [],
+        "locale": "en_US"
+    },
+    "metadata": {
+        "qgis_desktop_version": 32815,
+        "lizmap_plugin_version_str": "4.1.8",
+        "lizmap_plugin_version": 40108,
+        "lizmap_web_client_target_version": 30800,
+        "lizmap_web_client_target_status": "Dev",
+        "instance_target_url": "http:\/\/localhost:8130\/",
+        "instance_target_repository": "testsrepository"
+    },
+    "options": {
+        "projection": {
+            "proj4": "+proj=merc +a=6378137 +b=6378137 +lat_ts=0 +lon_0=0 +x_0=0 +y_0=0 +k=1 +units=m +nadgrids=@null +wktext +no_defs",
+            "ref": "EPSG:3857"
+        },
+        "bbox": [
+            "398421.76870000001508743",
+            "5386390.87650000024586916",
+            "467349.17460000002756715",
+            "5427483.42289999965578318"
+        ],
+        "mapScales": [
+            1,
+            1000000000
+        ],
+        "minScale": 1,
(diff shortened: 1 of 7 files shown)


Gustry:
Thanks for tackling this !
I will let others commenting about the technical part. I thought some work was done about these layers.

Just a note for now :

> This functionality relies for now on the default layer name assigned by QMS when it is added on QGIS project, so to get it work the layer name should not be changed (I'm not sure how limiting this might be)

> Default names of Bing maps currently supported:

This is fragile and will be broken IMHO. (or a dev in the plugin must be done, but I don't think it's the way to go)

In the QGS file of your PR : 

```xml
<layer-tree-layer source="type=xyz&amp;zmin=1&amp;zmax=18&amp;url=https://ecn.t3.tiles.virtualearth.net/tiles/a{q}.jpeg?g%3D0%26dir%3Ddir_n'" checked="Qt::Checked" patch_size="-1,-1" name="Bing Satellite" providerKey="wms" legend_exp="" expanded="1" legend_split_behavior="0" id="Bing_Satellite_9ceaa33e_9c61_4fd0_b607_68ffd5614d88">
```

QGIS is checking if the `{q}` is in the datasource : https://github.com/qgis/QGIS/blob/61fe17774e1eae721b6a616a2bbe934ca399a6ca/src/providers/wms/qgswmsprovider.cpp#L1679
if we want less fragile, we could check the URL is `virtualearth.net` ?

mind84:
> This is fragile and will be broken IMHO. (or a dev in the plugin must be done, but I don't think it's the way to go)

I agree with you and I think this is the most tricky part. Indeed, this is the main reason I've created a draft :)

> if we want less fragile, we could check the URL is virtualearth.net ?

Thanks for checking this! Surely the `url` is much more stable than the `default name`. I will try with this solution but I think that probably it requires some server side implementation.

Thanks

mind84:
@Gustry,

I've changed the logic using the `url` instead of the `name`. It's a little tricky because I've not found a very strong way to separate `roads` from `satellite` map type using just the url.
I don't know if this result would be solid enough/acceptable. This way, however, users will gain more flexibility on the layer name and will be able to change it in QGIS project.

Gustry (approved):
LGTM from the user point of view

The URL checks looks better like this.

tests/js-units/data/bing_basemap-config.json
```diff
@@ -0,0 +1,206 @@
...
+        "maxScale": 1000000000,
+        "use_native_zoom_levels": true,
+        "hide_numeric_scale_value": false,
+        "initialExtent": [
+            398421.7687,
+            5386390.8765,
+            467349.1746,
+            5427483.4229
+        ],
+        "bingKey": "AgBOKzHM1S17uF6szXY93vepXIx2Cl91mkrlP-yrsngz1jdREjrtucUduz56hDNT",
```
nworr, line 179: is this key a dev key ? can it be exposed publicly  ?

tests/qgis-projects/tests/bing_basemap.qgs.cfg
```diff
@@ -0,0 +1,169 @@
...
+        "maxScale": 1000000000,
+        "use_native_zoom_levels": true,
+        "hide_numeric_scale_value": false,
+        "initialExtent": [
+            398421.7687,
+            5386390.8765,
+            467349.1746,
+            5427483.4229
+        ],
+        "bingKey": "AgBOKzHM1S17uF6szXY93vepXIx2Cl91mkrlP-yrsngz1jdREjrtucUduz56hDNT",
```
nworr, line 37: same question

nworr (commented):
Good PR, thanks for the detailed description. 
Only  a question about the exposed bing key

tests/js-units/node/state/baselayer.test.js
```diff
@@ -290,4 +290,26 @@ describe('BaseLayersState', function () {
...
+    it('Bing Maps baselayers', function () {
+        const baseLayers = getBaseLayersState('bing_basemap')
+        expect(baseLayers.selectedBaseLayerName).to.be.eq('bing roads')
+        expect(baseLayers.selectedBaseLayer).to.not.be.undefined
+        expect(baseLayers.selectedBaseLayer.name).to.be.eq('bing roads')
+        expect(baseLayers.baseLayerNames).to.be.an('array')
+        .that.have.length(2)
+        .that.be.deep.eq([
+            "bing roads",
+            "bing areals",
```
Gustry, line 302:
```suggestion
            "bing aerial",
```
?

tests/qgis-projects/tests/bing_basemap.qgs
```diff
@@ -0,0 +1,778 @@
...
+        <Option type="Map">
+          <Option type="QString" value="baselayers" name="wmsShortName"/>
+        </Option>
+      </customproperties>
+      <layer-tree-layer legend_split_behavior="0" expanded="1" providerKey="wms" source="crs=EPSG:3857&amp;format&amp;type=xyz&amp;url=https://ecn.dynamic.t0.tiles.virtualearth.net/comp/CompositionHandler/%7Bq%7D?mkt%3Den-us%26it%3DG,VE,BX,L,LA%26shading%3Dhill&amp;zmax=19&amp;zmin=1" id="Bing_Map_a346c65b_6c10_4b31_80b7_9ad63af606b6" name="bing roads" checked="Qt::Checked" legend_exp="" patch_size="-1,-1">
+        <customproperties>
+          <Option/>
+        </customproperties>
+      </layer-tree-layer>
+      <layer-tree-layer legend_split_behavior="0" expanded="1" providerKey="wms" source="crs=EPSG:3857&amp;format&amp;type=xyz&amp;url=https://ecn.t3.tiles.virtualearth.net/tiles/a%7Bq%7D.jpeg?g%3D0%26dir%3Ddir_n'&amp;zmax=18&amp;zmin=1" id="Bing_Satellite_9ceaa33e_9c61_4fd0_b607_68ffd5614d88" name="bing areals" checked="Qt::Unchecked" legend_exp="" patch_size="-1,-1">
```
Gustry, line 35:
```suggestion
      <layer-tree-layer legend_split_behavior="0" expanded="1" providerKey="wms" source="crs=EPSG:3857&amp;format&amp;type=xyz&amp;url=https://ecn.t3.tiles.virtualearth.net/tiles/a%7Bq%7D.jpeg?g%3D0%26dir%3Ddir_n'&amp;zmax=18&amp;zmin=1" id="Bing_Satellite_9ceaa33e_9c61_4fd0_b607_68ffd5614d88" name="bing aerial" checked="Qt::Unchecked" legend_exp="" patch_size="-1,-1">
```

tests/qgis-projects/tests/bing_basemap.qgs
```diff
@@ -0,0 +1,778 @@
...
+  </mapcanvas>
+  <projectModels/>
+  <legend updateDrawingOrder="true">
+    <legendgroup open="true" name="baselayers" checked="Qt::Checked">
+      <legendlayer showFeatureCount="0" open="true" drawingOrder="-1" checked="Qt::Checked" name="bing roads">
+        <filegroup open="true" hidden="false">
+          <legendlayerfile layerid="Bing_Map_a346c65b_6c10_4b31_80b7_9ad63af606b6" isInOverview="0" visible="1"/>
+        </filegroup>
+      </legendlayer>
+      <legendlayer showFeatureCount="0" open="true" drawingOrder="-1" checked="Qt::Unchecked" name="bing areals">
```
Gustry, line 84:
```suggestion
      <legendlayer showFeatureCount="0" open="true" drawingOrder="-1" checked="Qt::Unchecked" name="bing aerial">
```

tests/qgis-projects/tests/bing_basemap.qgs
```diff
@@ -0,0 +1,778 @@
...
+        <xmax>179.99999999999997158</xmax>
+        <ymax>85.05112877980660357</ymax>
+      </wgs84extent>
+      <id>Bing_Satellite_9ceaa33e_9c61_4fd0_b607_68ffd5614d88</id>
+      <datasource>crs=EPSG:3857&amp;format&amp;type=xyz&amp;url=https://ecn.t3.tiles.virtualearth.net/tiles/a%7Bq%7D.jpeg?g%3D0%26dir%3Ddir_n'&amp;zmax=18&amp;zmin=1</datasource>
+      <shortname>Bing_Satellite</shortname>
+      <keywordList>
+        <value></value>
+      </keywordList>
+      <layername>bing areals</layername>
```
Gustry, line 365:
```suggestion
      <layername>bing aerial</layername>
```

tests/qgis-projects/tests/bing_basemap.qgs.cfg
```diff
@@ -0,0 +1,169 @@
...
+            "groupAsLayer": "False",
+            "baseLayer": "False",
+            "displayInLegend": "True",
+            "group_visibility": [],
+            "singleTile": "True",
+            "imageFormat": "image/png",
+            "cached": "False",
+            "clientCacheExpiration": 300
+        },
+        "bing areals": {
```
Gustry, line 110: same here and other lines below

Gustry (commented):
Is-it a typo ?

tests/js-units/data/bing_basemap-config.json
mind84: Hi @nworr,
I thought I could used it because it is already [exposed](https://github.com/3liz/lizmap-web-client/blob/2d836511de728cf24730a53ee98b0febbbd32ebb/tests/qgis-projects/tests/base_layers.qgs.cfg#L45) in `base_layers.qgs.cfg`. 

Thanks

tests/js-units/node/state/baselayer.test.js
mind84: It is

tests/js-units/data/bing_basemap-config.json
nworr: i didn't even notice this project use bing map :)

nworr (approved):

Gustry:
Just for information, see #5944

PR #4174, opened 2024-02-06
Fix: Get back to the way lizmap groups and layers are checked

rldhont:


rldhont:
I close it because I propose Double Clicking #4179

PR #4179, opened 2024-02-08
UI: Double clicking on group propagating the checked state

rldhont:
When double clicking on a group label or checkbox, the new checked state of the group will be propagated through the tree to all the children.

With this capability, the user will be able to check all the layers contained in a group whatever the depth.

lizmap-dblclicking-treeview

### DIFF
--- a/assets/src/components/Treeview.js
+++ b/assets/src/components/Treeview.js
@@ -21,11 +21,13 @@ export default class Treeview extends HTMLElement {
     constructor() {
         super();
         this._itemNameSelected;
+        this._clickTimestamp;
     }
 
     connectedCallback() {
 
         this._onChange = () => {
+            if (this._freeze) return;
             render(this._rootTemplate(mainLizmap.state.layerTree), this);
         };
 
@@ -96,10 +98,14 @@ export default class Treeview extends HTMLElement {
             }
             <div class="${layer.checked ? 'checked' : ''} ${layer.type} ${layer.name === this._itemNameSelected ? 'selected' : ''}">
                 <div class="loading ${layer.loadStatus === MapLayerLoadStatus.Loading ? 'spinner' : ''}"></div>
-                <input type="checkbox" class="${parent.mutuallyExclusive ? 'rounded-checkbox' : ''}" id="node-${layer.name}" .checked=${layer.checked} @click=${() => layer.checked = !layer.checked} >
+                <input type="checkbox"
+                    class="${parent.mutuallyExclusive ? 'rounded-checkbox' : ''}"
+                    id="node-${layer.name}"
+                    .checked=${layer.checked}
+                    @click=${() => layer.checked = !layer.checked} >
                 <div class="node ${layer.isFiltered ? 'filtered' : ''}">
                     <img class="legend" src="${layer.icon}">
-                    <label for="node-${layer.name}">${layer.layerConfig.title}</label>
+                    <label for="node-${layer.name}" >${layer.layerConfig.title}</label>
                     <div class="layer-actions">
                         <a href="${this._createDocLink(layer.name)}" target="_blank" title="${lizDict['tree.button.link']}">
                             <i class="icon-share"></i>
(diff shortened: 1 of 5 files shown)


Gustry:
Nice idea !

assets/src/components/Treeview.js
```diff
@@ -51,13 +51,18 @@ export default class Treeview extends HTMLElement {
...
+                        .checked=${item.checked}
+                        @click=${() => item.checked = !item.checked}
+                        @dblclick=${() => this._dblclicked(item)} >
                     <div class="node ${item.isFiltered ? 'filtered' : ''}">
                         ${item.type === 'layer'
                             ? html`<img class="legend" src="${item.icon}">`
                             : ''
                         }
-                        <label for="node-${item.name}">${item.layerConfig.title}</label>
+                        <label for="node-${item.name}" style="user-select: none;" @dblclick=${(e) => {e.preventDefault(); this._dblclicked(item);}} >${item.layerConfig.title}</label>
```
nboisteault, line 65: Better put styles in CSS.

assets/src/components/Treeview.js
```diff
@@ -132,6 +137,15 @@ export default class Treeview extends HTMLElement {
         return visibility;
     }
 
+    _dblclicked(item) {
+        if (item.type != 'group') {
```
nboisteault, line 141: To avoid unwanted listeners, we could add `@dblclick` event listener only for groups here : https://github.com/3liz/lizmap-web-client/pull/4179/files#diff-25aa8d9133469826d98486d1fbe1d18f36ca5c72a569d73aa6e56c9c07d49836R65

nboisteault (changes requested):

assets/src/components/Treeview.js
rldhont: @nboisteault have you an idea how to do so in CSS ? which file ? which css rule ?

assets/src/components/Treeview.js
nboisteault: It might not be important so you can ignore this comment.

nboisteault:
Sometimes the double click does not work. It might due to the label being linked to the checkbox.
@Gustry @mdouchin could you also test?


https://github.com/3liz/lizmap-web-client/assets/2145040/3f62fe94-a20e-4785-8dfa-559c6c625714

Gustry:
> Sometimes the double click does not work.

Indeed, I can confirm.
An idea, maybe an icon could be added, similar to the "I" ? (not the best, but similar)

rldhont:
@Gustry @nboisteault @mdouchin It is ready for review!

nboisteault (approved):
Tested and approved!

Gustry:
Strange, I can't even try on the "Montpellier" project : 

image

@nboisteault Which project did you use ?

rldhont:
> My comments above remains valid, I don't have any chekbox to try on the "Montpellier" project. On which project are you trying ?

The "Montpellier" project activates the option to disabled checkboxes for groups.

I am using the "testsrepository/treeview" project.

Gustry:
> I am using the "testsrepository/treeview" project.

Ok, but the first "group" in this project, it doesn't work for me with this PR.

rldhont:
> > I am using the "testsrepository/treeview" project.
> 
> Ok, but the first "group" in this project, it doesn't work for me with this PR.

What do you mean by it doesn't work for me ?

Gustry:
Freshly load the map http://lizmap.local:8130/index.php/view/map?repository=testsrepository&project=treeview
Peek 2024-04-15 16-06

I do only double click

Gustry:
This PR is a good move, but I'm not 100% convinced by the approach, given the number of feedbacks on this. A double click is "unknown" on a web interface.

Fine to discuss

rldhont:
@nboisteault @Gustry @mdouchin @mind84 @gioman @josemvm @guenterw 

I proposed the double click to propagate checked status to tree children, but I can implement an other way to do so. What is your prefered way to propagate checked status to tree children: 

1.  Double click
2. Click right
3. Click middle
4. Long press click

guenterw:
If I understand (translate) it correctly, it is the function that is used in QGIS with Ctrl+click.
Since not every user uses QGIS, I think the double-click is also best

josemvm:
hi @rldhont for LWC >=3.7 we lost the  text of the tooltip that is shown when the mouse passes over the layer name in the layer tree and i don't know if that was intentional... regardless of the choice, how do you intend to inform the user of the existence of this functionality?

however, i prefer double click

guenterw:
One more basic thought about the group checkboxes:
Since you have to decide whether to display them or not, I always choose to hide the group checkboxes.
For example, if you have a group of aerial images with a checkbox, with the aerial images of the last 10 years, then there is always the risk that the group is checked and thus all 10 aerial images are loaded on top of each other.
Or in the case of development plans: There is a group of development plans, including groups for 6 districts and the respective plans. Now it can make sense to show all plans for a district, but not necessarily for the whole municipality.

On the other hand, it often makes sense to activate all layers of a group. I would currently control this via the QGIS themes.

Thus, there is actually the desire to determine the display of the group checkbox per group.
But this then leads to an inconsistent picture of the legend. Is this acceptable?

@Gustry  I think the tutorial is a good idea

PR #4185, opened 2024-02-09
Docs - Add JSDoc

Gustry:
Not perfect, just a POC
I will let JS developers to tweak the output

Temporary preview
https://gustry.github.io/lizmap-web-client/

CC @rldhont and @nboisteault 

Let's revive PHPDoc as well

### DIFF
--- a/assets/src/modules/config/FormFilter.js
+++ b/assets/src/modules/config/FormFilter.js
@@ -1,9 +1,9 @@
 /**
- * @module utils/FormFilter.js
+ * @module config/FormFilter.js
  * @name FormFilter
  * @copyright 2023 3Liz
  * @author DHONT René-Luc
- * @license MPL-2.0 - Mozilla Public License 2.0 : http://www.mozilla.org/MPL/
+ * @license MPL-2.0
  */
 
 import { BaseObjectConfig } from './BaseObject.js';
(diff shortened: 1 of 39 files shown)


assets/src/modules/config/FormFilter.js
```diff
@@ -1,6 +1,7 @@
 /**
  * @module utils/FormFilter.js
- * @copyright 2023 3Liz
+ * @name FormFilter
+ * @copyright 2025 3Liz
```
Gustry, line 4:
```suggestion
 * @copyright 2023 3Liz
```

assets/src/modules/state/MapLayer.js
```diff
@@ -437,7 +448,7 @@ export class MapGroupState extends MapItemState {
 
     /**
      * Find layer and group items
-     * @returns {Array<MapLayerState|MapGroupState>}
+     * @returns {Array<MapLayerState|MapGroupState>} All tThe map layer and map group states
```
Gustry, line 451:
```suggestion
     * @returns {Array<MapLayerState|MapGroupState>} All map layers and map group states
```

assets/src/modules/state/MapLayer.js
```diff
@@ -574,7 +585,7 @@ export class MapLayerState extends MapItemState {
     /**
      * Update WMS selected layer style name
      * based on wmsStyles list
-     * @param {string} styleName
+     * @param {string} styleName - the WMS layer style name to select
```
Gustry, line 588:
```suggestion
     * @param {string} styleName - The WMS layer style name to select
```

Gustry:
For now, to have CI passing with a green status, I have forced the exit code `make js || exit 0`.

I would like to remove the `exit 0` because it's not showing us potential JS error in parsing.

Would it be possible to fix https://github.com/Gustry/lizmap-web-client/actions/runs/7869841518/job/21469785233 ?

PR #4190, opened 2024-02-12
JS - Update package.json version

Gustry:
Following #4185 it seems that this version should be increased.

I'm still not sure what version we should use. I don't think `3.8` is a valid value, and I would like to keep semver.

But JSDoc will build a different website for each different version. I need to check that.

Maybe only update X.Y.0, whatever the bugfix release number ?


nboisteault (approved):
It is ok for me.

PR #4093, opened 2024-01-24
[Bugfix] Too many embedded layers cause php to hit max_execution_time 

mind84:
There are cases where a project might embed several layers from another project. 

In this scenario Lizmap [loads](https://github.com/3liz/lizmap-web-client/blob/1b302d40f667706f05f4f4b0e8882720a779449f/lizmap/modules/lizmap/lib/Project/QgisProject.php#L1494) the same project (embedded project) multiple times for each of these layers and this behavior could cause php to hit `max_execution_time` if there are many embedded layers.

To fix that I've grouped the embedded layers and load the related QGIS embedded project once.

**Test note**
I haven't add php unit tests since seems to me that the current tests already covers this changes. Of course, if you think that more tests are needed, I'll add them.

Funded by Faunalia

PR #4204, opened 2024-02-13
[Bugfix] Drag and drop form for embedded layers

mind84:
**Drag and drop** form configuration is not taken into account for the embedded layers. 

This is due to the fact that in the `readEditionForms` [method](https://github.com/3liz/lizmap-web-client/blob/4a63151845de41156ad7256d5c80d4536ac5dad7/lizmap/modules/lizmap/lib/Project/QgisProject.php#L890), the layer configuration is read from the "current" project and not from the embedded project.

To fix this, I did something similar to what I did for this [PR](https://github.com/3liz/lizmap-web-client/pull/4093). 

This time I've also add  a `cache check`. I don't know if it's worth it, since in most cases the modified methods are called before the project cache is set. However, if you think the cache-checking is fine, I could add it to the other methods modified in the previous PR.

I've added unit and playwright tests, since this case is not covered yet.

Thanks

Funded by Faunalia

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/QgisProject.php
+++ b/lizmap/modules/lizmap/lib/Project/QgisProject.php
@@ -856,6 +856,7 @@ class QgisProject
      */
     public function readEditionLayers($editionLayers)
     {
+        $embeddedEditionLayers = array();
         foreach ($editionLayers as $key => $obj) {
             // Improve performance by getting provider directly from config
             // Available for lizmap plugin >= 3.3.2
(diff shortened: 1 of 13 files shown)


Gustry:
Thanks for the fix.
I'm adding a backport 3.7 ?

mind84:
> Thanks for the fix. I'm adding a backport 3.7 ?

Yes thank you

rldhont (approved):

lizmap/modules/lizmap/lib/Project/QgisProject.php
```diff
@@ -886,8 +916,39 @@ public function readEditionLayers($editionLayers)
...
+
+                    $embeddedEditionLayers[$layerDefinition['projectPath']] = $qgisProject;
+                }
+
+                $xml = $qgisProject->getXml();
+            } else {
+                $xml = $this->xml;
+            }
+
+            $layerXml = $this->getXmlLayer2($xml, $obj->layerId);
```
mdouchin, line 951: I think you should create a new dedicated private function to extract this information and use it in `readEditionLayers` and `readEditionForms` methods. This will avoid to duplicate the code

mdouchin (changes requested):

Issue #4036, opened 2023-12-20
[Bug]: LWC 3.7.0 and QGIS 3.34 raster - symbology


josemvm:
### What is the bug?

qgis desktop 3.22.16
![imagem](https://github.com/3liz/lizmap-web-client/assets/10053874/cd21c64e-cc24-4c53-8b6f-30ea1933e392)
LWC 3.6.8
![imagem](https://github.com/3liz/lizmap-web-client/assets/10053874/c9e9b7dd-e50d-4eeb-b79a-7c10a56f5389)

===========
qgis desktop 3.28.13
![imagem](https://github.com/3liz/lizmap-web-client/assets/10053874/ac8c7e95-aef9-425e-9d69-8373984a0a2e)
LWC 3.7.0
![imagem](https://github.com/3liz/lizmap-web-client/assets/10053874/0ee2d0b5-333d-42b2-b7ab-33974964aabf)

### Steps to reproduce the issue

.

### Versions

**Versions :**

* Lizmap Web Client : 3.7.0
* Lizmap plugin : 4.1.1
* QGIS Desktop : 3.28.13
* QGIS Server : 3.28.13
* Py-QGIS-Server : not used
* QGIS Server plugin atlasprint : 3.3.2
* QGIS Server plugin lizmap_server : 2.8.1
* QGIS Server plugin wfsOutputExtension : 1.8.0

<details>
<summary>List of safeguards :</summary>
<br/>

* Mode : normal
* Allow parent folder : no
* Prevent other drive : no
* Prevent PG service : no
* Prevent PG Auth DB : no
* Force PG user&pass : no
* Prevent ECW : no
</details>

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 22.04

### Browsers

Firefox

### Browsers version

Firefox 120.0.1

josemvm:
in case of linear interpolation, not for discrete or exact interpolation

Gustry:
Similar to https://github.com/qgis/QGIS/issues/55651 ?

josemvm:
no, only in case of linear interpolation type not for discrete or exact interpolation type

Antoviscomi:
@josemvm in my LWC 3.7 the color ramp it's not rendered as in screenshot below
[...]

PR #4211, opened 2024-02-15
[Bugfix] WMS GetLegendGraphic for no symbols layer

rldhont:
Funded by Conseil Départemental du Calvados


josemvm:
@rldhont Is it expected that this solution will fix this problem https://github.com/3liz/lizmap-web-client/issues/4036 ?

rldhont:
@josemvm It will not fix #4036, it fixes the way GetLegendGraphic returns ERROR 500 if a layer has no style.

josemvm:
@rldhont thanks

PR #4215, opened 2024-02-16
[Feature] Improve snapping functionalities

mind84:
A bunch of changes on the snapping functionality:

- make the `snap layer` transparent, so that it does not overlap with the map layer symbology.
- disable snap on the hidden layers. If a layer is turned off,  then users can refresh the snap to disable snapping on that layer. 
- add a panel to select which layers to snap to. 
Below the snap button I added a panel with the list of snap layers configured for the layer being edited. 

Selection_014

Hidden layers are placed at the bottom of this list and users cannot act on them. When a layer is turned off from `layertree`, then this list is updated accordingly. 
Users can turn on/off the snap on visible layers. Then, as for previous point, users can refresh the snap to see the changes.

I'm not sure about this last point, I think it could be useful if there are a lot of snap layers configured. If you think so, I can go ahead with e2e tests, otherwise I'll roll back this last feature and only keep the first two.

You can check this functionality on the test project attached to this PR  (`form_edition_multilayer_snap.qgs`).

Thanks

Funded by Faunalia


rldhont (changes requested):
Thanks, only docstrings.

assets/src/components/Snapping.js
```diff
@@ -37,6 +37,20 @@ export default class Snapping extends HTMLElement {
                     </div>
                 </div>
             </div>
+            ${mainLizmap.snapping.active ?
+                html`<div class="control-group">
```
mdouchin, line 41: Please do not use the same level for the existing "Snapping" ( `<h3>` ) and the new "Snap on". For me, the latter belongs to the div Snapping. It is a new configuration option and should be put just under the buttons Stop & refresh, with no use of `h3` , and possibly in the same `<div>` as the existing Snapping group.

Please also add a `title` attribute to the new items to let the user read a tooltip by hovering the mouse

* Possible generic title when hovering the layer title (and the checkbox before): `Toggle snapping for this layer`
* For disabled layers, a precise enough sentence explaining why the layer is disabled and how to enable snapping for it

mdouchin (changes requested):
Thanks for this feature. I requested to add some text to help the user

Gustry:
I didn't check carefully your PR yet, sorry, I have noticed you have included a Playwright test with some snapping. Simulating mouse drag etc ?

I'm asking for curiosity this because some manual test before each release, one is snapping, because with Cypress, we couldn't integrate it in E2E at that time. CF [the current manual test](https://github.com/3liz/lizmap-web-client/blob/master/tests/qgis-projects/tests/form_edition_snap.md) CC @nboisteault

mind84:
> because with Cypress, we couldn't integrate it in E2E at that time

Hi!
I'm curios too because I hit the same problem and I didn't reach that UI interaction level. I'm testing the new panel behavior and the application XHR activities when user disable/enable layer on the layer tree or toggle the snap directly from the new panel (and refresh the snap, of course). 

I assume that if the geometry requests to the server are consistent with the user's interactions then the snap works as expected. Maybe not a strong assumption, but not so weak and IHMO acceptable.

So basically yes, it's a manual testing :)

I think I could spend some time to investigate if this can be done with playwright, but not sure about timing.

rldhont (approved):

Gustry:
> So basically yes, it's a manual testing :)

Ok, thanks for the explanations.

I'm going to test the PR.

Gustry:
On my screen, new options are quite on the right side, and I have an horizontal scroll bar which is not very good for UX, given the lot of space on the left side : 

image

I was also visiting the other snapping project : http://lizmap.local:8130/index.php/view/map?repository=testsrepository&project=form_edition_snap

I guess it's linked to the request from @mdouchin [above](https://github.com/3liz/lizmap-web-client/pull/4215#discussion_r1519481726) ?

mind84:
> On my screen, new options are quite on the right side, and I have an horizontal scroll bar which is not very good for UX, given the lot of space on the left side :

Yes, I've placed the new panel in the same control group of the buttons. Initially the center position looks good to me, but in the end I think I will place it left aligned

Thanks

mind84:
@Gustry 

now you should see the list in the right place

Thanks

Gustry:
Thanks, it's a better UX without any horizontal scrollbar like before : 

image

Gustry (approved):
From the UX

mdouchin (approved):

PR #4236, opened 2024-02-26
Provide a QGIS desktop version in metadata

Gustry:
Funded by 3Liz

### DIFF
--- a/lizmap/modules/lizmap/lib/Server/Server.php
+++ b/lizmap/modules/lizmap/lib/Server/Server.php
@@ -103,6 +103,19 @@ class Server
         return version_compare($currentVersion, $requiredVersion) < 0;
     }
 
+    /** Get the current recommended/required Lizmap desktop plugin.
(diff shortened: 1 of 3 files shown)


lizmap/modules/lizmap/lib/Server/Server.php
```diff
@@ -103,6 +103,15 @@ public function versionCompare($currentVersion, $requiredVersion)
         return version_compare($currentVersion, $requiredVersion) < 0;
     }
 
+    /** Get the current recommended/required Lizmap desktop plugin.
```
nworr, line 106: recommended or required ?

nworr (approved):
May be the method name should explictly tell if the version is recommended or required 

refering to the mainconfig section name , it's a required version

Gustry:
I have rephrased the docstring.
For now it's experimental about "required"/"recommended". I will see later.

PR #4242, opened 2024-02-27
locale for project warning can't be html, escaping it before display

nworr:
Ticket : # 

Funded by 3Liz


Gustry:
@rldhont approved

rldhont (approved):

PR #4226, opened 2024-02-20
Display a warning box when the CFG contains warnings

Gustry:
image

image

_Since the screenshot, it's now possible to close the dialog warning on the map_

Maybe it would be nice to cross info with "Administration panel → QGIS projects" ?

Funded by 3Liz

PR #4245, opened 2024-02-27
Fix Display a warning box when the CFG contains warnings

rldhont:
After #4226 and #4242, the warning message can generate the error `Uncaught SyntaxError: missing ) after argument list`
and `Uncaught ReferenceError: lizUrls is not defined`. These error messages are generated by a quote in the warning message.

To fix it, we use the Lizmap dictionnary which is a JSON version of the Dictionnary local.


Gustry:
unrelated test failure about playwright, it will be fixed in #4246

mdouchin (approved):

PR #4267, opened 2024-03-04
UX - Display the path name of the repository in the tooltip

Gustry:
I don't want the absolute path of the repository, but only the path set in the other admin panel.
~I didn't the find the correct property to use @nworr ?~

Just to be quicker to debug from the QGIS project panel

### DIFF
--- a/lizmap/modules/admin/zones/project_list.zone.php
+++ b/lizmap/modules/admin/zones/project_list.zone.php
@@ -149,6 +149,12 @@ class project_listZone extends jZone
      */
     private function getProjectListItem($inspectionDirectoryPath, $projectMetadata)
     {
+        $rootRepositories = lizmap::getServices()->getRootRepositories();
+        $repository = lizmap::getRepository($projectMetadata->getRepository())->getOriginalPath();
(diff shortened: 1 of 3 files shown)


mdouchin (approved):

lizmap/modules/admin/zones/project_list.zone.php
```diff
@@ -149,6 +149,15 @@ protected function _prepareTpl()
      */
     private function getProjectListItem($inspectionDirectoryPath, $projectMetadata)
     {
+        $services = lizmap::getServices();
+        $rootRepositories = $services->getRootRepositories();
+        $repository = lizmap::getRepository($projectMetadata->getRepository())->getOriginalPath();
+        if ($rootRepositories != '' && strpos($repository, $rootRepositories) === 0) {
+            $repository = str_replace($rootRepositories, '', $repository);
+            $repository = trim($repository, '/');
+            $repository .= '/';
```
nworr, line 158: if repository can only be a direct child of rootRepository , a dedicated php function fits

```suggestion
            $repository .= basename($repository);
```

Gustry:
Thanks @nworr 
Indeed, slightly better :)


image

PR #4269, opened 2024-03-05
Check QGIS desktop plugin version for edited QGS file

Gustry:
* For **edited** QGS file recently, we check if the plugin was up to date (less than a 2 months for instance)

This is to avoid a situation I got recently with QGIS Desktop 3.34 but a plugin from end of 2022.

To be discussed and polished

Funded by 3Liz

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/ProjectMetadata.php
+++ b/lizmap/modules/lizmap/lib/Project/ProjectMetadata.php
@@ -41,6 +41,7 @@ class ProjectMetadata
             'map' => $project->getRelativeQgisPath(),
             'acl' => $project->checkAcl(),
             'qgisProjectVersion' => $project->getQgisProjectVersion(),
+            'lastSaveDateTime' => $project->getLastSaveDateTime(),
             'lizmapPluginVersion' => $project->getLizmapPluginVersion(),
             'lizmapWebClientTargetVersion' => $project->getLizmapWebCLientTargetVersion(),
             'needsUpdateError' => $project->needsUpdateError(),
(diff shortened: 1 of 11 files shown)


lizmap/modules/lizmap/lib/Project/ProjectMetadata.php
```diff
@@ -211,6 +222,56 @@ public function getLizmapWebCLientTargetVersion()
...
+     * The check is not done only if the project has been edited recently compare to the date of the Lizmap Web Client.
+     *
+     * @return bool If the plugin should be updated in QGIS Desktop
+     */
+    public function updateQgisLizmapPlugin()
+    {
+        // Not the best choice, maybe because of git for dev, dates are not correct
+        // $projectDate = date($this->getFileTime());
+        // Better to use lastSaveDateTime in the QGS file
+        $projectDate = strtotime($this->getLastSaveDateTime());
```
mdouchin, line 236: Perhaps add here a fallback to `getFileTime` if `getLastSaveDateTime` returns no data ?

lizmap/modules/lizmap/lib/Project/QgisProject.php
```diff
@@ -1259,6 +1278,21 @@ protected function readWMSInformation($qgsLoad)
...
+     *
+     * @return string
+     */
+    protected function readLastSaveDateTime($xml)
+    {
+        $qgisRoot = $xml->xpath('//qgis');
+        $qgisRootZero = $qgisRoot[0];
+
+        return (string) $qgisRootZero->attributes()->saveDateTime;
+    }
```
mdouchin, line 1294: This would consume a lot of RAM, we need to check how to solve this. Let-s check other functions in the same file too.

mdouchin (changes requested):

Gustry:
Ok, thanks for the review.

I will come back very soon with an alternative. I keep this branch as an archive for now.

Gustry:
After some discussions, reopening this PR.

Gustry:
Archiving this PR, I made #4287 instead for now.
We can come back later with this PR about a better edited date detection.

PR #4278, opened 2024-03-08
[Bugfix] Apply min and max resolutions to base layers

rldhont:
Now it is possible to define base layers in the project, so it is possible de defined min and max scale for these layers.

Funded by Terre de Provence Agglomération

### DIFF
--- a/assets/src/modules/map.js
+++ b/assets/src/modules/map.js
@@ -337,13 +337,33 @@ export default class map extends olMap {
             0
         ) - 1;
 
+        const proj3857 = getProjection('EPSG:3857');
+        const max3857Resolution = getWidth(proj3857.getExtent()) / 256;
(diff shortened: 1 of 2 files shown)


mdouchin (approved):

assets/src/modules/map.js
```diff
@@ -342,8 +342,28 @@ export default class map extends olMap {
 
         for (const baseLayerState of mainLizmap.state.baseLayers.getBaseLayers()) {
             let baseLayer;
+            let layerMinResolution = undefined;
+            let layerMaxResolution = undefined;
```
nboisteault, line 346:
```suggestion
            let layerMinResolution;
            let layerMaxResolution;
```

rldhont: Thanks

PR #4287, opened 2024-03-12
Check plugin version for recently edited QGS file

Gustry:
Alternative approach for #4269 : not using xpath, but the date in the file metadata.

But for developers, Git is not making dates correct. We can come back later for another approach (reading the first X lines in the QGS file or using xpath carefully like in #4269)

For recently edited QGS file (edited after the release date of  LWC), we check against an hardcoded plugin version.

For now, the warning is only displayed for admin (when connected) on the map, like #4270

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/ProjectMetadata.php
+++ b/lizmap/modules/lizmap/lib/Project/ProjectMetadata.php
@@ -212,6 +212,57 @@ class ProjectMetadata
     }
 
     /**
+     * If the QGIS desktop needs an update of the plugin.
+     * The check is done only if the project has been edited recently compare to the date of the Lizmap Web Client.
+     *
+     * @return bool If the plugin should be updated in QGIS Desktop
+     */
+    public function qgisLizmapPluginUpdateNeeded()
+    {
+        $projectInfos = \Jelix\Core\Infos\AppInfos::load();
+        $releaseVersion = $projectInfos->version;
+        // Version of the current application Lizmap Web Client
+        if (substr_count($releaseVersion, '-') == 1 && substr_count($releaseVersion, '.') == 2) {
+            // One dash and two dots
+            // We are in dev mode : version is like "3.8.0-pre" for instance
+            // Dates on QGS files are not totally correct, due to git changing the creation date when changing branch
+            // We do not process this check for now
+            return false;
+        }
+        // Version is now either "3.6.11-pre.6402" → non official package
+        // Or stable release "3.6.11" → official package
+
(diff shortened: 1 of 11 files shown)


mdouchin (approved):
Big up

lizmap/modules/lizmap/lib/Project/ProjectMetadata.php
```diff
@@ -211,6 +211,57 @@ public function getLizmapWebCLientTargetVersion()
...
+            return false;
+        }
+
+        // Project file is newer than the release date
+        // We check against the hard coded version
+        $recommendedVersion = \jApp::config()->minimumRequiredVersion['lizmapDesktopPlugin'];
+
+        $projectVersion = $this->getLizmapPluginVersion();
+        if (!is_numeric($projectVersion)) {
+            // It shouldn't happen to much, but before the version was like "master".
```
nworr, line 248: Is it for old plugin that return version as human readable string ?
should'nt return true ?

lizmap/modules/lizmap/lib/Project/ProjectMetadata.php
```diff
@@ -211,6 +211,57 @@ public function getLizmapWebCLientTargetVersion()
...
+            return false;
+        }
+
+        // TODO It would be nice to avoid code duplication,
+        // we have code to manage versions for sorting, display, integer comparaison...
+        $intVersion6Digit = (strlen($projectVersion) == 6 ? $projectVersion : '0'.$projectVersion);
```
nworr, line 254: Alternative : store `lizmapDesktopPlugin `as int , type will match with projectVersion and native int comparison can be performed

?

lizmap/modules/lizmap/lib/Project/ProjectMetadata.php
Gustry: I will rephrase.
`master` is still written in the git repository : https://github.com/3liz/lizmap-plugin/blob/master/lizmap/metadata.txt#L17
It's rare, but it can happen if we are running QGIS in a docker container for instance (GIT on the host to fetch the git repository and then running QGIS desktop in a docker container (so without git), the version will be `master` in the CFG file. I will keep this in mind.

Gustry:
@mdouchin I'm not convinced by this approach, I guess we need to get back to #4269 using xpath. For instance yesterday I redeployed all tests projects on our test servers by doing a "big" copy/paste of all projects. I didn't touch these projects on my local, but now LWC thinks the reverse ...

I made ordering on purpose : 
image

PR #4298, opened 2024-03-13
Use the date stored in the QGS file instead of the file creation date

Gustry:
The PR #4287 is not perfect, in case of bulk upload of files.
We can notice on our test server right now. I updated all tests files on the server by doing a "big" copy/paste of all files and now LWC thinks I have touched all QGS files, which is not true.

I assume we should find `saveDateTime` in the first 5 lines of XML.

```xml
<qgis projectname="" saveDateTime="2023-09-07T13:56:00" saveUser="XXXXX" saveUserFull="XXXXXX" version="3.22.16-Białowieża">
```

Backport will be done manually in #4296 and #4297. No backport label on purpose.

Keen to have feedbacks from @mdouchin (and @nworr )

### DIFF
--- a/lizmap/modules/admin/zones/project_list.zone.php
+++ b/lizmap/modules/admin/zones/project_list.zone.php
@@ -319,6 +319,10 @@ class project_listZone extends jZone
      */
     private function pluginIntVersionToSortableString(string $intVersion): string
     {
+        if ($intVersion == 'master' || $intVersion == 'dev') {
+            return '00.00.00';
+        }
+
         // in some old plugin the version is already human readable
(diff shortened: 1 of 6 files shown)


nworr (commented):
As far as i understand, the ProjectCache Objet will store `lastSaveDateTime`
if the qgis file is outdated , the whole file will be read, and  `lastSaveDateTime` can be extracted using xpath

Gustry:
As we discussed just now, we are going to merge it.
The solution with XPath was done earlier in #4269 but @mdouchin was mentioning some performance issue (for landing page and admin panel).

Thanks for the review, let's try it.

PR #4299, opened 2024-03-13
[Bugfix] Printing base layers withlayer in hidden group

rldhont:
To keep compatibility with lizmap base layers config for version before 3.7, the hidden group can provide config for external baselayers.


mdouchin (approved):

Gustry:
This CFG test case included must stay in 3.6 format, right ? (it mustn't be updated to 3.7, yes?)

PR #4300, opened 2024-03-13
The layer definition extracted from XML could be null

rldhont:
The layer definition extracted from XML and identified by the layer id could be null.


laurentj (approved):

Gustry (approved):

rldhont:
No need to backport to release_3_6

PR #4305, opened 2024-03-14
[Bugfix] Webdav url replacement in popup childrens compact table

mind84:
Added support for the webdav url replacement in the child compact table when children are displayed in popup.

Updated tests accordingly.

Funded by Faunalia


mind84:
Seems to me that `WMTS` and `permalink` Playwright tests are failing randomly, is that possible? Often, after simply rebasing, they run successfully.

rldhont:
> Seems to me that `WMTS` and `permalink` Playwright tests are failing randomly, is that possible? Often, after simply rebasing, they run successfully.

it is possible

rldhont (approved):

PR #4313, opened 2024-03-19
Remove legacy switcher-layers code

nboisteault:
Funded by 3Liz


nboisteault:
@mdouchin is this file still used ? https://github.com/3liz/lizmap-web-client/blob/master/lizmap/www/assets/css/admin_lizpoi.css

PR #4319, opened 2024-03-21
Use `default_background_color_index`

nboisteault:
Linked to https://github.com/3liz/lizmap-plugin/pull/574

Funded by 3Liz

### DIFF
--- a/assets/src/modules/config/BaseLayer.js
+++ b/assets/src/modules/config/BaseLayer.js
@@ -924,6 +924,21 @@ export class BaseLayersConfig {
             }
         }
 
+        // Add base layer as project default background color
+        const default_background_color_index = options.hasOwnProperty('default_background_color_index') ? options.default_background_color_index : -1;
+        if (isNaN(default_background_color_index)) {
+            throw new ValidationError('The default_background_color_index parameter is not a number!');
+        }
+        if (default_background_color_index !== -1
+            && names.indexOf('empty') == -1
+            && names.indexOf('project-background-color') == -1) {
+            if (default_background_color_index < names.length) {
+                names.splice(default_background_color_index, 0, 'project-background-color');
+            } else {
+                names.push('project-background-color');
+            }
+        }
+
         this._names = [];
         this._configs = [];
         for (const key of names) {
--- a/tests/qgis-projects/tests/base_layers_user_defined.qgs.cfg
+++ b/tests/qgis-projects/tests/base_layers_user_defined.qgs.cfg
@@ -1,8 +1,8 @@
 {
     "metadata": {
         "qgis_desktop_version": 32815,
-        "lizmap_plugin_version_str": "4.2.7-alpha",
-        "lizmap_plugin_version": 40207,
+        "lizmap_plugin_version_str": "4.2.8-alpha",
+        "lizmap_plugin_version": 40208,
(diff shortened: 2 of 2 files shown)


assets/src/modules/config/BaseLayer.js
```diff
@@ -926,6 +935,7 @@ export class BaseLayersConfig {
```
rldhont, line 935: I propose to do so instead of your proposed code.
```suggestion
        const default_background_color_index = options.hasOwnProperty('default_background_color_index') ? options.default_background_color_index : -1;
        if (isNaN(default_background_color_index)) {
            throw new ValidationError('The default_background_color_index parameter is not a number!');
        }
        if (default_background_color_index !== -1) {
            if (default_background_color_index < names.length) {
                names.splice(default_background_color_index, 0, 'project-background-color');
            } else {
                names.push('project-background-color');
            }
        }
```

rldhont (changes requested):

Gustry:
Thanks for the fix.

Just to be sure, with QGIS < 3.34, it doesn't insert twice right ? Because the layer will be in the GetCapabilities.

assets/src/modules/config/BaseLayer.js
```diff
@@ -924,6 +924,19 @@ export class BaseLayersConfig {
             }
         }
 
+        // Add base layer as project default background color
+        const default_background_color_index = options.hasOwnProperty('default_background_color_index') ? options.default_background_color_index : -1;
+        if (isNaN(default_background_color_index)) {
+            throw new ValidationError('The default_background_color_index parameter is not a number!');
+        }
+        if (default_background_color_index !== -1) {
```
rldhont, line 932: Linked to @Gustry comment
```suggestion
        if (default_background_color_index !== -1
            && names.indexOf('empty') == -1
            && names.indexOf('project-background-color') == -1) {
```

rldhont (changes requested):

nboisteault:
@Gustry with the `base_layers_user_defined` project, `default_background_color_index` is equal to `4` instead of `2` because there is two empty groups which should not be counted.

Gustry:
in QGIS, it's correct, it's `4` :)

So you want a "special" index, not counting empty groups ? (similar to the server side).
Looking quickly

Gustry:
Ok, @nboisteault fixed in latest commit, please pull, it should be fine.

Just in case, in case of bug (or filtered layers etc), always check that if `default_background_color_index >= 0`, it has been inserted, at least at the end in the worst sitution. Maybe due to invalid layers in QGIS server.

nboisteault:
> Ok, @nboisteault fixed in latest commit, please pull, it should be fine.
> 
> Just in case, in case of bug (or filtered layers etc), always check that if `default_background_color_index >= 0`, it has been inserted, at least at the end in the worst sitution. Maybe due to invalid layers in QGIS server.

Yes it is done here : https://github.com/3liz/lizmap-web-client/pull/4319/files#diff-090101e050d6a2279c7421cc38a96ec441d18132aead9cf71d0c53dd0a38e932R938

Gustry:
Side note related, there a few tests failing for missing `project-background-color` in GetCapabilities
when we switch to QGIS server 3.34 in the docker test stack.

Tests in #4323

PR #4323, opened 2024-03-21
Tests - Switch to QGIS 3.34 in Makefile

Gustry:
Not sure if we want to merge this, but it's to check which one failing.

Related to #4319


Gustry:
Closing in favor of #4367

PR #4329, opened 2024-03-26
Speed up FCP (first contentful paint)

nboisteault:
Speed up FCP (first contentful paint). This PR allows JavaScript scripts to be loaded after the HTML is loaded and parsed thanks to `defer` attribute. It is interesting for users with a slow connection, this way they will not wait too long with a blank page.

Funded by 3Liz

TODO:
- [x] https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/view/controllers/lizMap.classic.php#L489 ? Is it documented?

### DIFF
--- a/assets/src/index.js
+++ b/assets/src/index.js
@@ -51,6 +51,7 @@ lizMap.events.on({
         lizMap.mainLizmap = mainLizmap;
         lizMap.mainEventDispatcher = mainEventDispatcher;
 
-        executeJSFromServer();
     }
 });
+
+executeJSFromServer();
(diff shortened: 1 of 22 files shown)


rldhont:
@nboisteault can you add some description to this PR to explain how your speed up lizmap loading ?

rldhont (changes requested):
Describe changes

nboisteault:
> Describe changes

Done.

rldhont (approved):
Thanks

laurentj (approved):

PR #4333, opened 2024-03-28
Move inline JS scripts

nboisteault:
Funded by 3Liz

### DIFF
--- a/assets/src/legacy/map.js
+++ b/assets/src/legacy/map.js
@@ -4005,12 +4005,24 @@ window.lizMap = function() {
             return deactivateMaplayerFilter(layername);
         },
 
-
-
         /**
          * Method: init
          */
         init: function() {
+            // Initialize global variables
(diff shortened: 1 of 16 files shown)


mdouchin (approved):
🚀

assets/src/legacy/map.js
```diff
@@ -4491,6 +4491,10 @@ $(document).ready(function () {
         .children('.ui-dialog-titlebar').removeClass('ui-corner-all');
     // configurate OpenLayers
     OpenLayers.DOTS_PER_INCH = 96;
+    const lizmapVariables = JSON.parse(document.getElementById('lizmap-vars').innerText);
```
laurentj, line 4494: probably we should catch potential exception in the case of a bad JSON structure? And check if it is not empty?

laurentj (approved):

rldhont (approved):

Issue #3408, opened 2023-01-11
[Bug]: Double toolbar visible in the popup


Gustry:
### What is the bug?

There is a double toolbar visible when clicking on a cat.

![image](https://user-images.githubusercontent.com/1609292/211830452-a2992ed2-031e-4e0e-8a56-0fd178b97d50.png)

CC @mdouchin @nboisteault 

### Steps to reproduce the issue

* Visit https://demo.lizmap.com/lizmap_3_6/index.php/view/map?repository=features&project=cats
* Click on a cat, got the popup with **single** toolbar
* Click on another cat (or even the same cat), with a **double** toolbar
* Check if login is triggering the issue ?

### Versions

LWC 3.6

### Check Lizmap plugin

- [ ] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### QGIS server version, only if the section above doesn't mention the QGIS Server version

3.22

### Operating system

Ubuntu

### Browsers

Firefox

### Browsers version

Latest snap

nboisteault:
I can reproduce when clicking twice not once.

nboisteault:
I can't reproduce [here](https://demo.snap.lizmap.com/lizmap_3_7/index.php/view/map?repository=testsrepository&project=feature_toolbar) and [here](https://demo.snap.lizmap.com/lizmap_3_6/index.php/view/map?repository=testsrepository&project=feature_toolbar).
That might be a lol cat bug.

Gustry:
Maybe depending of buttons available in the toolbar ?

Indeed, I can't reproduce on https://demo.snap.lizmap.com/lizmap_3_7/index.php/view/map?repository=testsrepository&project=form_edit_related_child_data

I'm fine to close unfortunately ... :)

Gustry:
I can reproduce again this bug. Still on https://demo.lizmap.com/lizmap_3_6/index.php/view/map?repository=features&project=cats

Still not everytime, but quite a few times.

nworr:
It seems that the featuretoolbar is doubled only when the popup panel is active and with content

can it be relate to the selection.changed event ? 
https://github.com/3liz/lizmap-web-client/blob/35acbd7c458b1cdbd4815b438524d253ee97f4e0/assets/src/components/FeatureToolbar.js#L124

Gustry:
Should be fixed with #4346

PR #4346, opened 2024-04-04
[Bugfix] Dataviz in popup generate two feature toolbar in parent popup

mind84:
When a `dataviz` plot is displayed as child in a parent feature, if the user reopen the same popup or open a new popup on another feature then the `feature-toolbar` of the parent layer is duplicated.

Updated tests.

Test project to reproduce the issue:

[popup_bar..zip](https://github.com/3liz/lizmap-web-client/files/14873944/popup_bar.zip)

Funded by Faunalia


Gustry:
@mind84 I just edited the comment FYI about existing project for this test case.

Thanks anyway

mind84:
@Gustry
Thanks for all the pointers!

I checked the levels of the bakeries in Montepellier, apparently there is no suitable project for this case :/

Anyway, I changed the layer sources to `geojson` instead of `geopackage`

> I guess that fixes https://github.com/3liz/lizmap-web-client/issues/3408?

I think so. I didn't notice there was already an open ticket on this.

Thank you!

rldhont (approved):
Thanks @mind84

Issue #4192, opened 2024-02-12
[Bug]: Bing Satellite added with QMS plugin do not show as baselayer


gioman:
### What is the bug?

Adding "Bing Satellite" with Quick Map Services plugin does work as a normal layers, but does not show if added in the Baselayers group.

I had already observed something similar here

https://github.com/3liz/lizmap-web-client/issues/4030#issuecomment-1858880634

and now it is partially solved.

### Steps to reproduce the issue

See above.

### Versions, safeguards, check summary etc

LMWC 3.7.3 and QGIS Desktop/Server 3.28.15 LM Plugin 4.2.0

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 22.04

### Browsers

Firefox, Chrome

### Browsers version

Latest updates

### Relevant log output

```shell
Nothing relevant.
```

Gustry:
Did you set a API key ?

gioman:
> Did you set a API key ?

@Gustry I have, but anyway this question is baffling me. That layer (and as well Google Satellite) do not need any API key to show in QGIS when added with QMS. And in fact Google Satellite works well in LMWC as normal layer and as baselayer. Bing does not work inly a baselayer. What am I missing?

Gustry:
Can you check your HTTP requests status for Bing tiles ?

aeduard:
With the newest version of the lizmap plugin (4.2.0) Google Hybrid does not work with the API key set. Strangely Google Satellite works. I don't have bing maps api key to test.

gioman:
> Can you check your HTTP requests status for Bing tiles ?

@Gustry This is what I'm observing:

* layers added with QMS (i.e. Google/Bing Satellite) outside the _baselayers_ group, are always requested as WMS via QGIS Server (as expected), regardless if the "Gets Images Directly from WMS Server" checkbox is enabled (less expected, as it happens also with WMS layers for remote services that are not provided by QMS but rather a standard WMS connection)

* when some layers are moved into de "baselayers" group (definitely  Google/Bing Satellite added with QMS) these are automatically requested from the original tile servers. i.e.

`https://ecn.t3.tiles.virtualearth.net/tiles/a%7Bq%7D.jpeg?g=0&dir=dir_n%27`

[...]

Issue #4029, opened 2023-12-16
LMWC 3.7.0: adding the project-background-color group with the proper button in LM plugin and leaving it empty, causes the project to stop to load


gioman:
### What is the bug?

![image](https://github.com/3liz/lizmap-web-client/assets/1951107/5cbc0f75-5a54-4058-8e58-47c38eb5e57e)

### Steps to reproduce the issue

See title. Test project here: 

[dorset.zip](https://github.com/3liz/lizmap-web-client/files/13692628/dorset.zip)

### Versions

LMWC 3.7.0

On Ubuntu 22.04 with QGIS Server 3.28.12, Project saved on QGIS 3.28.12 ans LM plugin 4.1.1

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 22.04

### Browsers

Firefox, Chrome

### Browsers version

Latest updates

Gustry:
I think it's linked due to old version of the plugin.
You have a difference of layer count between your QGS file and CFG file.

gioman:
> I think it's linked due to old version of the plugin.
> You have a difference of layer count between your QGS file and CFG file.

I referred the wrong plugin version in the description (now changed) and for good measure I tried again now (with LM plugin 4.1.1), same result.

Gustry:
I didn't replicate this issue exactly.
The `project-background-color` is missing from your test project. I have added it, it's working as expected about this. (I met another error, not linked to this)

Are you using a some legacy checkbox about base layers AND the new `baselayers` group ?
So you were mixing both configurations ?
Was your CFG file updated ? (as always in LWC, the CFG file must have the same layer tree as in QGIS) No warning when saving about a different count of layers ?

Gustry:
> leaving it empty

Well, it must stay empty anyway, as it's just to add the "default background color" only

Gustry:
Please make "F12" and check the console, there might be a clue about the error it self. ("console" or "request" tab).

gioman:
It happens even having just the "baselayers" group empty.

Gustry:
If "baselayers" is empty, there isn't any need for the group. I guess it must have at least one item in it.

I will make a patch in the plugin to block saving if "baselayers" is empty : https://github.com/3liz/lizmap-plugin/issues/540

Can we close here ?

gioman:
[...]

PR #4357, opened 2024-04-10
[Feature] Add QMS Google Maps Tiles

mind84:
Since release 9.0, OpenLayers has introduced support form Google Maps Tiles.

This PR relies basically on the same logic of this one https://github.com/3liz/lizmap-web-client/pull/4150

If a valid `Google Api Key` is provided in the Lizmap plugin, then the application uses the OL dedicated class to get tiles from the Google endpoint and the API key is used to issue requests.

If no key is provided, then these tiles are treated as a `generic WMTS external layer` and the application get the tiles directly from the default URI set by QMS QGIS plugin. In this case it is not guaranteed that the tiles will work.

Maps supported:
- Satellite
- Roads

**TEST NOTES**

For now testing is manually. In the test repository there is a `google_basemap.qgs` project. You can configure the Lizmap plugin by adding your own `Google Api Key` and then check the functionality.

Funded by Faunalia

### DIFF
--- a/assets/src/modules/map.js
+++ b/assets/src/modules/map.js
@@ -24,6 +24,7 @@ import TileGrid from 'ol/tilegrid/TileGrid.js';
 import TileWMS from 'ol/source/TileWMS.js';
 import XYZ from 'ol/source/XYZ.js';
 import BingMaps from 'ol/source/BingMaps.js';
+import Google from 'ol/source/Google.js';
 import {BaseLayer as LayerBase} from 'ol/layer/Base.js';
 import LayerGroup from 'ol/layer/Group.js';
 import { Vector as VectorSource } from 'ol/source.js';
@@ -446,6 +447,16 @@ export default class map extends olMap {
                     // use maxZoom 19 to see stretched tiles instead of the BingMaps
                     // "no photos at this zoom level" tiles
                     // maxZoom: 19
+                    }),
+                });
+            } else if (baseLayerState.type === BaseLayerTypes.Google) {
+                baseLayer = new TileLayer({
+                    minResolution: layerMinResolution,
+                    maxResolution: layerMaxResolution,
+                    preload: Infinity,
+                    source: new Google({
+                        key: baseLayerState.key,
+                        mapType: baseLayerState.mapType,
                     }),
                 });
             } else if (baseLayerState.type === BaseLayerTypes.Lizmap) {
--- a/tests/end2end/playwright/google-basemap.spec.js
+++ b/tests/end2end/playwright/google-basemap.spec.js
@@ -1,0 +1,39 @@
+
+const { test, expect } = require('@playwright/test');
+const { gotoMap } = require('./globals');
+
+test.describe('Google Maps Baselayers', () => {
+    test('Load map with no API Key', async ({ page }) => {
+        const url = '/index.php/view/map?repository=testsrepository&project=google_basemap';
+        await gotoMap(url, page);
+
+        // baselayers group should be hidden since it is empty due to the default STRICT_GOOGLE_TOS_CHECK env variable value = TRUE
+        await expect(page.locator('#switcher-baselayer.hide')).toHaveCount(1);
+
+    });
+
+    test('Load map with dummy API Key', async ({ page }) => {
+        // listen to the requests to intercept failing ones
+        let initGoogleRequestsCount = 0;
+        page.on('response', response => {
+            if(response.url().includes('createSession?key=dummy') && response.status() != 200) {
+                initGoogleRequestsCount++;
+            }
+        });
+
+        const url = '/index.php/view/map?repository=testsrepository&project=google_apikey_basemap';
+        await gotoMap(url, page);
+
+        // there are three Google base layers in the project, so the expected number of failing requests is three
+        expect(initGoogleRequestsCount).toBe(3);
+        // baselayers group should be visible...
+        await expect(page.locator('#switcher-baselayer')).toBeVisible();
+
+        //.. and should contains the three Google base layers (not loaded)
+        let options = page.locator('#switcher-baselayer').getByRole('combobox').locator('option');
+        await expect(options).toHaveCount(3);
+        expect(await options.nth(0).getAttribute('value')).toBe('Google Streets');
+        expect(await options.nth(1).getAttribute('value')).toBe('Google Satellite');
+        expect(await options.nth(2).getAttribute('value')).toBe('Google Hybrid');
+    });
+});
--- a/assets/src/modules/config/BaseLayer.js
+++ b/assets/src/modules/config/BaseLayer.js
@@ -31,6 +31,7 @@ export const BaseLayerTypes = createEnum({
     'WMTS': 'wmts',
     'WMS': 'wms',
     'Lizmap': 'lizmap',
+    'Google': 'google',
 });
 
 /**
@@ -305,6 +306,52 @@ export class BingBaseLayerConfig extends BaseLayerConfig {
      */
     get imagerySet() {
         return this._imagerySet;
+    }
+
+}
+
+const googleProperties = {
+    'title': { type: 'string' },
+    'mapType': { type: 'string' }
+}
+
+const googleOptionalProperties = {
+    'key': { type: 'string', nullable: true }
+}
+
+/**
+ * Class representing a Google base layer config
+ * @class
+ * @augments BaseLayerConfig
+ */
+export class GoogleBaseLayerConfig extends BaseLayerConfig {
+    /**
+     * Create a GOOGLE base layer config based on a config object
+     * @param {string} name           - the base layer name
+     * @param {object} cfg            - the lizmap config object for GOOGLE base layer
+     * @param {string} cfg.title      - the base layer title
+     * @param {string} cfg.mapType    - the base layer mapType
+     * @param {string} [cfg.key]      - the base layer key
+     */
+    constructor(name, cfg) {
+        if (!cfg || typeof cfg !== "object") {
+            throw new ValidationError('The cfg parameter is not an Object!');
+        }
+
+        if (Object.getOwnPropertyNames(cfg).length == 0) {
+            throw new ValidationError('The cfg parameter is empty!');
+        }
+
+        super(name, cfg, googleProperties, googleOptionalProperties)
+        this._type = BaseLayerTypes.Google;
+    }
+
+    /**
+     * The Google mapType
+     * @type {string}
+     */
+    get mapType() {
+        return this._mapType;
     }
 
 }
@@ -745,6 +792,18 @@ const QMSExternalLayer = {
         "imagerySet": "Aerial",
         "key": "",
     },
+    "google-streets": {
+        "type" :"google",
+        "title": "Google Streets",
+        "mapType": "roadmap",
+        "key":""
+    },
+    "google-satellite": {
+        "type" :"google",
+        "title": "Google Satellite",
+        "mapType": "satellite",
+        "key":""
+    }
 }
 
 /**
@@ -839,7 +898,20 @@ export class BaseLayersConfig {
                                 }
                                 // add the apikey to the configuration
                                 Object.assign(extendedCfg[layerTreeItem.name],{key:options["bingKey"]})
-                            } else {
+                            } else if (externalUrl && externalUrl.includes('google.com') && options["googleKey"]){
+                                if (externalUrl.includes('lyrs=m')) {
+                                    // roads
+                                    extendedCfg[layerTreeItem.name] = structuredClone(QMSExternalLayer["google-streets"])
+                                } else if (externalUrl.includes('lyrs=s')){
+                                    // fallback on satellite map
+                                    extendedCfg[layerTreeItem.name] = structuredClone(QMSExternalLayer["google-satellite"])
+                                } else {
+                                    extendedCfg[layerTreeItem.name] = structuredClone(QMSExternalLayer["google-streets"])
+                                }
+                                // add the apikey to the configuration
+                                Object.assign(extendedCfg[layerTreeItem.name],{key:options["googleKey"]})
+                            }
+                            else {
                                 // layer could be converted to XYZ or WMTS background layers
                                 extendedCfg[layerTreeItem.name] = structuredClone(layerTreeItem.layerConfig.externalAccess);
                             }
@@ -993,6 +1065,10 @@ export class BaseLayersConfig {
                     break;
                 case BaseLayerTypes.Bing:
                     this._configs.push(new BingBaseLayerConfig(key, blCfg));
+                    this._names.push(key);
+                    break;
+                case BaseLayerTypes.Google:
+                    this._configs.push(new GoogleBaseLayerConfig(key, blCfg));
                     this._names.push(key);
                     break;
                 case BaseLayerTypes.WMTS:
--- a/assets/src/modules/config/LayerTree.js
+++ b/assets/src/modules/config/LayerTree.js
@@ -420,7 +420,11 @@ function buildLayerTreeGroupConfigItems(wmsCapaLayerGroup, layersCfg, level) {
             const groupItems = buildLayerTreeGroupConfigItems(wmsCapaLayer, layersCfg, level+1);
             items.push(new LayerTreeGroupConfig(cfg.name, level+1, groupItems, wmsCapaLayer, cfg));
         } else {
-            items.push(new LayerTreeLayerConfig(cfg.name, level+1, wmsCapaLayer, cfg));
+            // avoid to add the baseLayers group to the map if doesn't contains any layer.
+            if(wmsName.toLowerCase() != 'baselayers') {
+                items.push(new LayerTreeLayerConfig(cfg.name, level+1, wmsCapaLayer, cfg));
+            }
+
         }
     }
     return items;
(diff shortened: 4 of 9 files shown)


assets/src/modules/map.js
```diff
@@ -426,6 +427,19 @@ export default class map extends olMap {
...
                 });
+            } else if (baseLayerState.type === BaseLayerTypes.Google) {
+                baseLayer = new TileLayer({
+                    minResolution: layerMinResolution,
+                    maxResolution: layerMaxResolution,
+                    preload: Infinity,
+                    source: new Google({
+                        key: baseLayerState.key,
+                        mapType: baseLayerState.mapType,
+                    // use maxZoom 19 to see stretched tiles instead of the BingMaps
```
Gustry, line 438: Bingmaps ?
Maybe not directly linked to your PR, but when I see your comment, I'm wondering what happen when in zoom in too much, below what Google is providing ? Does it become blank or pixelated satellite imagery ?

Gustry (commented):
Nice addition.
So it works if you use QMS AND you set a API key in the plugin right ?
Sorry, I see you description, we will have to warn users about some corner cases. Honestly, for me using Google without an API key is discouraged as it's against Google TOS.

I'm wondering if there should some kind of automatic attribution done ? AS we know there is a Google layer, maybe the attribution on the bottom right corner should be automatically added ? Maybe it's done already ? I'm not sure what the Google TOS about this ?

Anyway, thanks for you work.

Gustry:
Can you rename your commit, like a similar title as your PR ? it will make the git history better.

mind84:
> So it works if you use QMS AND you set a API key in the plugin right ?
Sorry, I see you description, we will have to warn users about some corner cases. Honestly, for me using Google without an API key is discouraged as it's against Google TOS.

Yes, as BingMaps. If you don't specify any KEY in Lizmap plugin, then the `basemap` will be loaded as generic external layer.

Do you think this behavior should be prevented and we should raise an error in this case? IMHO it's up to the user to make this choice

>Maybe it's done already ?

Yes, it is already done

Screenshot_20240415_120154


Google Maps TOS is quite restrictive and some aspect are rather unclear IMHO.  For example, take this paragraph extracted from https://cloud.google.com/maps-platform/terms:

_(e) No Use With Non-Google Maps. To avoid quality issues and/or brand confusion, Customer will not use the Google Maps Core Services with or near a non-Google Map in a Customer Application. For example, Customer will not (i) display or use Places content on a non-Google map, (ii) display Street View imagery and non-Google maps on the same screen, or (iii) link a Google Map to non-Google Maps content or a non-Google map._

I don't fully understand this point: there are a lot of cases (over 90% I guess?) where "Customer **WILL USE** the Google Maps Core Services with or near a non-Google Map" and they also paying to use these API. If this is not allowed, I'm wondering about the utility of integrate a Google Maps in any Costumer Application. My idea is that Google wants to avoid being associated to any improper use of their maps and to do so it has inserted a general clause and then possibly enforced it in case there were disputes of some kind. I don't think that Lizmap fits into this case.

assets/src/modules/map.js
mind84: > Bingmaps ?

typo, sorry

> I'm wondering what happen when in zoom in too much

I try to load the project using  the `Use native scales according to the project CRS (better display)` option in the Lizmap plugin with min scale = 1, the min displaying scale provided by google in EPSG:3857 is 1:141.

Screenshot_20240415_121220

Under this scale the map becomes pixelated.

rldhont (approved):

Gustry:
> Do you think this behavior should be prevented and we should raise an error in this case? IMHO it's up to the user to make this choice

Maybe a new employee in an organization is not aware about publication rule and push a project with a Google layer and/or the Lizmap system administrator (provider) want to be clear about TOS used on his own server.

Would it be possible to have a config in the `ini` file about this ?

In `lizmap/var/config/profiles.ini.php.dist`
```ini
[tosCheckExternalTiles:default]
;google=yes
;bing=no
```

mdouchin:
> > Do you think this behavior should be prevented and we should raise an error in this case? IMHO it's up to the user to make this choice
> 
> Maybe a new employee in an organization is not aware about publication rule and push a project with a Google layer and/or the Lizmap system administrator (provider) want to be clear about TOS used on his own server.
> 
> Would it be possible to have a config in the `ini` file about this ?
> 
> In `lizmap/var/config/profiles.ini.php.dist`
> 
> ```ini
> [tosCheckExternalTiles:default]
> ;google=yes
> ;bing=no
> ```

Ok for adding a configuration, but instead of `profiles.ini.php`, better use dedicated Lizmap `lizmapConfig.ini.php` which is editable with the admin interface
We should mark this optiosn as sensitive, hidden when the flag `hideSensitiveServicesProperties`
https://github.com/3liz/lizmap-web-client/blob/master/tests/docker-conf/phpfpm/lizmapConfig.ini.php#L3

rldhont:
@mind84 can you add the changes requested by @Gustry and @mdouchin  ?

rldhont (changes requested):
Implementing the changes requested by @Gustry and @mdouchin

mind84:
> @mind84 can you add the changes requested by @Gustry and @mdouchin ?

Hi @rldhont!

I think I'll need some clarifications about these changes, probably next week I could start working on this. 

Thanks

Gustry:
Thanks @mind84 

Later, we will forward this new flag in the JSON metadata to make the plugin aware about these layers.

mind84:
Hi @Gustry !

For now I have added the additional configuration in the Lizmap admin panel in the existing `Interface` section. I chose this section because it seemed suitable for the purpose, but also because I don't want to add further sections.

Any suggestion about labes or options placement would be appreciated :)

@mdouchin they are also marked as sensitive, so when the corresponding flag is enabled, they do not appear in the config interface.

image

So, this will probably be a very silly question, but I want to be sure about the logic after this point. If, say, `Google` is disabled, then it should not appear as baselayer and should be removed from baselayers list? In this case I think the best way to do it is to remove it from the `getProjectConfig` service response.

Thanks!

mind84:
Hi @Gustry @mdouchin !

Do you have any feedback on this?

> If, say, Google is disabled, then it should not appear as baselayer and should be removed from baselayers list? In this case I think the best way to do it is to remove it from the getProjectConfig service response.


Thanks!

Gustry:
Sorry, I missed your question. I would say yes, to "remove the layer" on the server side, so the frontend won't be aware.

Would you be able to forward these values in the JSON metadata from the server : 
[lizmap/modules/lizmap/lib/Server/Server.php](https://github.com/Gustry/lizmap-web-client/blob/master/lizmap/modules/lizmap/lib/Server/Server.php#L264)

like 
```json
"authorized_external_providers" : ["google", "bing"],`
```
I can plan a dev on the Lizmap plugin side.

mdouchin:
> If, say, Google is disabled, then it should not appear as baselayer and should be removed from baselayers list? In this case I think the best way to do it is to remove it from the getProjectConfig service response.

I think we should indeed remove the layer from the baselayer list. I a not sure though that removing it only from `getProjectConfig` would be enough, @rldhont & @nboisteault would know best.

Side question : what about if the layer is not in the baselayers group, but anywhere else ? Not sure this PR should tackle it, but we should think about it too.

mind84:
Hi @mdouchin, @rldhont, @nboisteault,

I'm working on it, but several questions arose during the process.

>I think we should indeed remove the layer from the baselayer list. I a not sure though that removing it only from getProjectConfig would be enough

If I remove the layer from the config then Lizmap throws an error in console, eg:
`The WMS layer name Google_Satellite is unknown!`
so I think that the bet way to remove these layers from the baselayers is to remove (or rather, avoid to add) the `externalAccess->url` property in the config and add a property like 'disabled', and then do some checking via js:

```json
"externalAccess": {
                "crs": "EPSG:3857",
                "format": "",
                "type": "xyz",
                "url": "",<----------------
                "zmax": "20",
                "zmin": "0",
                "disabled": true <-----------------------------------
            }
```
This way, also the url is hidden from the client (not a big achievement but it's better IHMO)

>what about if the layer is not in the baselayers group, but anywhere else ?

Generally speaking, TOS has nothing to do with the Api Key
Provide an API Key should be mandatory in the first place. So the new control on the admin is meant to allow external google and bing WMTS, indipendently from the API key.


This is my proposal (eg for Google, extended to Bing end so on):


<table>
    <thead>
        <tr>
            <th>Map Tiles Enabled</th>
            <th>Base Layers </th>
            <th>Lizmap Layers</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <th >enableGoogleMapsTiles = ON</th>
            <td>
               <b>API Key IS SETTED</b> <br /> Load WMTS trough OpenLayers API<br /></br />
               <b>API Key IS NOT SETTED</b><br /> avoid to display the layer in the baselayer and console.warn() that the api key is needed (avoid also to call the OL API if the the key is not provided)
            </td>
            <td>There is no control on the API Key in this scenario, the layer is loaded by QGIS server so I think should be displayed according to the enableGoogleMapsTiles = on</td>
        </tr>
        <tr>
            <th>enableGoogleMapsTiles = OFF</th>
            <td colspan='2'>No external google WMTS should be displayed, despite API Key</td>
        </tr>
    </tbody>
</table>

Please let me know if I completely misinterpreted the request.

Thanks

mind84:
EDIT:

> If I remove the layer from the config then Lizmap throws an error in console

This also happens when a layer is only visible to certain groups:

image

In console:
image

To keep the same logic, probably the option to completely  remove the layer from config is still valid

mdouchin:
We should have a simple way to disable any group or layer to be displayed in the layer tree & the map. cc @rldhont @nboisteault 
We should keep it safe and simple.

Gustry:
I don't think the logic was correct.

> enableGoogleMapsTiles = OFF → No external google WMTS should be displayed, despite API Key

* In the config INI file, does the server has a strict check of TOS for Google or Bing : 

  * **Yes**, so any layer loaded with QGIS Desktop for Google or Bing must a have valid API key. If no API key provided, the layer is discarded from the project.
  * **No**, current behavior of Lizmap. If API key is set, well, all good, it worls. If not API key, maybe it works, maybe not. It depends of the provider, it seems #4192 Bing is not working. But I know Google works sometimes ... 🤷
  * **Not set** in Ini file, we will choose a default value, set to `yes` for instance.

But maybe, the first PR is to do the PHP config file `lizmapConfig.ini.php`  and the forward in the JSON. Do you want me to have a look ?

Gustry:
@mind84 We have discussed.
We are fine with the current state of the PR, Javascript only. Can you wait a little ?

mind84:
> We are fine with the current state of the PR, Javascript only. Can you wait a little ?

Of course, if you want I have also the updated PR on my local machine with the admin panel part, the metadata service updated (for lizmap plugin) and the controls on the the layers (remove layers when enableGoogleMapsTiles = OFF, for instance)

If could share it without push it, if you just want to take a look.

Let me know!

Thanks!

Gustry:
Hi @mind84 

You can look at my PR https://github.com/3liz/qgis-lizmap-server-plugin/pull/80

This will discard any Google or Bing Layers from the `GetCapibilities`, so it will make your PR easier. That's why you can keep this PR with JavaScript only, because everything will be done upfront, in QGIS server.
Sorry you started the form in Lizmap administration panel ! It looks more logical for me as well to have a UI to set up this configuration, that's I talked about this solution.
But thinking about it, it's way easier that layers are discarded straight from QGIS server response.

mind84:
@Gustry ,
Ok thanks, looks reasonable too me too.

Side question:
> You can look at my PR https://github.com/3liz/qgis-lizmap-server-plugin/pull/80

The overall logic is pretty simple, but I'm not very familiar with the Lizmap plugin server code. How can I enable/disable the `ApiKey` control? 

Thanks

Gustry:
> The overall logic is pretty simple, but I'm not very familiar with the Lizmap plugin server code. How can I enable/disable the ApiKey control?

I just pushed again on the PR. Can you have a look to the `README.md` explanation, and let me know if it's clear ? I would prefer to explain through the doc instead of here ;-)

We can merge if you want, so you would have a zip on https://packages.3liz.org/pub/lizmap_server-qgis-plugin/unstable/ or on custom QGIS repository https://packages.3liz.org/pub/server-plugins-repository/unstable/

Gustry:
> Ok! For me it's clear. Sys admin can set environmental variables (e.g. configured in docker compose or whatever) to enable/disable the Api Key check.

Yes it's the purpose. I was wondering if just a single variable would be OK, because I don't think there is a use case to enable just one and the other one. But let's go like this.

> I only point this out because from the README it seems that, by switching the API Key control from False to True, the system becomes TOS compliance too.

Yes true, I will see how I will rephrase on that. I will merge, so the ZIP will be available. Maybe after in the documentation docs.lizmap.com

rldhont (approved):

rldhont:
Backports ?

mind84:
> Could you try with the latest version of the plugin and was it successful ?

yep, I'll keep you updated

mind84:
@Gustry @rldhont,

Server side seems all good to me. I've added both env variables on the `docker-compose`:
image

and system responds as documented.

Problem is, when the `tos_check` removes all layers from the getCapabilities response, the `baseLayers` group ends up being empty, and then the client crashes:

image

Not a big deal I think, I can fix it but it will probably take some time.

I'll keep you updated

Thanks

Gustry:
> Not a big deal I think, I can fix it but it will probably take some time.

Nice thanks.
Indeed, at least skipping "nicely" the loop if it's empty would be better.

Because in your case, you fall back on https://github.com/3liz/lizmap-web-client/issues/4029 which was "fixed" with a fix on the plugin side https://github.com/3liz/lizmap-plugin/issues/540 but I didn't expect the `baselayers` group to become empty when layers are discarded on runtime.

Thanks for this patch anyway !

Gustry (approved):
Nice !
LGTM

What do you think about adding a very basic test loading `google_basemap.qgs` and just testing no error is raised, or the expected google layer is present ? by using the new helper `gotoMap` ?
It seems the new project is not used for automatic testing.

mind84:
> What do you think about adding a very basic test loading `google_basemap.qgs` and just testing no error is raised

Yes I could add a basic test, of course it will not be capable to test the whole functionality but its better than nothing

Thanks

Gustry:
> Yes I could add a basic test, of course it will not be capable to test the whole functionality but its better than nothing

No sure, we do not want to test the OpenLayers Google class :)

Just that the map load correctly, and there isn't any JS error when loading the map.
Maybe similar to what you have done for Bing https://github.com/3liz/lizmap-web-client/blob/master/tests/end2end/playwright/bing-basemap.spec.js

tests/end2end/playwright/google-basemap.spec.js
```diff
@@ -0,0 +1,14 @@
+
+const { test, expect } = require('@playwright/test');
+const { gotoMap } = require('./globals');
+
+test.describe('Google Maps Baselayers', () => {
+    test('Load map', async ({ page }) => {
+        const url = '/index.php/view/map?repository=testsrepository&project=google_basemap';
+        await gotoMap(url, page);
+
+        // baselayers group should be hidden since it is empty due to the default STRICT_GOOGLE_TOS_CHECK env variable value = TRUE
```
Gustry, line 10: For testing purpose, maybe we can switch to `False` in the docker-compose file, or add a dummy Google API key in the CFG file, so the transformation to OL Google layer can be tested ?

mind84: Switching the docker-compose variable to `false` could lead to non predictable response, I don't know if the test could be solid this way, but it is an option. Adding a dummy API Key will raise a console error for sure, so it's predictable on 100%. 

If the purpose is to test the OL Google initialization I would choose the last one.

Another way is to duplicate the project by keeping the default `true` value for the `tos check`:
1. project wit no API key -> testing baselayers group is empty (no error in console)
2. project with dummy API Key -> testing OL class init

What do you think about it?

Gustry: > Adding a dummy API Key will raise a console error for sure, so it's predictable on 100%.

I though this would lead to empty tiles or similar (tiles with the Google Watermark saying "For dev purpose only")  or a JS warning, not error.

I though it would be good to test "your" actual Javascript content from this PR, so indeed the OL Google init from LWC.

It seems good what you propose.

mind84: > I though this would lead to empty tiles or similar

Sort of. I can confirm that OL class requires a valid API Key to create the session. The is no option to get a dummy api key only for testing purpose to get things works with no errors. 

image

I can try to catch these errors to test that the system tries to start the Google session issuing a generic APY key

Gustry: Ok, thanks for checking.

> I can try to catch these errors to test that the system tries to start the Google session issuing a generic APY key

Ideally, yes, if you want to improve and warn

So let's merge like this ?

mind84: For me yes

Gustry (approved):

nboisteault:
Thank you @mind84!
@rldhont @Gustry I'm not for a backport to 3.8 as we made effort to stabilize it and are almost ready to release it.

Gustry:
Thanks @nboisteault Do you still approve the PR ?

I personally see :

* one bugfix, about empty `baselayers` group, which should be backported anyway
* the code related to Google which seems "reachable" mainly/only with some `if google layer is detected then ...` so it's looks fine to me
* Google layer is important for legal issues IMHO

So I'm in favor of merging.

I will let @rldhont choose

nboisteault (approved):

PR #4367, opened 2024-04-15
Make QGIS Server 3.34 for testing in LWC 3.8

Gustry:
While #4323 is only targeting tests, I'm wondering about raising QGIS server min to 3.34 to the minimum required version for 3.8.

I know we got some discussions on 3.7 to limit to 3.22, because of some dependencies on Ubuntu 22.04 or similar.

But now, LWC 3.7 relies a lot the JSON legend from QGIS Server : 

* this JSON legend is "quite" new in the QGIS server implementation
* There are a few bugfix on QGIS server about this JSON legend and it can only be backported to 3.34 minimum now. So for the quality of the project and related to bug reports, it would make sense to have latest bugfix from QGIS server about the legend.
* We don't have 3.28 anymore most of Ubuntu users because of APT, so therefore generating some 3.34 projects. This is a problem
* 3.28 is already not maintained (but on the other side, we can see it as an "stable" 3.28 now as well)

Open to discuss

@mdouchin @rldhont @nboisteault @mind84 @gioman @josemvm


mdouchin:
+1 for me

guenterw:
Many thanks for the information. It is understandable that future versions of Lizmap should be based on the latest possible QGIS server versions. And at some point, an upgrade to Ubuntu 22.04 or a new server is due.

However, my observation is that Lizmap 3.7 and 3.8 is slower than Lizmap 3.6, even with QGIS Server 3.34. and that especially in the legend the loading of the layers (rotating circle) is not displayed correctly.
I have carried out test projects with the same QGIS project for:
https://cx21-6.webgis.biz  Lizmap 3.6.11
https://cx21-7.webgis.biz  Lizmap 3.7.6
https://cx21-8.webgis.biz  Lizmap 3.8.0 Alpha.1

With QGIS-Server/Desktop 3.34.6, Lizmap-Plugin 4.3.6
The server is quite slow (small cloud server for testing)

nboisteault:
@guenterw I used lighthouse with your Lizmap 3.6, performance score is 35/100. With 3.8, performance score is 40/100. Those performance score could be way better with right server configuration as enabling http2 for example. But anyway 3.8 score is best.

guenterw:
@nboisteault Thank you very much. It is very interesting that the subjective perception of speed is quite different. Version 3.7 and 3.8 simply don't load as smoothly subjectively. This is also the observation of other users and my customers.
I will test the use of http2.

guenterw:
Sorry, this is not really part of the topic but it is very important for me to use the new versions.

I have now activated http2 but this does not bring any change. The measurements with lighthouse fluctuate a lot and give a completely different picture than the normal test in the browser.
So I preferred to stop the speed with the watch. Several times, each time clearing the browser cache.
It is interesting that there are big differences between the PC (older i5 Win10), the notebook (newer i7 Win10) and mobile devices (Samsung tablet).
In each case from the start of the project (Bauleitplanung),
image

immediately zoomed to a maximum scale of 1:250 (PC and notebook with mouse scroll wheel):

https://cx21-6.webgis.biz/ Lizmap 3.6.11:  PC 10 sec., NB 5 sec. Mobile < 5 sec.
https://cx21-7.webgis.biz/ Lizmap 3.7.6:    PC 30 sec., NB 16 sec., Mobile 10-15 sec.
https://cx21-8.webgis.biz/ Lizmap 3.8.0 Alpha.1  PC 54 sec., NB 26 sec., Mobile 10-15 sec.

How can we explain these differences?
And version 3.6 is always significantly faster in all tests

With versions 3.7 and 3.8 you sit in front of a white screen for a long time (especially on the PC)
image

nboisteault:
@guenterw please open a new ticket for this issue.

Gustry:
After some discussion with @rldhont I will revert the part that makes QGIS Server 3.34 required. It will be 3.28.
However, people will be encouraged to migrate to QGIS Server 3.34, because for instance bugfix can only land on the **current maintained** branch by QGIS, so 3.34 (JSON legend etc)

We are only talking about QGIS Server version.
A follow up PR will be about QGIS Desktop.

Gustry:
Temporary cherry-picked PR #4516 to test on QGIS Server 3.34 @rldhont

Gustry:
E2E tests have been temporary disabled about the default background color.

PR #4368, opened 2024-04-16
add STYLE parameter to GetFeatureInfo request

mind84:
When a `GetFeatureInfo` request is issued to the server and no `STYLE` parameter is passed, QGIS Server
uses the style of the layer that was active when the `QGIS` project was saved.

If a layer has multiple style, Lizmap allows to switch between styles on client side and this could lead to inconsistencies as the server always uses the same style when processing the response.

Take for example this test project:
[get_feature_info_style.zip](https://github.com/3liz/lizmap-web-client/files/14987788/get_feature_info_style.zip)

The `natural_areas` layer has 3 feature and is configured with 2 rule based styles: `default` and `ids`. The project is saved with the `default` style active.

With the `default` style, the feature with **id = 3 is not shown** on the map while with the `ids` style all features are shown.

On client side, if the user switches to `ids` style and clicks on the feature with **id = 3** (now visible) to get the `popup`, the server response will be empty because the `default` style is used.

I added the `STYLE` parameter to the `GetFeatureInfo` requests and updated `Popup` tests to take into account this possible scenario.

Funded by Faunalia

### DIFF
--- a/assets/src/modules/Popup.js
+++ b/assets/src/modules/Popup.js
@@ -109,6 +109,7 @@ export default class Popup {
         }
 
         const layersWMS = candidateLayers.map(layer => layer.wmsName).join();
+        const layersStyles = candidateLayers.map(layer => layer.wmsSelectedStyleName || "").join()
 
         const wms = new WMS();
 
@@ -123,6 +124,7 @@ export default class Popup {
         const wmsParams = {
             QUERY_LAYERS: layersWMS,
             LAYERS: layersWMS,
+            STYLE: layersStyles,
             CRS: mainLizmap.projection,
             BBOX: bbox,
             FEATURE_COUNT: 10,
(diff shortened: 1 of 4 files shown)


rldhont:
@mind84 thanks.

Did you test this case ?
* 2 styles : One with QGIS tooltip, the second without
* Lizmap config with Popup based on QGIS Tooltip

mind84:
> Did you test this case ?
> 
>     * 2 styles : One with QGIS tooltip, the second without
> 
>     * Lizmap config with Popup based on QGIS Tooltip


Hi @rldhont,

I've updated the test to cover this case.
 - `default` style has QGIS Tooltip configured
 - `ids` style has no QGIS Tooltip configured
 - project is saved with `ids` style active
 
On Lizmap:

- default style active-> popup is correctly rendered as QGIS Tooltip
- when switch to the `ids` style (no QGIS Tooltip configured), `default automatic Lizmap popup` is rendered (no errors occurs)

Thanks

rldhont:
Good catch @mind84 !

rldhont (approved):

PR #4376, opened 2024-04-22
Draw: keep draw visible when closing minidock

nboisteault:
Funded by 3Liz


mdouchin (approved):
Thanks. I think this should be backported into 3.7

Gustry:
I didn't try this PR, but yes, let's the user manage the visibility of his drawings. (if the PR works in that way).

nboisteault:
Unfortunately it can't be backported to 3.7 because of switch between OL 2 and 9 maps.

PR #4383, opened 2024-04-24
[Bugfix] Display error message on error fecthing edition form

rldhont:
If the serveur does encountered an error or if it does not provide the right content, an error message is displayed to the user.


lizmap/modules/view/locales/en_US/dictionnary.UTF-8.properties
```diff
@@ -56,6 +56,7 @@ edition.confirm.cancel=Are you sure you want to cancel the edition? You will los
 edition.confirm.restart-drawing=Are you sure you want to delete the existing geometry and redraw it from scratch?
 edition.confirm.paste=Are you sure you want to replace the existing geometry with the one in the clipboard?
 edition.geom.paste=Paste the geometry
+edition.message.error.fetch.form=Form fetching error!
```
Gustry, line 59:
```suggestion
edition.message.error.fetch.form=Error while fetching the editing form for that layer. Please contact the system administrator.
```

I don't think you should use `!` in error message. It's the not the fault of the end user.

Gustry (commented):
UX review about the error message.

Is-it temporary until we can do better for admin versus non admin user ?

tests/end2end/playwright/edition-form.spec.js
```diff
@@ -0,0 +1,114 @@
...
+      // display form
+      await page.locator('#button-edition').click();
+      await page.locator('a#edition-draw').click();
+
+      // add data
+      await page.locator('#jforms_view_edition input[name="integer_field"]').fill('50');
+
+      await page.route('**/edition/saveFeature*', async route => {
+          await route.fulfill({
+              status: 404,
```
Gustry, line 89: If it works for 50X error as well, LGTM ;-)

Thanks a lot

This is only client side. For PHP side, we must check that the 500 error is correctly logged in the correct log file for GIS administrator.

Gustry (approved):

PR #4384, opened 2024-04-24
[Bugfix] Display error message on error printing

rldhont:
If the serveur does encountered an error during generating print content, an error message is displayed to the user.

### DIFF
--- a/lizmap/modules/view/locales/en_US/dictionnary.UTF-8.properties
+++ b/lizmap/modules/view/locales/en_US/dictionnary.UTF-8.properties
@@ -14,6 +14,7 @@ tree.button.collapse=Collapse
 print.launch=Print
 print.activate=Place the rectangle for printing
 print.started=Print started, please wait…
+print.error=The output is currently not available. Please contact the system administrator.
 print.customLabels.tooltip=Set custom labels for print
 print.toolbar.layout=Layout
 print.toolbar.scale=Scale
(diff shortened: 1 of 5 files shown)


lizmap/modules/view/locales/en_US/dictionnary.UTF-8.properties
```diff
@@ -14,6 +14,7 @@ tree.button.collapse=Collapse
 print.launch=Print
 print.activate=Place the rectangle for printing
 print.started=Print started, please wait…
+print.error=Error while printing!
```
Gustry, line 17:
```suggestion
print.error=The output is currently not available. Please contact the system administrator.
```

tests/end2end/playwright/print.spec.js
```diff
@@ -615,3 +615,67 @@ test.describe('Print base layers', () => {
...
+                    contentType: 'text/plain',
+                    body: 'Not Found!'
+                });
+            else
+                await route.continue();
+        });
+
+        await page.locator('#print-launch').click();
+
+        await expect(page.getByText('Error while printing!')).toBeVisible();
```
Gustry, line 646:
```suggestion
        await expect(page.getByText('Error while printing!')).toBeVisible();
```

use CSS selector instead of text ?

tests/end2end/playwright/print.spec.js
```diff
@@ -615,3 +615,67 @@ test.describe('Print base layers', () => {
...
+                    contentType: 'text/plain',
+                    body: 'Not Found!'
+                });
+            else
+                await route.continue();
+        });
+
+        await featureAtlasQuartiers.locator('button').click();
+
+        await expect(page.getByText('Error while printing!')).toBeVisible();
```
Gustry, line 675: Same

Or update the string to match the proposition.

PR #4398, opened 2024-05-01
[Fix] Add Popup dock server side

rldhont:
The popup dock is build client side by the JS and depends on `lizDict`. Since https://github.com/3liz/lizmap-web-client/pull/4329 the popup dock could be created before the `lizDict` is available.


rldhont:
@mdouchin @nboisteault ready for review. backport or not ?

mdouchin (approved):

mdouchin:
backport to 3.7 would be great

PR #4401, opened 2024-05-02
Fix HTML for dataviz filtered in popup

nboisteault:
@mdouchin this line makes an error in the HTML plot I have added : https://github.com/3liz/lizmap-web-client/commit/cf12831fa477438ab75b1f70d37621dec5e67a4b#diff-e736f3d4588193cb20a1abc33bb3dbcf9b1ba1aed7d96afde0a8e7f2e7381a26R456


mdouchin:
thanks for this, I will tackle it

PR #4403, opened 2024-05-03
JS: Move lizmap variables from window to globalThis

rldhont:
Move lizmap variables `lizUrls`, `lizProj4`, `filterConfigData`, `filterConfig` and `datavizConfig` from `window` to `globalThis`; and access them through `globalThis`.


mdouchin (approved):

Gustry:
So I guess it breaks all usage of `lizUrls.params.repository` for instance ?
in javascript library, in plausible, in the doc https://docs.lizmap.com/current/fr/publish/customization/javascript.html#url-of-a-static-file

rldhont:
> So I guess it breaks all usage of `lizUrls.params.repository` for instance ? in javascript library, in plausible, in the doc https://docs.lizmap.com/current/fr/publish/customization/javascript.html#url-of-a-static-file

It does not change anything for javascript library. It will be help to write tests in other JS environment than browser.

Gustry:
Sorry, noob question, but I won't need to change `lizUrls` to `globalThis['lizUrls']` ? It seems you did it, that's  why i wanted to confirm.

if so, the label "doc-needed" can be removed ?

rldhont:
> Sorry, noob question, but I won't need to change `lizUrls` to `globalThis['lizUrls']` ? It seems you did it, that's why i wanted to confirm.
> 
> if so, the label "doc-needed" can be removed ?

In javascript library you don't need to change `lizUrls` to `globalThis['lizUrls']`. I do it to be sure in Browser or Node to access `lizUrls`. It is to be able to write unit tests.

assets/.eslintrc.json
```diff
@@ -2,6 +2,7 @@
     "env": {
         "browser": true,
         "es6": true,
+        "es2020": true,
```
nboisteault, line 5: In this case, you can remove the line above I guess.

nboisteault (approved):

PR #4408, opened 2024-05-03
Tests e2e: Test error occured

rldhont:
Abort request to test the `error occured` message dispalyed.


Gustry (approved):
LGTM as it is ;-)
I would be nice to give more context when this message is displayed.
The user is facing a kind of "unknown" error without any clue about to look for, for a PR later as a new improvment ;-)

Issue #4396, opened 2024-04-29
[Bug]: data - table, virtual field with expression to show small image preview


josemvm:
### What is the bug? (in English)

this expression here https://docs.lizmap.com/next/en/publish/lizmap_plugin/attribute_table.html#tips stop working on 3.7.7-pre.7395

### Steps to reproduce the issue

try to use this expression between 3.7.6 and 3.7.7-pre versions

### Versions, safeguards, check summary etc

**Versions :**

* Lizmap Web Client : 3.7.7-pre.7395
* Lizmap plugin : 4.3.5
* QGIS Desktop : 3.28.15
* QGIS Server : 3.28.15
* Py-QGIS-Server : not used
* QGIS Server plugin atlasprint : 3.3.2
* QGIS Server plugin lizmap_server : 2.8.6
* QGIS Server plugin wfsOutputExtension : 1.8.0

<details>
<summary>List of safeguards :</summary>
<br/>

* Mode : normal
* Allow parent folder : no
* Prevent other drive : no
* Prevent PG service : no
* Prevent PG Auth DB : no
* Force PG user&pass : no
* Prevent ECW : no
</details>

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 22.04

### Browsers

Firefox

### Browsers version

122.0.1

Gustry:
Can you elaborate ? What is is not working ? What is the error in the console ? Any picture ?

josemvm:
@Gustry the table is not displayd at all on data tool

Gustry:
Any error message displayed in the console ? (F12)

josemvm:
no, no errors displayed in the console, just `11:31:07 WARNING Server[15550]: Error parsing post data as XML: at line 1, column 1: error occurred while parsing element. Assuming urlencoded query string sent in the post body.` in qgis server log file...

Gustry:
Please describe your issue @josemvm 

> this expression here [URL] stop working on 3.7.7-pre.7395

is very vague.

Is the panel empty ? Is the table there but empty (table headers) ?
Does it work if you remove the virtual field ? (maybe try to not expose it with WMS/WFS?)
Is the button to open the "Attribute table" panel OK ? Is the button to choose the layer to display in the panel OK ?
[...]

PR #4409, opened 2024-05-03
Content sanitisation in attributes table

nworr:
Ticket : #4396 

more precision in [comment](https://github.com/3liz/lizmap-web-client/issues/4396#issuecomment-2093118247)

I completed  the cypress test with a check that inline js is cleaned (whole attributes is removed)

Funded by 3Liz


nboisteault (approved):
LGTM

PR #4412, opened 2024-05-06
Datatables v2: update files

nboisteault:
- [ ] `fn*` functions must be upgraded to the new API: https://datatables.net/upgrade/1.10-convert#API
- [ ] understand why there is no `dataTables_wrapper` class on div containing the table (https://datatables.net/blog/2012/bootstrap-2#Layout)

Funded by 3Liz


nboisteault:
Supersed by https://github.com/3liz/lizmap-web-client/pull/5562

PR #4393, opened 2024-04-26
[Backport release_3_7] JS factorize some map and treeview code

rldhont:
Backport https://github.com/3liz/lizmap-web-client/pull/4391
Authored by: @rldhont

PR #4416, opened 2024-05-07
Tests - Add helper to test if the map is OK when opening it

Gustry:
@nboisteault This kind of helper would have helped me a lot yesterday about #4386 or for @rldhont about #4393

If possible @nboisteault should we use **everywhere** this kind of helper to open a map ?

Sorry, I'm not a JS dev...
I know there is an error about playwright now

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -259,7 +259,7 @@ You can then :
 - execute `npx playwright test mytest.spec.js --project=chromium --debug` to execute one test with the Chromium browser in debug mode
 - other command line : https://playwright.dev/docs/intro#command-line
 
-You can also install the handy Visual Studio Code extension : https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright
+You can also install the handy [Playwright extension](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright) on VSCode.
 
 ### Artifacts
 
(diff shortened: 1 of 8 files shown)


Gustry:
> @Gustry this is not possible w/ playwright?

What do you mean ? As I wrote, I'm not a JS dev, and I first wanted the idea of this PR.
I was quoting this link in my PR : 
https://github.com/microsoft/playwright/issues/10358

Gustry:
I kind of `gotoMap` to visit a map and check some basic requests such as `getCapabilities` or basic button in the interface.

rldhont:
A try for playwright - to be generalized https://github.com/3liz/lizmap-web-client/pull/4451

Gustry:
@rldhont @nboisteault I have rebased after the PR from @rldhont #4451 

I don't know how we can pass the flag for triggering a false positive test about the error.

I will merge this PR if OK for you. This work is be continued in a next step : 

* migrate all `gotoMap` for both cypress and playwright, like #4478
* Improve both function `gotoMap` to check WMS requests and avoid a wasted `wait` etc, because this will considerable slowdown tests when running them all otherwise
* Uncomment the false positiv test about invalid layer in both Cypress/playwright

Stoping on my side on this

PR #4421, opened 2024-05-13
[Fix] Popup children compact table

rldhont:
When the popup of 2 features in the same layer with children is displayed, DataTable displays a dialog about this error https://datatables.net/manual/tech-notes/3

Funded by Terre de Provence Agglomération


mdouchin (approved):
thx

PR #4435, opened 2024-05-14
 Fix: Add project-background-color even if the default_background_color_index is not defined

rldhont:
Since QGIS Server 3.34, empty groups are not available in WMS GetCapabilities. project-background-color is an empty group placed in baselayers group to provide no baselayer in Lizmap Web Client User Interface.
Since Lizmap Desktop Plugin 4.3.0, an option default_background_color_index provides the index to add project-background-color in the list of baselayers.

If the configuration has not been updated, but the QGIS Server has been, Lizmap Web CLient will not provide a no baselayer to the user.
This changes wil provide a no baselayer capability, if the project-background-color is found in the Lizmap config at the right place.

Funded by 3liz

### DIFF
--- a/assets/src/modules/config/BaseLayer.js
+++ b/assets/src/modules/config/BaseLayer.js
@@ -925,17 +925,48 @@ export class BaseLayersConfig {
         }
 
         // Add base layer as project default background color
+        // Get provided default background color index from options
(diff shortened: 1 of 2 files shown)


Gustry:
Linked to #4367 and #4319

Gustry (approved):
By reading the code, it's way better for the compatibility. Thanks.
LGTM

PR #4425, opened 2024-05-13
Dev - Add basic pre-commit hook

Gustry:
It uses the default config of pre-commit for now : https://pre-commit.com

```yaml
    -   id: trailing-whitespace
    -   id: end-of-file-fixer
    -   id: check-yaml
    -   id: check-added-large-files
```

It's Python tool. It's not yet required, just an helper to fix files.

~I will revert changes on some path~ (it should be OK now) 

Possible to add more later : 

* Eslint https://github.com/pre-commit/mirrors-eslint

I'm already using it successfully on some Python project, since April 2023 https://github.com/3liz/lizmap-plugin/commit/1ca3988d3efedbfdd0bdb2b6342b1fd26cc4fadf

Related to PR https://github.com/3liz/lizmap-web-client/pull/4421/commits where we shouldn't have some lint commits IMHO.

If accepted, I will manually backport to current supported branches.

PR #4436, opened 2024-05-14
Fix all trailing-whitespaces and end-of-file

Gustry:
Fix all trailing-whitespaces and end-of-file

Manual backport from #4425


Gustry:
Unrelated test failure about cypress

PR #4440, opened 2024-05-14
Tests - Switch to QGIS 3.28

Gustry:
Tests - Switch to QGIS 3.28


Gustry:
Anything against this change @rldhont @mdouchin @nboisteault @mind84 ?

I would preferred that we test against legacy versions, but it's very hard to maintain, and let's keep it simple and I would consider 3.28 as legacy now.

This is for 3.7 branch. Very tempted to add the backport 3.6

nboisteault (approved):

PR #4447, opened 2024-05-16
Fix: Group with layers can be used as baselayer

rldhont:
Fix regression from https://github.com/3liz/lizmap-web-client/pull/4435

Funded by CC Bièvre Est


assets/src/modules/map.js
```diff
@@ -479,6 +479,7 @@ export default class map extends olMap {
                                 })
                             });
                         } else {
+                            console.log(baseLayerState.name);
```
Gustry, line 482: Just a question, should we let this line ? i don't know what is the best practice.

PR #4456, opened 2024-05-21
[Bugfix] SingleWMSLayer - refresh layer after editing

mind84:
When a layer is included in the `single WMS image` it is not refreshed after updating/inserting a new feature.

Funded by Faunalia

### DIFF
--- a/tests/end2end/playwright/singleWMS.spec.ts
+++ b/tests/end2end/playwright/singleWMS.spec.ts
@@ -411,4 +411,44 @@ test.describe('Single WMS layer', () => {
         expect(switchBaseLayersReqPromiseresp?.status()).toBe(200)
 
     })
+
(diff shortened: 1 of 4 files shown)


rldhont (approved):

tests/end2end/playwright/singleWMS.spec.ts
```diff
@@ -411,4 +411,44 @@ test.describe('Single WMS layer', () => {
         expect(switchBaseLayersReqPromiseresp?.status()).toBe(200)
 
     })
+
+    test('Edit a layer', async ({ page }) => {
+        const url = '/index.php/view/map/?repository=testsrepository&project=single_wms_image';
+        await page.goto(url,{waitUntil:'networkidle'});
```
Gustry, line 417: @rldhont Shouldn't we use `gotomap` ?

PR #4466, opened 2024-05-23
[Bugfix] Treeeview symbols not in scale

rldhont:
A class is added to the symbol li if the symbol is out of scope and the category is italic and greyed.

Capture d’écran 2024-05-24 à 10 15 20

Capture d’écran 2024-05-24 à 10 15 32

Funded by EPF Haut De France

### DIFF
--- a/assets/src/components/Treeview.js
+++ b/assets/src/components/Treeview.js
@@ -64,7 +64,7 @@ export default class Treeview extends HTMLElement {
 
         this._symbolTemplate = symbol =>
             html`
-        <li class="symbol">
+        <li class="symbol${this._isInScale(symbol) ? '' : ' not-in-scale'}">
             ${(symbol.childrenCount)
                 ? html`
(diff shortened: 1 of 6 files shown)


Gustry:
Do you mean for rules which have predefined scales, like this ?

image

rldhont:
Yes @Gustry , how this is displayed in QGIS treeview ?

Gustry:
By default, strictly no change...

Peek 2024-05-24 09-57

Only if you toggle this : 

image


Then it's very dynamic, but not related to scale range of the symbol, but only related to symbols drawn : 

Peek 2024-05-24 10-00

But your idea is good. I would make it italic and/or greyed, similar to layer scale visibility : 

image

rldhont:
Thanks @Gustry .

I have updated the main comment with previews.

rldhont:
@Gustry ready for review

PR #4473, opened 2024-05-24
[Bugfix] Treeview: italic and greyscaled uncheked symbols and their children

rldhont:
Capture d’écran 2024-05-24 à 13 23 55

Capture d’écran 2024-05-24 à 13 24 02


Gustry:
Looks nice.

I wonder if for such rules, we should expand/collapse group (`50 K+` if we are at less than 50 K). But maybe this would be in conflict with other rules like theme etc. It would make a "transition" when zoom in for instance. This is another request :)

PR #4478, opened 2024-05-25
 Tests e2e playwright homogenize files

rldhont:
* Move from TypeScript to JavaScript files
* Format document with 4 spaces
* Using gotoMap method


Gustry:
Nice move !

@rldhont Tip, you can use `git mv`, like `git mv tests/end2end/playwright/dnd-form.spec.ts tests/end2end/playwright/dnd-form.spec.js`, it will make GIT aware that the file has been renamed and it will keep the git history, instead seing that a file has been removed and a new one with similar name has been created.


CC @mind84 for new coding standards FYI as well, like `gotoMap`

PR #4502, opened 2024-06-06
New Lizmap web component lizmap-features-table to display a compact list of features

mdouchin:
This `lizmap-features-table` web component allows to display a **compact list of vector layer features**:

* The QGIS "**Display name**" is used as the feature label: an expression can be used to tweak its content, configured in the Tooltip tab of the QGIS vector layer properties.
* If the **popup** is configured for the given layer, a click on a feature label display its popup under the list of features

Peek 31-07-2024 09-37

It will be used in future in several contexts:

* To display the child elements of a popup object
* To display the child elements in an editing form
* To display a compact list of features anywhere else needed (for example in the expression based HTML popup)

As any **HTML web component**, it can be used by writing an HTML element with attributes allowing to configure its source data and behaviour, for example:

```html
<lizmap-features-table
    uniquefield="id"
    expressionfilter="quartmno = 'MC'"
    withgeometry="1"
    sortingfield="libsquart"
    sortingorder="asc"
    draggable="yes"
/>
```

The following attributes allow configuring the table:

* `layertitle` : the title to use at the top of the table
* `layerid`: The QGIS vector layer internal ID. It can be found in QGIS vector layer properties dialogue, tab `Variables` 
* `uniquefield`: the unique identifier of the layer features. Use the primary key if possible.
* `expressionfilter`: a QGIS expression allowing to filter the displayed features. If no filter is set , all the features will be shown. Beware of bad performances for heavy datasets.
* `withgeometry`: If set to `1`, the geometries will also be used and downloaded. Use it at present if the feature display expression depends on the geometry, for example if it contains calculation made from the geometry, e.g. `area(@geometry)`. In the future, this will also be used to draw the geometry on the map.
* `sortingfield` and `sortingorder`: allow to define in which order the features must be displayed. Default is the alphabetical order of the display expression values. `asc` means ascending order, and `desc` means descending order.
* `draggable`: if set with the value `yes`, the user will be able to change the position of an item in the table by dragging and dropping it.

Events will be fired

* `features.table.rendered` When the table has been successfully displayed. The event carries the `lizmap-features-table` HTML element ID
* `features.table.item.dragged` when the user has dropped an item in a new position. The event carries 3 properties:
  * `itemFeatureId`: the vector feature ID
  * `itemOldLineId`: the original line ID before dropping the item
  * `itemNewLineId`: the new line ID after dropping the item in a new position

Funded by Ville d'Avignon https://cartes.mairie-avignon.com/


nboisteault (approved):

assets/src/components/FeaturesTable.js
```diff
@@ -0,0 +1,446 @@
...
+        // Layer id
+        this.layerId = this.getAttribute('layerId');
+
+        // Error text
+        this.error = null;
+
+        // Get the layer name & configuration
+        const layerName = lizMap.getLayerConfigById(this.layerId)?.[0];
+        let layerConfig = lizMap.getLayerConfigById(this.layerId)?.[1];
+        this.layerConfig = layerConfig;
```
rldhont, line 37:
```suggestion
        // Get the layer name & configuration
        this.layerConfig = mainLizmap.initialConfig.layers.getLayerConfigByLayerId(this.layerId);
        const layerName = this.layerConfig.name;
```
If `this.layerId` is unknown, the `getLayerConfigByLayerId` will throw an error.
You can check if the llayerId is known with `mainLizmap.initialConfig.layers.layerIds.includes(this.layerId)`

assets/src/components/FeaturesTable.js
```diff
@@ -0,0 +1,446 @@
...
+
+        // Primary key field
+        this.uniqueField = this.getAttribute('uniqueField');
+
+        // Expression filter
+        this.expressionFilter = this.getAttribute('expressionFilter');
+
+        // Get the geometry or NetworkError
+        this.withGeometry = (this.getAttribute('withGeometry') !== null) ? this.getAttribute('withGeometry') : 0;
```
rldhont, line 46:
```suggestion
        this.withGeometry = this.hasAttribute('withGeometry');
```

assets/src/components/FeaturesTable.js
```diff
@@ -0,0 +1,446 @@
...
+
+        // Sorting attribute and direction
+        this.sortingField = this.getAttribute('sortingField');
+        const sortingOrder = this.getAttribute('sortingOrder');
+        this.sortingOrder = (sortingOrder !== null && ['asc', 'desc'].includes(sortingOrder)) ? this.sortingField : 'asc';
+
+        // open popup ?
+        this.openPopup = (layerConfig && layerConfig.popup.toLowerCase() === 'true');
```
rldhont, line 54:
```suggestion
        this.openPopup = (layerConfig && layerConfig.popup);
```
In `LayerConfig`, `popup` is a boolean

assets/src/modules/FeaturesTable.js
```diff
@@ -0,0 +1,135 @@
...
+     * @param {string} uniqueField   Field containing unique values (used to set the filter for the WMS request)
+     * @param {HTMLElement} targetElement Target HTML element to display the popup content for the given feature
+     * @param {callBack} callBack Callback function
+     */
+    openPopup(layerId, feature, uniqueField, targetElement, aCallBack) {
+
+        // Get the layer name & configuration
+        const layerName = lizMap.getLayerConfigById(layerId)?.[0];
+        if (!layerName) return null;
+        let layerConfig = lizMap.getLayerConfigById(layerId)?.[1];
```
rldhont, line 93:
```suggestion
        if(!mainLizmap.initialConfig.layers.layerIds.includes(this.layerId)) {
            return null;
        }
        let layerConfig = mainLizmap.initialConfig.layers.getLayerConfigByLayerId(layerId);
        const layerName = layerConfig.name;
```

assets/src/modules/FeaturesTable.js
```diff
@@ -0,0 +1,135 @@
...
+        // Layer WMS name
+        const wmsName = layerConfig?.shortname || layerConfig?.name || layerName;
+
+        // Filter
+        const filter = `${wmsName}:"${uniqueField}" = '${feature.properties[uniqueField]}'`;
+
+        var crs = 'EPSG:4326';
+        if(('crs' in lizMap.config.layers[layerName]) && lizMap.config.layers[layerName].crs != ''){
+            crs = lizMap.config.layers[layerName].crs;
+        }
```
rldhont, line 104:
```suggestion
        if(layerConfig.crs && layerConfig.crs != ''){
            crs = layerConfig.crs;
        }
```

rldhont (changes requested):
Use the new API.

rldhont (approved):

Gustry:
I just noticed this while translating on Transifex new strings.
This PR uses `item/items` : 

image

For consistency with QGIS and other parts of Lizmap, shouldn't we use `feature` ?
I'm just trying to be consistent across the application.

erw-1:
When a feature table item is selected, other items are automatically hidden, as expected. However, if there is another feature table in the selected feature, new items are also hidden before they can be picked, because they share the same CSS selector.

You could introduce a hierarchy level into the selector to differentiate between parent and child feature tables.

Video example (FRA): https://youtu.be/zZafmypUBVs

rldhont:
Hi @erw-1 if you find an issue, please open one.

rldhont:
@erw-1 you can also provide a Pull Request if you have a solution to fix it.

erw-1:
Here, this is my first contribution/ pull request ever! Hopefully I did things right.
I'm a GIS student, learning dev stuff recreationally at the moment.

https://github.com/3liz/lizmap-web-client/pull/5037

vlebert:
Hi @mdouchin 

For this new component is there signal like `lizmappopupdisplayed' but for the feature on the feature table ? To trigger JS script when an item is displayed ?

vlebert:
Ok I just found it sorry 

features.table.rendered

vlebert:
@mdouchin 
What is the correct syntax to call the event

```
lizMap.events.on({
  lizmappopupdisplayed: function (e) {
    try {
      window.renderMermaidDiagrams();
    } catch (error) {}
  },
  "features.table.rendered": function (e) {
    try {
      window.renderMermaidDiagrams();
    } catch (error) {}
  },
});
```

This won't fork for `features.table.rendered`

mdouchin:
@vlebert the `lizMap.events.on` syntax is historical, and is not used for the new web components events.
```javascript
lizMap.mainEventDispatcher.addListener(
  (lizmapEvent) => {
    // Log the lizmap-features-table web component HTML element ID
    console.log(`lizmap-features-table rendered, ID = ${lizmapEvent.elementId}`);
  },
  ['features.table.rendered']
);
```

vlebert:
@mdouchin Is there an event triggered when a single feature from the features-table is rendered ?

mdouchin:
> @mdouchin Is there an event triggered when a single feature from the features-table is rendered ?

You can find all events by searching `mainEventDispatcher.dispatch` in https://github.com/3liz/lizmap-web-client/blob/release_3_8/assets/src/components/FeaturesTable.js which show only 3 events: `features.table.rendered`, `features.table.item.highlighted` and `features.table.item.dragged`. See comments and properties for more information.

vlebert:
Hi @mdouchin 

Ok so I need to use `features.table.item.highlighted`. I did try to use it and it works as I expects. However I think we miss the layer ID in addition to the feature ID. Otherwise we don't know which `lizmap-features-table-container` did trigger the signal and we may have different behaviour for different `lizmap-features-table-container`

mdouchin:
Agreed, we should allow to easily know which element triggered the event. Lets continue the discussion in #5667

PR #4513, opened 2024-06-11
[Bugfix] Still support bbox parameter

rldhont:
`bbox` and `crs` parameter was removed in favor of hash permalink.

Funded by Karum, Conseil Départemental du Calvados, Terre de Provence Agglomération


assets/src/modules/Permalink.js
```diff
@@ -201,16 +201,37 @@ export default class Permalink {
 
     // Set URL in permalink component's input
     _refreshURLsInPermalinkComponent() {
+        console.log('_refreshURLsInPermalinkComponent')
```
Gustry, line 204:
```suggestion
```

PR #4549, opened 2024-06-25
[Bugfix] Error building GetPrint parameters for empty baselayer

rldhont:
Funded by Agglomération Fécamp Caux Littoral


Gustry:
@rldhont If you write "Funded by", please add the label as well ;-) Thanks

PR #4586, opened 2024-07-05
New option to disable the automatic permalink

rldhont:
In the main section of the JSON, add a new option to enable the automatic permalink : automatic_permalink.

It will be enabled by default.

Linked to https://github.com/3liz/lizmap-plugin/pull/590

### DIFF
--- a/assets/src/modules/Permalink.js
+++ b/assets/src/modules/Permalink.js
@@ -301,8 +301,10 @@ export default class Permalink {
         // Saved new hash
         this._hash = '#'+hash;
         // Finally override URL fragment
-        this._ignoreHashChange = true;
-        window.location.hash = hash;
+        if (mainLizmap.initialConfig.options.automatic_permalink) {
+            this._ignoreHashChange = true;
+            window.location.hash = hash;
+        }
 
         this._refreshURLsInPermalinkComponent();
     }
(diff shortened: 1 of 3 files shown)


Gustry:
> It will be enabled by default.

You mean "automatic permalink" will be enabled by default ? So it will be the reverse compared to the plugin ?

rldhont:
> > It will be enabled by default.
> 
> You mean "automatic permalink" will be enabled by default ? So it will be the reverse compared to the plugin ?

I keep the LWC 3.7 fonctionnality

nboisteault:
> > > It will be enabled by default.
> > 
> > 
> > You mean "automatic permalink" will be enabled by default ? So it will be the reverse compared to the plugin ?
> 
> I keep the LWC 3.7 fonctionnality

It could be disabled by default in LWC 3.8.

Gustry:
> I keep the LWC 3.7 fonctionnality

We need to be consistent with the plugin so :) Which is not the case, or to define what happen if the key is not found.
So you consider if the key is not found or `True`, then the automatic permalink is done, right ?

> It could be disabled by default in LWC 3.8.

Well, if we want to change the default behavior, I think it's better to do it ASAP.

People are creating new projects. The more we wait, the more they will have to manually update their CFG file.

PR #4591, opened 2024-07-08
Automatic permalink will be disabled by default

rldhont:
After https://github.com/3liz/lizmap-web-client/pull/4586 that added `automatic_permalink` option the default value will be false.


rldhont:
@Gustry the PR is ready.

Gustry (approved):
LGTM

PR #4599, opened 2024-07-12
Bug: filter form should take care of restricted user rights

nboisteault:
- Log as `admin`
- See only one point is visible
- But filter form display all possible values


nboisteault:
@mdouchin to discuss

Gustry:
Side note, but there is already [filter_layer_by_user.qgs](https://github.com/3liz/lizmap-web-client/blob/master/tests/qgis-projects/tests/filter_layer_by_user.qgs) or similar project.
I think, I would appreciate if a project called `filtered_layer_by_user` has different aspects of this feature, like form filter. 

Adding new layer doesn't matter, it takes times to update/upgrade a few projects for now).

PR #4601, opened 2024-07-15
E2E - use the new method gotoMap in some tests

Gustry:


tests/end2end/playwright/snap.spec.js
```diff
@@ -1,10 +1,11 @@
 // @ts-check
 import { test, expect } from '@playwright/test';
+const { gotoMap } = require('./globals')
```
rldhont, line 3:
```suggestion
import { gotoMap } from './globals';
```

tests/end2end/playwright/treeview.spec.js
```diff
@@ -1,5 +1,6 @@
 // @ts-check
 import { test, expect } from '@playwright/test';
+const { gotoMap } = require('./globals')
```
rldhont, line 3:
```suggestion
import { gotoMap } from './globals';
```

tests/end2end/playwright/webdav-upload.spec.js
```diff
@@ -1,10 +1,11 @@
 // @ts-check
 import { test, expect } from '@playwright/test';
+const { gotoMap } = require('./globals')
```
rldhont, line 3:
```suggestion
import { gotoMap } from './globals';
```

rldhont (changes requested):
Prefer `import ... from` instead of `require`

Gustry:
One suggestion is enough @rldhont ;-)

I will close and open again after #4601 and #4602

PR #4625, opened 2024-07-16
Bugfix JS: sync newOL and OL maps

rldhont:
The OL2 map was not sync with newOL map, so when the viewport is resize the map comes back to the initial zoom.


nboisteault:
@rldhont Instead I propose https://github.com/3liz/lizmap-web-client/pull/4626

rldhont:
Superseeded by #4626

PR #4626, opened 2024-07-16
Fix: avoid zoom to initial extent on window resize

nboisteault:


rldhont:
Can you add tests to fully superseed #4625 ?

Issue #3965, opened 2023-11-16
[Bug]: treecreated event is fired before layer tree is created


mind84:
### What is the bug?

Hi,
I noticed that the `treecreated` event fires before the new `lizmap-treeview` web component is added to the DOM, wich happen [after](https://github.com/3liz/lizmap-web-client/blob/e83415b9066e103dc9cfb66afcb30b157fdec360/assets/src/legacy/map.js#L5023) `uicreated` event fires. 

Se image below:
![treeview](https://github.com/3liz/lizmap-web-client/assets/20710470/9343f5b4-31fc-492b-861b-cbf1ae3b3001)
The `treecreated` event appears to be designed to only work with the old Lizmap 3.6 tree view.

Any chances to port this event in LWC >= 3.7 ?

Thanks!

### Steps to reproduce the issue

add this external js to reproduce the issue:

[layertree_event.js.zip](https://github.com/3liz/lizmap-web-client/files/13374340/layertree_event.js.zip)

### Versions

**Versions :**

* Lizmap Web Client : 3.7.0-beta.1
* Lizmap plugin : 3.17.0
* QGIS Desktop : 3.28.7
* QGIS Server : 3.28.5
* Py-QGIS-Server : 
* QGIS Server plugin atlasprint : not found
* QGIS Server plugin lizmap_server : 2.8.1
* QGIS Server plugin wfsOutputExtension : 1.7.0

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Debian GNU/Linux trixie/sid

### Browsers

Firefox, Chrome

### Browsers version

last

cfsgarcia:
Same problem occurs in LWC 3.7.8. 
`$("lizmap-treeview")` returns nothing after `treecreated` or `uicreated` have been fired.

PR #4658, opened 2024-07-29
[Bugfix] Fire treecreated event at proper time

mind84:
Starting from versions 3.7 the `treecreated` event is raised before the `lizmap-treeview` component is properly initialized.

To fix that, the `treecreated` event must be fired after `uicreated` event is fired since `Treeview` component is updated in the `uicreated` listener (see [index.js](https://github.com/3liz/lizmap-web-client/blob/05dded203440cfecda5d98a2eb81f6c89a74573c/assets/src/index.js#L41))

This behavior might be confusing, as you might think that the `uicreated` event fires when all UI components are correctly initialized, which is not the case.

As a proposal, it might be useful to reconsider name of the uicreated event (e.g. `uiinitialized` or whatever) and fire the uicreated event when components initialization is completed (see index.js above)

**NOTE:** 
`treecreated` event is fired regardless the `getLegendGraphic` request since for heavy projects this request could take a lot of time.

Ticket : #3965

Funded by Faunalia

### DIFF
--- a/assets/src/components/Treeview.js
+++ b/assets/src/components/Treeview.js
@@ -198,6 +198,9 @@ export default class Treeview extends HTMLElement {
         mainEventDispatcher.addListener(
             this._onChange, ['resolution.changed']
         );
+
+        // layertree has been created, fire corresponding event
+        mainLizmap.lizmap3.events.triggerEvent('treecreated');
     }
 
     disconnectedCallback() {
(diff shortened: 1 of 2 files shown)


rldhont (approved):

nboisteault (approved):

Gustry:
Thanks @mind84 
Still something to do in #3965 ?

mind84:
>Still something to do in #3965 ?

Nope, it can be closed for me

Gustry:
Ok, you could add `fix #3965` in the PR description ;-)

PR #4666, opened 2024-07-31
Add link to the commit SHA on GitHub if possible

Gustry:
As a draft for now, I would have prefered to use the `project.xml` file https://github.com/3liz/lizmap-web-client/blob/master/lizmap/project.xml but it's triggering a warning in the Jelix toolbar 

> [Creation of dynamic property Jelix\Core\Infos\AppInfos::$commitId is deprecated](http://lizmap.local:8130/admin.php/admin/server_information#)
> [Creation of dynamic property Jelix\Core\Infos\AppInfos::$commit is deprecated](http://lizmap.local:8130/admin.php/admin/server_information#)

So I used the mainconfig.ini file. But it doesn't seem the perfect place. Anyway, feel free to let me know what you think.

We need to remove one of the "wrong" `sed` before merging

image


Gustry:
According to @nworr it's not possible to add new attributes or tags in the XML, so I removed the line

It was : 
 `sed -i "s@<commit id=\"\"/>@<commit id=\"$(COMMITID)\"/>@g" $(DIST)/lizmap/project.xml`

mdouchin (approved):

Gustry:
Let's give it a try on 3.9 snap first

PR #4679, opened 2024-08-01
Tests - Update phpunit version

Gustry:
Linked to https://github.com/3liz/lizmap-web-client/pull/4675

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -16,12 +16,15 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-version: [
-          '8.0',
-          '8.1',
-          '8.2',
-          '8.3',
-        ]
+        include:
+          - php-version: "8.0"
+            php-unit: "9.5.7"
+          - php-version: "8.1"
+            php-unit: "10.5.29"
+          - php-version: "8.2"
+            php-unit: "11.2.9"
+          - php-version: "8.3"
+            php-unit: "11.2.9"
     steps:
 
       - name: Check out repository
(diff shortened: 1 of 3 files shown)


Gustry:
Let's skip the failure about PHP 8.0 not supported by this release of PHPUnit. We can pin the PHP unit version per version of PHP if needed.

I see quite a few errors in the logs of PHP 8.2 for instance. @nworr Do you have any idea ?

nworr:
I add 2 commit to fix phpUnit 10 deprecated methods, seem to works for php > 8.0
(we could pin phpunit 9 for php 8.0 or skip it ?)

Gustry:
@nworr I tried in the same PR now to update to PHPUnit 11.2.9 (instead of PHPunit 10.X.X in #4683)
They are new errors now, if you wann have a look ;-)

I will try quickly to patch my PR to have a different version

Gustry:
Sorry, it's only deprecation notice https://github.com/3liz/lizmap-web-client/actions/runs/10316189433/job/28557866988?pr=4679 we can wait

nworr:
There's also a warning (since phpUnit 10) , not blocking but need to be investigated

Gustry:
@nworr OK for you ? There is now a different version of PHPUnit according to the version of PHP (because lower version of PHPUnit are not compatible with old version of PHP)

Gustry:
> There's also a warning (since phpUnit 10) , not blocking but need to be investigated

Yes, it would be nice to have a forward of these messages in the PR comment or GitHub comment, otherwise, we don't check manually these logs. I will try to have a look in another PR

nworr (approved):

PR #4683, opened 2024-08-07
Php unit 10 compatibility

nworr:
Fix phpunit deprecation warnings on phpunit 10, (see  https://github.com/3liz/lizmap-web-client/pull/4679)


Gustry (approved):
LGTM, thanks

I just added some backport labels. If it fails on 3.6, let's skip, only if it doesn't need more work ?

PR #4692, opened 2024-08-12
Remove some deprecated statements from Dockerfile

Gustry:
* https://docs.docker.com/go/dockerfile/rule/maintainer-deprecated/
* https://docs.docker.com/reference/build-checks/legacy-key-value-format/


docker/Dockerfile
```diff
@@ -5,7 +5,7 @@ COPY lizmap_web_client.zip .
 RUN unzip lizmap_web_client.zip
 
 FROM ${REGISTRY_PREFIX}alpine:3.17
```
Gustry, line 7:
```suggestion
```

Suggestion from @dmarteau

dmarteau: LGTM

PR #4705, opened 2024-08-23
PHP - Add helper for property_exists

Gustry:
Trying to add a small helper to avoid writing twice the key.

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/Project.php
+++ b/lizmap/modules/lizmap/lib/Project/Project.php
@@ -1387,6 +1387,22 @@ class Project
         return $layer_config;
     }
 
+    /**
+     * @param object $object   The configuration
+     * @param string $property String about the property to request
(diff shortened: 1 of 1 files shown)


Gustry:
I only updated two occurrences first, to see how it goes. They are more in the code.

About the nested explode, I think they is the use case, when we read new configuration from LWC 3.X.

lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -1387,6 +1387,28 @@ public function getLayerPolygonFilterConfig($layerName, $editing_context = false
         return $layer_config;
     }
 
+    /**
+     * @param object $object   The configuration
+     * @param string $property String with nested properties 'prop1.nestedProp.deepvalue' to fetch
```
laurentj, line 1392: I'm not sure we need to support nested properties.

rldhont (approved):

PR #4710, opened 2024-08-27
Tests - Update PHPStan to 8.0, to match minimum of PHPUnit

Gustry:


Gustry:
Unrelated test failure, also in master

PR #4712, opened 2024-08-27
Fix phpstan error detected sinc phpstan 1.11

nworr:
The 1.12  version of phpstan improve `preg_match  / preg_match_all` detection, some new errors appears.
Since preg_match_all always fill `$matches` with 2 item, phpstan consider `count($matches)  ` comparison  useless.
 
Refactoring use return value from preg_match_all instead of `count($matches) `


laurentj:
The change of the phpstan version is missing from `.github/workflows/php-stan.yml`

PR #4737, opened 2024-09-06
Try harder to find layers when deciding whether to cache them or not.

ppetru:
Had to do this to get some layers of my project to cache. Their names show up changed in the WMS request so they would not be cached. For example:

"Overall boundary" -> "Overall_boundary"
"2024-08-31" -> "l_2024-08-31"


ppetru:
If approved, it would be nice to get this included in a 3.8 release. Thanks!

rldhont (approved):

Gustry:
@nworr Where could be the place to use the helper #4705 ?

Gustry:
Hi @ppetru 

Thanks for your contribution.
Do not worry about "outdated branch", it's adding new commits on your branch (4 now), while one is enough. We will just squash and merge later.
Or do a rebase, it's better.

nworr:
> @nworr Where could be the place to use the helper #4705 ? 


Not sure the helper is very usefull here, it's not sure the `$configLayer` is not null
the result code may look like this
```php
$configLayer = $lproj->findLayerByAnyName($layername);
if (!is_null($configLayer) ) { 
    $clientCacheExpiration = $lproj->getPropertyOrDefault($configLayer, 'clientCacheExpiration') ;
    if (!is_null($clientCacheExpitation) {
         $rep->setExpires('+'.(int)$clientCacheExpiration.' seconds');
    }
}
````

rldhont:
@Gustry the helper is not available in release_3_7 and it is pivate. We can merge it to get backport to release_3_7. And we have to enhance the Project class to easily use your helper.

Gustry:
> @Gustry the helper is not available in release_3_7 and it is private.

I know, that's why I was asking where we should add this function to make it available in this kind of PR ;-)

PR #4635, opened 2024-07-17
[Feature] Add text widget to the edition form

mind84:
Since version 3.30, QGIS has introduced a [Text Widget](https://qgis.org/project/visual-changelogs/visualchangelog330/#feature-dynamic-text-widgets-with-current_value-support) feature available in layer form configuration using the "Drag and Drop" form designer.

`Lizmap` already supports this feature on `Popup` (see `lizmap-plugin` release [3.17.1](https://github.com/3liz/lizmap-plugin/releases/tag/3.17.1) and `qgis-lizmap-server-plugin` release [2.8.1](https://github.com/3liz/qgis-lizmap-server-plugin/releases/tag/2.8.1))

With this PR I would introduce Text Widget support also on the **edition form**.

NOTES:

- currently the behavior is **NOT** dynamic, it means that the `Text Widget` expression is not evaluated "live" if it depends on some other layer field, but only after the form is saved.
- If the Text Widget expression contains geometry references, then it can be evaluated only if the `Add geometry to feature response` option in the QGIS Server tab is checked.
image
- invalid expressions are not evaluated and will be returned as "raw expression" to the client .

Funded by Faunalia

PR #4752, opened 2024-09-16
e2e: fix test

nboisteault:


mind84:
Hi @nboisteault,

I cannot figured out why the click doesn't fire. so I've replaced the `click` with `focus` + `enter` key and seems to work.

Something like 

```js
        await page.locator('#jforms_view_edition__submit_submit').focus();
        await page.keyboard.press('Enter');
```

Could you try?

This change is included in this PR https://github.com/3liz/lizmap-web-client/pull/4635 but I could undo the change so we can use this PR to keep things organized.

Let me know

Thanks!

nboisteault:
@mind84 I think this is due to the BS5 tooltips behavior that has change. I'll take a look.

nboisteault:
I merged [this commit](767279d2c4009c849f08d23a8dac91375cbbdf6f) to put tooltip behavior aside.
I'm closing this PR.

mind84:
> can you modify them in order to fix this issue

Yes

Gustry:
Yes, I already notice that in some tests, we are adding data, but then, we do not remove or clean the dataset.
Some tests can't be launched twice.

Something we need to keep in mind.

Thanks for debugging.

nboisteault:
Maybe `load_sql.sh` has to be run in an `afterEach` for tests modifying data.

mind84:
@Gustry @nboisteault 

this https://github.com/3liz/lizmap-web-client/pull/4775 is probably enough to solve the n_to_m_relations e2e test issue.
 
I'll do a PR later to make the n_to_m_relations tests independent

Thanks

PR #4753, opened 2024-09-16
Bugfix Permalink: checked groups are not respected when hash is applied

rldhont:
When a layer is checked, all his parents are checked. So unchecked groups provided by permalink has to be unchecked after the layers and from leaf to root.

Capture d’écran 2024-09-16 à 14 15 10

Funded by Conseil Départemental du Calvados

### DIFF
--- a/assets/src/modules/Permalink.js
+++ b/assets/src/modules/Permalink.js
@@ -175,7 +175,10 @@ export default class Permalink {
 
         this._hash = ''+window.location.hash;
 
-        const items = mainLizmap.state.layersAndGroupsCollection.layers.concat(mainLizmap.state.layersAndGroupsCollection.groups);
+        // items are layers then groups from leaf to root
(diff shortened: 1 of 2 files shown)


assets/src/modules/Permalink.js
```diff
@@ -175,7 +175,10 @@ export default class Permalink {
 
         this._hash = ''+window.location.hash;
 
-        const items = mainLizmap.state.layersAndGroupsCollection.layers.concat(mainLizmap.state.layersAndGroupsCollection.groups);
+        // items are layers then groups from leaf to root
+        const items = mainLizmap.state.layersAndGroupsCollection.layers.concat(
+            mainLizmap.state.layersAndGroupsCollection.groups.reverse() // revers groups array to get from leaf to root
```
Gustry, line 180:
```suggestion
            mainLizmap.state.layersAndGroupsCollection.groups.reverse() // reverse groups array to get from leaf to root
```

PR #4754, opened 2024-09-16
Action: allow geometry drawn by user

nboisteault:
Funded by WPD

https://github.com/user-attachments/assets/b439917d-55b0-46c8-8ed7-1a4d3ac48e87

### DIFF
--- a/assets/src/components/ActionSelector.js
+++ b/assets/src/components/ActionSelector.js
@@ -71,7 +71,7 @@ export default class ActionSelector extends HTMLElement {
 
     onActionSelectChange(event) {
         // Remove previous digitizing tool if any
-        document.querySelector('#project-action-selector-container lizmap-digitizing')?.remove();
+        document.getElementById('action-digitizing')?.remove();
 
         // Get the host component
         let host = event.target.closest("lizmap-action-selector");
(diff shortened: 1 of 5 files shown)


rldhont (approved):

rldhont:
waiting for @mdouchin review!

mdouchin (changes requested):
Thanks for your work.
Can you please improve the UI a bit :

* use the same space between the left mini-dock border and the digitizing toolbar (and message underneath) as the space used for the combobox and "grey" description. It seems. Be careful about testing it with a without Bootstrap 5
* we shoud probably remove the grey background color of the description and use a lighter style, for example a thin blue border as used in the editing form group 
```css
border: 1px solid var(--color-contrasted-elements-light);
border-radius: 5px;
``` 

Thanks !

Gustry:
Sorry @nboisteault I just tried this PR.

> we should probably remove the grey background color

I can only agree with this proposition, the current grey background is legacy and still present.

AS talked with @rldhont and @mdouchin, if the `geometry` is set for the action, so the geometry is **required**. We shouldn't be able to launch the action without a geometry.
I think the "Launch" button should be disabled until we do not have a geometry. Otherwise, we have the error message set the SQL function code :

image

PR #4455, opened 2024-05-21
Bootstrap 5

nboisteault:
Funded by 3Liz

PR #4769, opened 2024-09-18
Request GetFeatureInfo with Bootstrap5

Gustry:
No tested, no idea.

Just to think about https://github.com/3liz/lizmap-plugin/pull/604 and https://github.com/3liz/qgis-lizmap-server-plugin/pull/87 about the migration from BootStrap 2 to BootStrap 5 in #4455

CC @mdouchin @nboisteault and @mind84

### DIFF
--- a/lizmap/modules/lizmap/lib/Request/WMSRequest.php
+++ b/lizmap/modules/lizmap/lib/Request/WMSRequest.php
@@ -468,7 +468,7 @@ class WMSRequest extends OGCRequest
         $this->params['with_geometry'] = 'true';
         // Starting from LWC 3.9, we need to request the maptip with Bootstrap 5
         // TODO, remove later when the lizmapWebClientTargetVersion=30800 because all projects will be migrated.
-        $this->params['CSS_FRAMEWORK'] = 'BOOSTRAP5';
+        $this->params['CSS_FRAMEWORK'] = 'BOOTSTRAP5';
 
         // Get remote data
         $response = $this->request(true);
--- a/lizmap/modules/lizmap/lib/Request/WMSRequest.php
+++ b/lizmap/modules/lizmap/lib/Request/WMSRequest.php
@@ -466,6 +466,9 @@ class WMSRequest extends OGCRequest
         $this->params['with_maptip'] = 'true';
         // Always request geometry to QGIS server so we can decide if to use it later
         $this->params['with_geometry'] = 'true';
+        // Starting from LWC 3.9, we need to request the maptip with Bootstrap 5
+        // TODO, remove later when the lizmapWebClientTargetVersion=30800 because all projects will be migrated.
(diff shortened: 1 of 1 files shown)


mind84:
@Gustry,

this `$projectInfos->version` returns a string, something like `"3.9.0-pre"`. I think the version could easily be extracted from this string and converted in number directly on php side. If it's reliable, form me is good then

Thanks

Gustry:
I will first wait for @nworr and @laurentj about any slowdown if it's done at each HTTP request ?

For the string versus number, the advantage of strings is that it's more precise (beta, rc etc) if we need it later on Python side.

We can still use the string to decide between BS2 or BS5.

Gustry:
After a discussion with @nworr :

* it's too expensive to load this XML to know this version on each HTTP requests

Either, only do this on a GetFeatureInfo request, when we need it.

Or even easier, add a `BOOTSTRAP_VERSION=5` in the HTTP request. It's a hardcoded string.

nworr:
Maybe a more generic variable name ? `CSS_FRAMEWORK=BOOSTRAP5` 

https://github.com/3liz/lizmap-web-client/blob/a051777969cd2a6e12f32d7a93a4e47f1e9e23fc/lizmap/modules/lizmap/lib/Request/WMSRequest.php#L469

Gustry:
PR updated and ready I guess.

It needs the branch https://github.com/3liz/qgis-lizmap-server-plugin/pull/87

mind84:
@Gustry,

> It needs the branch [3liz/qgis-lizmap-server-plugin#87](https://github.com/3liz/qgis-lizmap-server-plugin/pull/87)

I see you updated the PR, can I do something?

Thanks

Gustry:
No, I just pushed to your branch just now. Tests are now passing.

This PR https://github.com/3liz/lizmap-web-client/pull/4769, and desktop side https://github.com/3liz/lizmap-plugin/pull/604 and server side https://github.com/3liz/qgis-lizmap-server-plugin/pull/87 are now ready ;-)

lizmap/modules/lizmap/lib/Request/WMSRequest.php
```diff
@@ -466,6 +466,9 @@ protected function process_getfeatureinfo()
         $this->params['with_maptip'] = 'true';
         // Always request geometry to QGIS server so we can decide if to use it later
         $this->params['with_geometry'] = 'true';
+        // Starting from LWC 3.9, we need to request the maptip with Bootstrap 5
+        // TODO, remove later when the lizmapWebClientTargetVersion=30800 because all projects will be migrated.
+        $this->params['CSS_FRAMEWORK'] = 'BOOSTRAP5';
```
Gustry, line 471:
```suggestion
        $this->params['CSS_FRAMEWORK'] = 'BOOTSTRAP5';
```

nworr (approved):

PR #4773, opened 2024-09-19
UX - Provide a HTTP link about the main plugin homepage/help

Gustry:
Needs https://github.com/3liz/qgis-lizmap-server-plugin/commit/c001bac718e9bf0ee2342cbb3cbf4bcb04eded2b on the Lizmap server side

Each plugin can provide the main help page, like for French Cadastre plugin, it leads to https://docs.3liz.org/QgisCadastrePlugin/ 

The link is displayed with the list of plugins installed on the server.

The next step would be to have the same behavior when a module is installed. When PgMetadata module is installed, we would be able to have a look to https://docs.3liz.org/qgis-pgmetadata-plugin/

CC @nworr and @mdouchin


Gustry:
I will manually backport to 3.8, because of bootstrap

nboisteault (approved):

PR #4768, opened 2024-09-18
[Feature] Popup: move related child in corresponding popup_lizmap_dd_relation div

mind84:
When a `Popup` is displayed with the QGIS Drag and Drop configuration, for each relations configured in the layout a container HTML element (`div.popup_lizmap_dd_relation`) is added to the Popup (see https://github.com/3liz/qgis-lizmap-server-plugin/commit/6a4240041858883a83baad08012a0604b19c8873)

This PR moves the related child (or children) popup in corresponding `popup_lizmap_dd_relation` div. The feature covers both `1:n` and `n:m` relations 

Ticket : fix #4407

Backport to 3.8

Funded by Faunalia

PR #4775, opened 2024-09-20
Bootstrap 5 tooltip on FeatureToolbar component

mind84:
Follow this PR https://github.com/3liz/lizmap-web-client/pull/4752

Linked to https://github.com/3liz/lizmap-web-client/pull/4768 and https://github.com/3liz/lizmap-web-client/pull/4635

Funded by Faunalia

### DIFF
--- a/assets/src/components/FeatureToolbar.js
+++ b/assets/src/components/FeatureToolbar.js
@@ -121,11 +121,6 @@ export default class FeatureToolbar extends HTMLElement {
 
         render(this._mainTemplate(), this);
 
-        // Add tooltip on buttons
-        $('.btn', this).tooltip({
-            placement: 'top'
-        });
-
         this._editableFeaturesCallBack = (editableFeatures) => {
             this.updateIsFeatureEditable(editableFeatures.properties);
             render(this._mainTemplate(), this);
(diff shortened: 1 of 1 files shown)


nboisteault:
Hi @mind84, sorry but the test still fail with this PR.

mind84:
> Hi @mind84, sorry but the test still fail with this PR.

yes, I see, very annoying :D

I'll put this on draft and push some version of update n_to_m_relation.spec.js with changes you suggested in #4752

Let's see

Thanks

mind84:
@nboisteault 

I think I did it without changing the` n_to_m_relations.spec.js`, so I can keep the PR separate.

Issue is related `unlink` option on the feature toolbar and its `tooltip`. The tooltip il placed in absolute position within the `body`, so after clicking the unlink button the corresponding `feature toolbar` is removed from DOM as a consequence of the attribute table update. This, apparently, leaves the tooltip orphan on the page (which is unexpected since the trigger is forced on `hover`). 

So I specify also the `container` option on tooltip initialization, to force the tooltip to be within the toolbar.

I think that for the feature toolbar component this behavior could be considered correct, I don't know for others tooltip

nboisteault:
@mind84 great investigation!
Thank you for this PR.

nboisteault (approved):

PR #4776, opened 2024-09-20
Count user defined js , adapt message when js exception

nworr:
When user is session is kind of admin , and if some user defined script are loaded, the javascript exception catcher add info about user scripts.

Screenshot 2024-09-23 at 12-26-06 Montpellier - Transports - Demo - Lizmap

edit : updated screenshot

it adds a new parameter in the URL `no_user_defined_js=1`

Funded by 3Liz

### DIFF
--- a/lizmap/modules/view/controllers/lizAjax.classic.php
+++ b/lizmap/modules/view/controllers/lizAjax.classic.php
@@ -187,6 +187,9 @@ class lizAjaxCtrl extends jController
             $lizUrls['removeCache'] = jUrl::getFull('admin~maps:removeLayerCache');
         }
 
+        if (jAcl2::check('lizmap.admin.access') || jAcl2::check('lizmap.admin.server.information.view')) {
+            $lizUrls['repositoryAdmin'] = jUrl::getFull('admin~maps:index');
+        }
         $content = '<script type="text/javascript" src="'.jUrl::getFull('view~translate:index').'"/>'."\n";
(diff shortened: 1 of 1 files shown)


lizmap/modules/view/locales/en_US/dictionnary.UTF-8.properties
```diff
@@ -1,4 +1,5 @@
 startup.error=An error occurred while loading this map. Some necessary resources may temporarily be unavailable. Please try again later.
+startup.user_defined_js=There's some user defined javascript scripts, that can be the cause of this message, try disabling them.
```
Gustry, line 2:
```suggestion
startup.user_defined_js=There's some user defined JavaScript scripts, that can be the cause of this message, try disabling them.
```

According to Wikipédia https://fr.wikipedia.org/wiki/JavaScript

Gustry (approved):
Very nice addition !

I got a few ideas, I will have a look after if you want :

* Include a link to the maps management page ? similar https://github.com/3liz/lizmap-web-client/blob/a051777969cd2a6e12f32d7a93a4e47f1e9e23fc/lizmap/modules/view/templates/main.tpl#L44C1-L44C118
* Include a link to the same map, but with a kind of flag `&no_user_defined_js=true`

rldhont (approved):

Gustry (approved):
Looks great !

PR #4784, opened 2024-09-24
Tests - Add E2E test about user defined JavaScript warning

Gustry:
Follow up from #4776 

This test project was not used for loading the map, only on the landing page. I just added a layer in the project, to not have an empty tree view.

But the E2E test was expecting to have an empty tree view if there is an error (because we did only some fake 500 error on GetCapabilities @rldhont ?)

**Not an administrator**

image

**As administrator**

image

### DIFF
--- a/lizmap/modules/view/locales/en_US/dictionnary.UTF-8.properties
+++ b/lizmap/modules/view/locales/en_US/dictionnary.UTF-8.properties
@@ -1,8 +1,10 @@
 startup.error=An error occurred while loading this map. Some necessary resources may temporarily be unavailable. Please try again later.
-startup.user_defined_js=There's some user defined JavaScript scripts, that can be the cause of this message, try disabling them.
-startup.goToProject=Go back to the home page.
-startup.goToRepositoryAdmin=Go to maps management page
-startup.projectWithoutJSLink=Project with user's scripts disabled
+startup.error.administrator=Please notify the administrator of this map to <strong>connect</strong> and <strong>visit</strong> this map.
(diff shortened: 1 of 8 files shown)


Gustry:
@nworr I have displayed the developer hint only if no additionnal JS were loaded. Maybe it would have been better with checking the flag ? Still not 100% convinced, but I think it's a small hint to display. @mdouchin ?

But I don't know, the E2E is failing on CI, not on local.
And the link to "Maps management" is now undefined : http://localhost:8130/index.php/view/undefined

lizmap/modules/view/locales/en_US/dictionnary.UTF-8.properties
```diff
@@ -1,8 +1,10 @@
...
-startup.user_defined_js=There's some user defined JavaScript scripts, that can be the cause of this message, try disabling them.
-startup.goToProject=Go back to the home page.
-startup.goToRepositoryAdmin=Go to maps management page
-startup.projectWithoutJSLink=Project with user's scripts disabled
+startup.error.administrator=Please notify the administrator of this map to <strong>connect</strong> and <strong>visit</strong> this map.
+startup.error.developer.tools=Maybe you will have a clue by opening your "Developer tools" in your web-browser. It's usually <kbd>F12</kbd>. Check "Console" and "Networks".
+startup.user_defined_js=This map contains some users additional JavaScript scripts. This can be the cause of this message, try disabling them.
+startup.goToProject=Home
+startup.goToRepositoryAdmin=Maps management page
+startup.projectWithoutJSLink=Project without additional JavaScript
```
mdouchin, line 7:
```suggestion
startup.projectWithoutJSLink=This map without additional JavaScript
```

mdouchin (approved):

PR #4785, opened 2024-09-24
Fix: allow tiled wms when `singleTile` is `false` and layer is not ca…

nboisteault:


nboisteault:
@rldhont ready for review

nboisteault:
@rldhont I changed https://github.com/3liz/lizmap-web-client/pull/4785/files#diff-bf6e96d0d15dec384eee57764b40c9215472541f9c6b333d202590d5b291c483R96

Is it ok for you?

PR #4786, opened 2024-09-25
Backport 3.7 : Count user defined js , adapt message when js exception

nworr:
count user js script and display message if js exception and user is kind of admin

Funded by 3Liz


Gustry (approved):
LGTM

Backport from #4776

Tests E2E coming in #4784

PR #4801, opened 2024-09-27
E2E - Fix version of Py-QGIS-Server

Gustry:


Gustry:
CC @dmarteau Even in 3.34 tag of the docker image, the version is `n/a` now.

PR #4804, opened 2024-09-27
Show Lizmap modules infos

nworr:
Funded by 3Liz
image

PR #4816, opened 2024-09-30
fix typo, improve html for module view

nworr:
improve #4804 

Funded by 3Liz

### DIFF
--- a/lizmap/modules/admin/templates/server_information.tpl
+++ b/lizmap/modules/admin/templates/server_information.tpl
@@ -171,7 +171,7 @@
             {* Fixed in lizmap_server plugin 1.3.2 https://github.com/3liz/qgis-lizmap-server-plugin/commit/eb6a773ba035f877e9fa91db5ef87911a2648ee1 *}
             <th style="width:20%;">
                 {$version['name']}
-                {if array_key_exists('homepage', $version)}
+                {if array_key_exists('homepage', $version) && $version['homepage']}
                     <a href="{$version['homepage']}" target="_blank"><span class='badge rounded-pill bg-secondary'>{@admin.server.information.qgis.plugin.help@}</span></a>
                 {/if}
             </th>
--- a/lizmap/modules/admin/templates/server_information.tpl
+++ b/lizmap/modules/admin/templates/server_information.tpl
@@ -37,17 +37,21 @@
     {if empty($modules)}
       <p>{@admin.server.information.no.module@}</p>
     {else}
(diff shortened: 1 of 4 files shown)


lizmap/modules/admin/templates/server_information.tpl
```diff
@@ -37,17 +37,21 @@
...
             <th>{@admin.server.information.module@}</th>
             <th>{@admin.server.information.module.version@}</th>
         </tr>
+        </thead>
+        <tbody>
         {foreach $modules as $module}
             <tr>
-                <th style='width:20%;'> {$module->slug}</th>
-                <td style='width:20%;'>{$module->version}</td>
+                <th>{$module->slug}</th>
```
Gustry, line 50: Not sure why, I took it from L165, it was this.
Before, it was 30% from your initial commit I think.

lizmap/modules/admin/templates/server_information.tpl
```diff
@@ -167,7 +171,7 @@
             {* Fixed in lizmap_server plugin 1.3.2 https://github.com/3liz/qgis-lizmap-server-plugin/commit/eb6a773ba035f877e9fa91db5ef87911a2648ee1 *}
             <th style="width:20%;">
                 {$version['name']}
-                {if $version['homepage']}
+                {if array_key_exists('homepage', $version)}
```
Gustry, line 174: Hum, now, the `homepage` is always provided, but the string can be empty I think.
```suggestion
                {if array_key_exists('homepage', $version) && $version['homepage']}
```
??

indeed, nice to check if the key exists anyway.

lizmap/modules/admin/templates/server_information.tpl
nworr: I remove inline style, and prefer to use `table-server-info` which is used in all other tables of this page

lizmap/modules/admin/templates/server_information.tpl
nworr: Home page is undefined for me : see jelix debug bar:
 image

Gustry: Ok, thanks.
So should we do both ? if the string is existing, but empty ?

Gustry (approved):
LGTM

Just E2E is failing, because of the new HTML table header

lizmap/modules/admin/templates/server_information.tpl
nworr: yeah that's it, need to keep both.

PR #4819, opened 2024-10-01
Parsing QGIS Project with PHP XMLReader instead of DOM

rldhont:
Instead of using PHP SimpleXml API to parse QGIS Project, we uses XMLReader. The reason to change it, it is to have something faster and lighter.


rldhont:
@mdouchin @nworr ready for review

PR #4836, opened 2024-10-04
[Bug] Enclose wms/wfs filters in brackets to guarantee filters logic

mind84:
Currently client filters and login filters are merged 'as is' before system sends WMS/WFS request to the server.

This could lead to unexpected behavior.

Take, for instance, the case where you have set some login filters in a layer and this layer is displayed as child of another one in the feature popup.

Suppose you want to limit the visibility of `child_layer` only to users belonging to `group1` and use a `lizmap_group` column on `child_layer` to achieve this.

So, if a user belonging to `group1` logs into the project and requests a `parent_layer` popup, the desired behavior is:

```
show only children related to the parent
AND
show only children that current user can see according to login filters
```
Current filter is issued like this:
```
"lizmap_group" = 'group1' OR "lizmap_group" = 'all' AND "fid_parent" = '2'
```
The filter concatenation in this case does not produce the desired response because the execution of `OR` and `AND` statements does not respect filters logic. 
This will result in a response containing all child features with `"lizmap_group"='all'` AND all features with `"lizmap_group"='group1'`, regardless of the `parent_layer` record to which they are related. 
In particular cases, this could result in thousands of features, or more, being incorrectly sent for each request.

correct filters should be:
```
( "lizmap_group" = 'group1' OR "lizmap_group" = 'all' ) AND "fid_parent" = '2' 
```

To fix this I simply enclose the filters in parentheses before they are merged. In some cases the result is a bit verbose:

```
( "lizmap_group" = 'group1' OR "lizmap_group" = 'all' ) AND ( "fid_parent" = '2' )
```
but ensures that the feature filter works properly.

Backport to 3.8, if possible.

Funded by Faunalia

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/Project.php
+++ b/lizmap/modules/lizmap/lib/Project/Project.php
@@ -1364,7 +1364,7 @@ class Project
 
             $filters[$layerName] = array_merge(
                 (array) $loginFilteredConfig,
-                array('filter' => $filter, 'layername' => $lName)
+                array('filter' => '( '.$filter.' )', 'layername' => $lName)
             );
         }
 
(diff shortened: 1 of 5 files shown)


Gustry (approved):

rldhont (approved):
Thanks @mind84 !

mind84:
@rldhont ,

> backport to 3.7 ?

Yes I think should be backported.

Thanks!

PR #4839, opened 2024-10-04
Deactivate Bootstrap 2 modal dialogs

mdouchin:
This is needed for LWC >=3.9 and the migration to bootstrap 5
This CSS addition prevents old JS scripts from making the map unusable
(example of deprecated scripts: https://github.com/3liz/lizmap-javascript-scripts/blob/master/library/ui/popup_metadata_info/popup.js )

We hide the `#lizmap-modal` div and sibling if it has a direct modal-header child div (Bootstrap 2)

Bootstrap 5 needs a `div.modal-dialog` which itself contains a `div.modal-header` and a `div.modal-body`
. In this case, the modal will not be hidden

Funded by 3Liz

### DIFF
--- a/lizmap/www/assets/css/map.css
+++ b/lizmap/www/assets/css/map.css
@@ -3702,3 +3702,19 @@ div.lizmap-features-table.popup-displayed div.lizmap-features-table-container + 
 .hide {
   display: none;
 }
+
+
+/* Deactivate the OLD bootstrap-2 modal
+This is needed for LWC >=3.9 and the migration to bootstrap 5
+This prevent old JS scripts to make the map unusable
+ex: https://github.com/3liz/lizmap-javascript-scripts/blob/master/library/ui/popup_metadata_info/popup.js
+
+We hide modal if it has a direct modal-header child div (Bootstrap 2)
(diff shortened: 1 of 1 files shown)


lizmap/www/assets/css/map.css
```diff
@@ -3702,3 +3702,19 @@ div.lizmap-features-table.popup-displayed div.lizmap-features-table-container +
...
+/* Deactivate the OLD bootstrap-2 modal
+This is needed for LWC >=3.9 and the migration to bootstrap 5
+This prevent old JS scripts to make the map unusable
+ex: https://github.com/3liz/lizmap-javascript-scripts/blob/master/library/ui/popup_metadata_info/popup.js
+
+We hide modal if it has a direct modal-header child div (Bootstrap 2)
+Bootstrap 5 needs a div.modal-dialog which itself
+contains a modal-header and modal-body
+so the new modal will not be hidden
+*/
```
Gustry, line 3716:
```suggestion
These lines will be removed in the future, when the minimum target version will be 3.9
*/
```

nworr (approved):

PR #4840, opened 2024-10-04
IGN search string length must be between 3 and 200 chars

nboisteault:
This can be tested here: https://data.geopf.fr/geocodage/completion/?text=fo&type=StreetAddress&maximumResponses=10&bbox=3.875336983791886,43.61120191129432,3.8782264085648497,43.613350750790985

Funded by 3Liz


nboisteault:
I won't backport in 3.7 as it is not critical.

PR #4847, opened 2024-10-07
[Bug] Add embedded layers to keyValueConfig object

mind84:
Embedded layers are not taken into account when `Lizmap` requires the the `key value configuration` for attribute table display.

I did a little refactoring on handling embedded layers during projects loading (it's in a separated commit, but it is strictly related to this bug)

Funded by Faunalia


mind84:
Backport in 3.8, if possible.

Thanks

Gustry:
Sorry, I just renamed the project, to not only cover "editing capabilities" in #4846. Can you rebase and update the project name ?

lizmap/modules/lizmap/lib/Project/QgisProject.php
```diff
@@ -644,27 +649,24 @@ public function xpathQuery($query)
      *
      * @deprecated
      *
-     * @param mixed      $layerId
-     * @param null|array $embeddedRelationsProjects reference to associative array path(key) - QgisProject instance (value)
+     * @param mixed $layerId
      *
      * @return \SimpleXMLElement[]
      */
-    public function getXmlLayer($layerId, &$embeddedRelationsProjects = null)
```
Gustry, line 652: Is this parameter was added in 3.8 or before ?

I read a few lines before L645 650 that this method is deprecated already. But I was wondering if we should mark this attribute as deprecated as well instead of removing ? I haven't checked if this function was called in another Lizmap module with 2 parameters ? (and it's an API break, all modules must be checked)

List of public modules : https://packagist.org/packages/lizmap/

just asking

rldhont (approved):

lizmap/modules/lizmap/lib/Project/QgisProject.php
mind84: Hi,

I added this parameter myself in this PR https://github.com/3liz/lizmap-web-client/pull/4093. At the time creating a class property to store the list of embedded projects seemed excessive to me, but due to this bug I decide to refactor this part and create a dedicated property instead of passing the list along the class methods. It seems cleaner to me and probably more testable.

Given that, I think it is not a braking change

Issue #4169, opened 2024-02-06
COG Raster as Embed layers are not displayed


arno974:
### What is the bug?

COG raster layer from a parent QGIS project in a children QGIS  are correctly displayed in QGIS but not ni Lizmap. 

For example my "parent" QGIS project contains 3 raster layers. 2 are cog raster and one is a gpkg raster.

![image](https://github.com/user-attachments/assets/0de297f3-59d1-444d-a3df-7eb8a75c6982)

My "children" QGIS project contains all the embed layers from the "parent" project and one raster layer added as a "normal" raster layer :

![image](https://github.com/user-attachments/assets/756ec797-16f3-43f5-962f-67d1d0501dc1)

In Lizmap, the layers from the "parent" project display perfectly. However, in the child project, only the GPKG embed layers and the raster layer directly integrated are displayed.

![image](https://github.com/user-attachments/assets/3d2bd5d6-aede-481f-a208-b296049d8235)

This used to work in previous 3.6 version. It was very useful as it was possible to centralize the tiles cache as they were generated for the "master" basemap project.

### Steps to reproduce the issue

1. Add an embed layer (or a group of layers) as basemap
2. Layers are correctly displayed in QGIS
3. Check in Lizmap, layers are not available

**Versions :**

* Lizmap Web Client : 3.7.11
* Lizmap plugin : 4.4.2
* QGIS Desktop : 3.28.11
* Lizmap.com : Yes
* QGIS Server : 3.34.11
* Py-QGIS-Server : 1.9.1
* QGIS Server plugin atlasprint : 3.4.1
* QGIS Server plugin cadastre : 1.20.0
* QGIS Server plugin lizmap_server : 2.11.0
* QGIS Server plugin wfsOutputExtension : 1.8.2

<details>
<summary>List of Lizmap Web Client modules :</summary>
<br/>
* Version Lizmap Web Client 3.8 needed
</details>

<details>
<summary>List of safeguards :</summary>
<br/>
* Mode : normal
* Allow parent folder : no
* Prevent other drive : yes
* Prevent PG service : yes
* Prevent PG Auth DB : yes
* Force PG user&pass : yes
* Prevent ECW : yes
</details>

### Check Lizmap plugin

- [ ] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu

### Browsers

Firefox

### Browsers version

122.0

[...]

Issue #3954, opened 2023-11-03
[Bug]: Embed layers in a project disapear in Lizmap


arno974:
### What is the bug?

Hi,

I notice, that embed layers from a master project are not displayed in an other project if the layer has space or accent in his name.

It used to be fine before and it was working perfectly. I can't tell you when, but in my recent project I can't use embed layers anymore as they tends to have many problems.

Have you already noticed these kind of issues ?

### Steps to reproduce the issue

1. add in another QGIS project embed layers from another project (try to have some layers with space and accent in their name)
2. Add the project in Lizmap
3. Opening the master project in Lizmap, layers are well displayed
4. Opening the other project using embed layers, most of them are not available

### Versions

**Versions :**

* Lizmap Web Client : 3.6.7
* Lizmap plugin : 3.18.1
* QGIS Desktop : 3.16.11
* QGIS Server : 3.16.11
* Py-QGIS-Server : not used
* QGIS Server plugin atlasprint : 3.3.2
* QGIS Server plugin cadastre : 1.18.0
* QGIS Server plugin lizmap_server : 2.8.1
* QGIS Server plugin wfsOutputExtension : not found

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 22.04.3 LTS

### Browsers

Firefox

### Browsers version

Firefox

r9zzai:
Maybe its related to otdated lizmap plugin, please update. See [here](https://github.com/3liz/lizmap-web-client/issues/3734).

arno974:
Just an update to specify that this problem still remain in the 3.7 Lizmap version

PR #4857, opened 2024-10-09
Tests - Add raster embed layer in child project

Gustry:
Tests - Add raster embed layer in child project
It works in the parent project, but not in the child project.

Linked to issue #3954 #4169

CC @arno974


arno974:
I've checked your embed_child project and it does not contains the "Plan IGN" layers from the parent project (embed_parents). I've added the "Plan IGN" layers in my embed_child project and then opened this project in the Lizmap web client. The IGN layer (which contains spaces in its name) does not display in the child project in lizmap web client.

Gustry:
Thanks @arno974 for testing this PR.

It helps a lot when the test data is already included in the LWC repository, to replicate the issue indeed. I have updated the project with the "IGN plan" layer

Gustry:
Note the accent and space was with the other layer `r@stér fîlê` ;-)

PR #4896, opened 2024-10-21
UX - Rename the 'only map' option to 'disable landing page'

Gustry:
image

IMHO "Only maps" is not really meaningful and it's hard to understand what it's about.

I propose to rephrase "Disable landing page". I would have prefered to not "negate" the sentence, but the boolean for now is "On" for "Disabling the landing page".

Note for my self, I should check the other tooltip, when not in form mode.


Gustry:
Quick analysis, it seems not possible to have the tooltip when not in editing form mode, so I removed the other help string.

nworr (approved):

PR #4921, opened 2024-10-29
Fixing sorting order of the FeaturesTable component

neo-garaix:
* Fix **sortingOrder** value in **FeaturesTable**
* Fix the **end2end playwright test**

Funded by 3liz


mdouchin (approved):
thanks

PR #4922, opened 2024-10-29
FeaturesTable improvements

neo-garaix:
### Add

* Replaced some "div" by "table" tags such as "tr", "th" ...
* Added columns compatibility : now we can add several columns through some `<lizmap-field ... > ... </lizmap-field>`
* Added header labels of columns
* Added regex verification for expressions in fields. For now, we only accept expressions in fields like "libsquart" and not expressions like rand(10,34).

### Fix

* Fixed the **end2end playwright test** about FeaturesTable

Funded by 3liz


mdouchin (changes requested):
Please add a `<td>` inside the `<tr>` to anticipated further column usage. Hopefully this will not break the current features

mdouchin (approved):
👍

mdouchin:
Please rebase your commits into one single commit (with a new commit message and description) and it will be mergeable
Thanks

rldhont (approved):

assets/src/components/FeaturesTable.js
```diff
@@ -65,6 +71,9 @@ export default class FeaturesTable extends HTMLElement {
         // Features
         this.features = [];
 
+        // Field JSON
+        this.fieldJSON = null;
```
mdouchin, line 75: Please rename to `additionalFields`

assets/src/components/FeaturesTable.js
```diff
@@ -82,8 +91,22 @@ export default class FeaturesTable extends HTMLElement {
...
+        if (!this.isFieldJSONEmpty()) {
+            uniqueFieldJSON = '{';
+
+            this.fieldJSON.field.forEach(field => {
+                uniqueFieldJSON += `"${field.alias}": "${field.expression}",`;
+            });
+
+            uniqueFieldJSON = uniqueFieldJSON.slice(0, -1);
+            uniqueFieldJSON += '}';
+        }
```
mdouchin, line 106: Please rename also and dot not create a string but a JavaScript object which will be transformed into JSON in the `getFeatures` method

assets/src/components/FeaturesTable.js
```diff
@@ -396,6 +423,27 @@ export default class FeaturesTable extends HTMLElement {
...
+                    return;
+                }
+                jsonField += '{ "alias" : "' + btoa(field.getAttribute('alias')) + '",' +
+                              ' "expression" : "' + field.getAttribute('expression') + '",' +
+                              ' "description" : "' + field.getAttribute('description') + '"},';
+                field.remove();
+            });
+
+            jsonField = jsonField.slice(0, -1);
+            jsonField += ']}';
```
mdouchin, line 442: Please also use a JS object instead of concatenating strings

assets/src/components/FeaturesTable.js
```diff
@@ -452,6 +504,120 @@ export default class FeaturesTable extends HTMLElement {
...
+     * @returns {TemplateResult<1>} The labels of the table
+     */
+    buildLabels() {
+        if (this.isFieldJSONEmpty()) {
+            return html``;
+        }
+
+        let result;
+
+        if (!this.isFieldJSONEmpty()) {
```
mdouchin, line 565: No need to use this test since the function returns empty HTML juste above

assets/src/modules/FeaturesTable.js
```diff
@@ -32,13 +32,14 @@ export default class FeaturesTable {
      * @param {string|null} filter An QGIS expression filter
      * @param {boolean}     withGeometry If we need to get the geometry
      * @param {string|null} fields List of field names separated by comma
+     * @param {string|null} fieldJSON JSON object with the field names and expressions
```
mdouchin, line 35: Not a string, must be an object

assets/src/modules/FeaturesTable.js
```diff
@@ -32,13 +32,14 @@ export default class FeaturesTable {
...
      *
      * @returns — A Promise that resolves with the result of parsing the response body text as JSON.
      * @throws — {ResponseError} In case of invalid content type (not application/json or application/vnd.geo+json) or Invalid JSON
      * @throws — {HttpError} In case of not successful response (status not in the range 200 – 299)
      * @throws — {NetworkError} In case of catch exceptions
      */
-    getFeatures(layerId, filter = null, withGeometry = false, fields = 'null') {
+    getFeatures(layerId, filter = null, withGeometry = false, fields = 'null', fieldJSON = 'null') {
```
mdouchin, line 42: use null instead of 'null'

lizmap/modules/lizmap/controllers/features.classic.php
```diff
@@ -154,6 +154,13 @@ public function displayExpression()
         // Filter
         $exp_filter = trim($this->param('exp_filter', 'FALSE'));
 
+        // FieldJSON
+        $fieldJSON = json_decode($this->param('fieldJSON', 'null'), true);
```
mdouchin, line 158: Use try/catch to be sure the `json_decode` succeeds. If not, use an empty PHP array on error for this field

mdouchin (changes requested):
☕

assets/src/components/FeaturesTable.js
```diff
@@ -17,6 +17,12 @@ import { mainLizmap, mainEventDispatcher } from '../modules/Globals.js';
  * @element lizmap-features-table
  * @fires features.table.item.dragged
  * @fires features.table.rendered
+ * <lizmap-features-table draggable="yes"sortingorder="asc" sortingfield="libsquart"
+ *                        withgeometry="1" expressionfilter="quartmno = 'HO'"
+ *                        uniquefield="id" layerid="sousquartiers_24ceec66_e7fe_46a2_b57a_af5c50389649"
+ *                        layertitle="child sub-districts" id="0782b34c-840c-4b0f-821c-1b66c964e371">
+ *      <lizmap-field alias="Âge du quartier" expression="rand(12,56)" description="Libellé de l'âge du quartier"></lizmap-field>
```
Gustry, line 24: Can you switch to English for these examples ?

I was wondering how behaves JSDoc for documenting this class with this HTML, if it can be auto-documented for https://docs.3liz.org/lizmap-web-client/js/module-FeaturesTable.html and later when we will let our users to write their HTML Element https://jsdoc.app/

PR #4925, opened 2024-10-31
[Bug]: Add layer opacity information to rasters

mind84:
`Opacity` property is not taken into account on `raster` layers.

**QUESTION REGARDING UNIT TESTS:**

I replaced the php `clone` method used in `QgisProjectTest.php` inside `testSetLayerOpacity` function with `unserialize/serialize`.

It seems to me that the `clone` method doesn't work here since `layers` property is a `StdClass` instance and cannot be cloned this way (?). Basically the test always passes because:

https://github.com/3liz/lizmap-web-client/blob/fc239c68428074326f7cacb983f9fc4f1b50707e/tests/units/classes/Project/QgisProjectTest.php#L394-L396

on line `396` the `$cfg` object is constructed with the `opacity` property already set (added in previous line).

Can you confirm this behavior?

Backport to 3.8, if possible.

Funded by Faunalia

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/QgisProject.php
+++ b/lizmap/modules/lizmap/lib/Project/QgisProject.php
@@ -377,13 +377,18 @@ class QgisProject
      */
     protected function setLayerOpacity(ProjectConfig $cfg)
     {
-        $layerWithOpacities = $this->xpathQuery('//maplayer/layerOpacity[.!=1]/parent::*');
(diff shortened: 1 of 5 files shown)


rldhont (approved):

rldhont:
@mind84 thanks ! It seems that print e2e test has to be updated.

mind84:
> @mind84 thanks ! It seems that print e2e test has to be updated.

yep, I'm on it.

Thanks!

PR #4931, opened 2024-11-04
Tests - Update redis version from 5 to 7.4

Gustry:
The latest version of Redis 5 is EOL since April 2020 :/

We need to update a few docker compose file when we find it. I was hesitating to use `latest`.

CC @rldhont @mdouchin @nboisteault @nworr 

@rldhont I haven't tested it, related to our discussion last week, and mainly about environment variables you were talking last week.

### DIFF
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -16,7 +16,7 @@ services:
         aliases:
           - db.lizmap.local
   redis:
-    image: redis:5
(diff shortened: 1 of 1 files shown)


Gustry:
Linked to https://github.com/3liz/lizmap-docker-compose/blob/master/docker-compose.yml#L55

rldhont (approved):

PR #4953, opened 2024-11-07
XSS: allow `iframe` in popups

nboisteault:
Allow iframe in popups but add `sandbox="allow-scripts allow-forms"` to avoid XSS

Funded by Valabre


josemvm:
@nboisteault thanks!

PR #4955, opened 2024-11-07
XSS: sanitize more GFI content

nboisteault:
Funded by 3Liz


Gustry:
Is-it ok to add a test similar to #4956 ? or is-it not relevant ?

nboisteault:
@Gustry I have added a test

PR #4960, opened 2024-11-07
Admin - See if an HTTP error occurred

Gustry:
@rldhont As seen with a use case today

I just tried with a `GetPrintAtlas` request by triggering a wrong layer name.

### DIFF
--- a/lizmap/modules/lizmap/lib/Request/Proxy.php
+++ b/lizmap/modules/lizmap/lib/Request/Proxy.php
@@ -484,6 +484,22 @@ class Proxy
     }
 
     /**
+     * Log if the HTTP code is a 4XX or 5XX error code.
+     *
+     * @param int    $httpCode The HTTP code of the request
+     * @param string $url      The URL of the request, for logging
+     */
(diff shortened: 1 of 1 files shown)


mdouchin (changes requested):
Please replace `str_starts_with` by a test with `substr`
https://www.php.net/manual/en/function.substr.php
to keep the compatibility with PHP 7.4

Gustry:
I think it's better to raise this 500 errors to the user if it's legit ? And maybe it makes an opportunity to inform the user about this issues.

QGIS has a cache on desktop to not flood in case of repeated error in the QgsMessageBar : https://github.com/qgis/QGIS/blob/11946b9c5fb862e62b7c6a7bf7b4aceb75a10c71/src/gui/qgsmessagebar.cpp#L338

rldhont:
@Gustry once merged, we can add the same logic to `Lizmap\Request\OGCRequest::request` to log parameters without `map` parameter replaced by `repository` and `project` in `lizmapadmin`.

PR #4966, opened 2024-11-13
[Bugfix] Feature toolbar: child layer is pivot

rldhont:
Funded by Le Grand Narbonne


rldhont:
Hi @mind84,

I have to update the way `isPivot` is calculated in `FeatureToolbar`. It is not because a table has multiple foreign keys, it is  a pivot table. We have a checkbox in lizmap plugin to defined it.

mind84:
Hi @nboisteault,

sure, I don't see any drawbacks.

Thanks!

PR #4973, opened 2024-11-14
Update Jelix file modified when building LWC

Gustry:
Update Jelix file modified when building LWC

I'm not sure to which branch it must be backported.


nworr:
all "active" branch (using jelix > 1.8.12) : 3.7 and upper

nworr (approved):

Issue #4861, opened 2024-10-09
Migration from Bootstrap 2 to Bootstrap 5 - List changes 


mdouchin:
This ticket aims to list all the changes to make in HTML code (in LWC source code but also in user-defined HTML, for example for popups with QGIS tooltips)

First migration made in this PR : https://github.com/3liz/lizmap-web-client/pull/4455

We need to fix the issues as we spot them, and should also have a look to the external modules, external JavaScript scripts, etc.

## Tables

* condensed table : replace `<table>` class `table-condensed` by `table-sm`

## Buttons

* smaller buttons: replace  `<button>` class `btn-small` by `btn-sm`
* really smaller buttons: class `btn-mini` does not exists anymore ! See https://getbootstrap.com/docs/5.3/components/buttons/#sizes We could have kept `btn-mini` class in the code and add a Lizmap CSS variable `--bs-btn-padding-y: .25rem; --bs-btn-padding-x: .5rem; --bs-btn-font-size: .75rem;"` (values not tested)

## Tabs

Doc: 
* Tabs: https://getbootstrap.com/docs/5.3/components/navs-tabs/#tabs
* Pills: https://getbootstrap.com/docs/5.3/components/navs-tabs/#pills
* PR in Lizmap server https://github.com/3liz/qgis-lizmap-server-plugin/pull/87

## Modal

* Example on the demo repository https://github.com/3liz/lizmap-web-client-demo/issues/14

## Popups and dataviz

* To be explained please https://github.com/3liz/lizmap-plugin/issues/592

## Forms

They seem really broken at present in master

## Well (Admin Dashboad items)

Wells has been replaced by [Cards](https://getbootstrap.com/docs/5.3/components/card/)  since boostrap 4 ([note](https://getbootstrap.com/docs/4.1/migration/#panels-thumbnails-and-wells)) , its only used by [dashboad item](https://github.com/3liz/lizmap-web-client/blob/release_3_10/lizmap/app/themes/default/master_admin/zone_dashboard.tpl) (class `well` ) .
We can use `card-header` for `h3` and `card-body` for the widget content, additionaly , `span6` can be replaced by `col-6`

mdouchin:
cc @nboisteault @rldhont @Gustry @nworr @mind84

ygorigor:
Also progress is not showing when loading dataviz plot

mdouchin:
[...]

PR #4974, opened 2024-11-14
Raise PHP minimum version to 8.1

Gustry:
Following the PHP roadmap : https://www.php.net/supported-versions.php

But the issue is about Debian bulleyes, for which it will not be compatible.

~We talked about targeting LWC 3.9 for this change, but I was wondering if we keep this for LWC 3.10 which already include a "breaking" change with the migration from Bootstrap 2 to Bootstrap 5 #4861. (we can check if LWC should switch to 4.0 ?)~

CC @laurentj @mdouchin @mind84 @guenter

### DIFF
--- a/.github/workflows/php-stan.yml
+++ b/.github/workflows/php-stan.yml
@@ -22,8 +22,6 @@ jobs:
     strategy:
       matrix:
         include:
-          - php-version: '7.4'
-          - php-version: '8.0'
           - php-version: '8.1'
           - php-version: '8.2'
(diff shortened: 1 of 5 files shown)


rldhont:
To enforce PHP version compatibility, I proposed to enhance Github Actions for PHPUnit and PHPStan
* https://github.com/3liz/lizmap-web-client/pull/5169
* https://github.com/3liz/lizmap-web-client/pull/5170

PR #4983, opened 2024-11-15
Js files finder

nworr:
Refacto of code of lizmap controller to a dedicated class : reusable , testable.

js files are in several directory to check multiple case (default, js in project directory , subsfolder)  
Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/lib/App/AppContextInterface.php
+++ b/lizmap/modules/lizmap/lib/App/AppContextInterface.php
@@ -269,8 +269,9 @@ interface AppContextInterface
      * Returns the URL corresponding to the Jelix Selector.
      *
      * @param string $selector The Jelix selector
+     * @param mixed  $params   action params
      */
-    public function getUrl($selector);
+    public function getUrl($selector, $params = array());
 
     /**
      * Returns the absolute Url.
--- a/lizmap/modules/lizmap/lib/App/JelixContext.php
+++ b/lizmap/modules/lizmap/lib/App/JelixContext.php
@@ -329,9 +329,9 @@ class JelixContext implements AppContextInterface
         return \jForms::create($formSel, $formId);
     }
 
-    public function getUrl($selector)
+    public function getUrl($selector, $params = array())
     {
-        return \jUrl::get($selector);
+        return \jUrl::get($selector, $params);
     }
 
     public function getFullUrl($selector, $params = array())
--- a/lizmap/modules/lizmap/lib/Project/ProjectFilesFinder.php
+++ b/lizmap/modules/lizmap/lib/Project/ProjectFilesFinder.php
@@ -1,0 +1,71 @@
+<?php
+
+/**
+ * @author    3Liz
+ * @copyright 2024 3liz.com
+ *
+ * @see      3liz.com
+ *
+ * @license   MIT
+ */
+
+namespace Lizmap\Project;
+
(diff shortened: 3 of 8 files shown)


lizmap/modules/lizmap/lib/Project/UserFilesFinder.php
```diff
@@ -0,0 +1,75 @@
...
+
+                $items = array(
+                    // current repository
+                    'media/js/',
+                    // or root (js shared over repositories)
+                    '../media/js/',
+                );
+                foreach ($items as $item) {
+                    $jsPathRoot = realpath($repositoryPath.$item.$dir);
+                    if (is_dir($jsPathRoot)) {
```
Gustry, line 37:
```suggestion
                    if (!is_dir($jsPathRoot)) {
```

To keep code on the left side of the editor, I find it more convenient negate, and a `continue` statement, so then all the code under can be pulled left.

Just a suggestion, about code visibility.

Gustry:
Nice addition !
We will add it after in the project page table ;-)

lizmap/modules/lizmap/lib/Project/UserFilesFinder.php
nworr: i'm not very keen of using "continue" statement, let's other reviewers decide ?

mdouchin: +1 for using continue.
This provides more readable content and avoids indentation nightmares

lizmap/modules/lizmap/lib/Project/UserFilesFinder.php
```diff
@@ -0,0 +1,70 @@
...
+ * @copyright 2024 3liz.com
+ *
+ * @see      3liz.com
+ *
+ * @license   MIT
+ */
+
+namespace Lizmap\Project;
+
+class UserFilesFinder
```
mdouchin, line 14: I would call it `projectFilesFinder` instead, even if in the future, list of files may depend on the authenticated user.

mdouchin (changes requested):
+1 for this nice modification, minor comments

lizmap/modules/lizmap/lib/Project/UserFilesFinder.php
Gustry: This was not a strong opinion, just a way of coding that I prefer better.

Imagine a very long `if` block ... inside a loop. Having a `continue` ensures me that there is nothing under this very long `if` block. It puts some kind of safeguards that nothing will be executed under this statement

Just my 2 cents ;-)

Gustry (approved):
Nice ;-)

nworr:
backport 3.9 3.8 3.7 ? 3.6 ??

Gustry:
Try them all :)
But I'm not sure the CI will be valid for 3.6. We can make a non-official release.

Gustry:
Good to go @nworr ?

rldhont (approved):

rldhont:
@mdouchin we need your feedback to merge

Gustry:
> @mdouchin we need your feedback to merge

@nworr did what was requested, just pressing the "resolved" button is enough ;-)

I just did it. Ok for merging ?

nworr:
Thinking this PR, I found out a method to find files for whole repository would be great.
So back to "draft"

PR #5005, opened 2024-11-25
Upgrade PHP-CS-Fixer from 3.8.0 to 3.40.0

Gustry:
It will be manually backported to all active branches if necessary (to avoid backport conflicts)


rldhont (approved):

nworr (approved):
Many files updated, but i think it will not introduce much conflicts with pending PRs.

Thanks

PR #5006, opened 2024-11-25
Add missing modules action and filter in PHP-CS-Fixer config

Gustry:
Will be manually backported to active branch


rldhont (approved):
Thanks @Gustry !

PR #5010, opened 2024-11-26
Import FlatGeobuf

neo-garaix:
### Now flatgeobuf files can be imported to the digitizing tool.

[Capture vidéo du 2024-11-26 12-49-07.webm](https://github.com/user-attachments/assets/69a1311d-996e-426f-a2ac-60ef4bddd262)

Funded by [Golfe du Morbihan Vannes agglomération](https://www.golfedumorbihan-vannesagglomeration.bzh/)


assets/src/modules/Digitizing.js
```diff
@@ -1195,6 +1224,38 @@ export class Digitizing {
...
+
+                            // We need a reprojection to be done because flatgeobuf files doessn't have a precise projection
+                            // We neither used wkt located in headers due to some errors to define it with proj4.
+                            // Nor 'fromEPSGcode()' because it returns a 'Projection' object that sometimes throws errors.
+                            let descriptor = await fetch('https://epsg.io/' + projCode + '.proj4');
+                            descriptor =  await descriptor.text();
+
+                            proj4.defs(projFGB, descriptor);
+                            register(proj4);
+
```
rldhont, line 1248: Checking if the projCode is allready known before requesting epsg.io

rldhont (changes requested):

assets/src/modules/Digitizing.js
Gustry: Should we check as well if the `descriptor` ended with a HTTP code < 400 ? (or similar)
https://fr.wikipedia.org/wiki/Liste_des_codes_HTTP

neo-garaix: > Should we check as well if the `descriptor` ended with a HTTP code < 400 ? (or similar) https://fr.wikipedia.org/wiki/Liste_des_codes_HTTP

I will replace my utilization of "fetch" by the one located directly in the [Utils module](https://github.com/3liz/lizmap-web-client/blob/master/assets/src/modules/Utils.js#L113) to take care of them.

rldhont (approved):

PR #5013, opened 2024-11-28
FeaturesTable structure & columns improvements 

neo-garaix:
### Added

* Replace the items `<div>` by an HTML `<table>` to prepare the addition of columns
* Allow to add columns from the layer fields : now we can add several columns by adding some `<lizmap-field ... > ... </lizmap-field>` children inside the `<lizmap-features-table>` element
* Add labels of columns in the table
* Add a regular expression verification for the used QGIS expressions in child fields. At present, we only accept expressions requesting fields like `"libsquart"` and not expressions like `rand(10,34)`.
* If fields have the same expression or aliases (except empty ones), only the first one is used.

Example code:

```html
<lizmap-features-table 
	uniquefield="id" 
	expressionfilter="quartmno = '[% "quartmno" %]'" 
	withgeometry="1" 
	sortingfield="libsquart" 
	sortingorder="asc" 
	draggable="yes"
>
	<lizmap-field 
		data-alias="Name" 
		data-description="Label of sub-district's"
	>"libsquart"</lizmap-field>
	
	<lizmap-field 
		data-alias="Code" 
		data-description="Code of sub-district's name"
	>"squartmno"</lizmap-field>
</lizmap-features-table>
```
How it is rendered (in release_3_9, not in master since bootstrap migration is not finished yet)

image

### Fixed

* Fixed the **end2end playwright test** about FeaturesTable

_Fix version of #4922_ 

Funded by 3liz

### DIFF
--- a/assets/src/components/FeaturesTable.js
+++ b/assets/src/components/FeaturesTable.js
@@ -17,6 +17,15 @@ import { mainLizmap, mainEventDispatcher } from '../modules/Globals.js';
  * @element lizmap-features-table
  * @fires features.table.item.dragged
  * @fires features.table.rendered
+ * @example <caption>Example of use</caption>
+ * <lizmap-features-table draggable="yes"sortingorder="asc" sortingfield="libsquart"
+ *                        withgeometry="1" expressionfilter="quartmno = 'HO'"
+ *                        uniquefield="id" layerid="subdistrict_24ceec66_e7fe_46a2_b57a_af5c50389649"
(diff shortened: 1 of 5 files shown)


rldhont (approved):

nboisteault (approved):

mdouchin (approved):
I modified the component & CSS to fix broken styles (blue background on hover, margins, fonts, etc.). It has been tested for `release_3_9`. Style is broken in master, as for other parts of LWC

PR #5036, opened 2024-11-29
Reduce mainLizmap dependencies in Action module

neo-garaix:
_Based on #3857_ 

Funded by 3liz

### DIFF
--- a/assets/src/modules/Action.js
+++ b/assets/src/modules/Action.js
@@ -6,7 +6,7 @@
  * @license MPL-2.0
  */
 
-import { mainLizmap } from '../modules/Globals.js';
+import SelectionTool from './SelectionTool.js';
 import { Vector as VectorSource } from 'ol/source.js';
 import { Vector as VectorLayer } from 'ol/layer.js';
 import GeoJSON from 'ol/format/GeoJSON.js';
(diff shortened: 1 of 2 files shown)


assets/src/modules/Action.js
```diff
@@ -56,8 +56,17 @@ export default class Action {
 
     /**
      * Build the lizmap Action instance
+     * @param {Map}           map           - OpenLayers map
+     * @param {SelectionTool} selectionTool - The lizmap selection tool
+     * @param {object}        lizmap3       - The old lizmap object
+     * @param {Digitizing}    digitizing    - The Lizmap digitizing instance
      */
-    constructor() {
+    constructor(map, selectionTool, lizmap3, digitizing) {
```
rldhont, line 64: It is hoped that the use of lizmap3 will disappear and be replaced by the new API. Can you set lizmap3 as the last parameter?

assets/src/modules/Lizmap.js
```diff
@@ -161,7 +161,7 @@ export default class Lizmap {
                 this.proxyEvents = new ProxyEvents();
                 this.wfs = new WFS();
                 this.wms = new WMS();
-                this.action = new Action();
+                this.action = new Action(this.map, this.selectionTool, this.lizmap3, this.digitizing);
```
rldhont, line 164:
```suggestion
                this.action = new Action(this.map, this.selectionTool, this.digitizing, this.lizmap3);
```

Same remarck

rldhont (changes requested):

PR #5045, opened 2024-12-02
Form filter: filter autocomplete list based on previous applied filters

nboisteault:
This PR changes the behaviour of the form filter. Previously, all unique values from the text field (with autocompletion activated) were downloaded once at startup. 

Now, every time the user types letters (ex: `road`), a new request is triggered with the current filter (depending on the other fields) combined with the searched term for this field (ex: `"field" LIKE '%road%' `).

NB: It is recommended to create a trigram index on the searched fields to improve performance, such as :

```sql
CREATE EXTENSION IF NOT EXISTS pg_trgm;
CREATE INDEX my_table_a_field_trgm_index ON "my_schema"."my_table" USING GIN ("a_field" gin_trgm_ops);
```

Funded by [VSB](https://www.vsb.energy/de/de/startseite/)

https://github.com/user-attachments/assets/6278ecb5-d21a-4024-8328-ce7e7cbc864e


Gustry:
Nice addition.

Related to form_filter, is there any chance we can remove the manual test about the "form filter" at the same time, as you moved from Cypress to Playwright ? https://github.com/3liz/lizmap-web-client/blob/master/tests/qgis-projects/tests/form_filter.md (maybe with the help of echoproxy ?)

nboisteault:
> Nice addition.
> 
> Related to form_filter, is there any chance we can remove the manual test about the "form filter" at the same time, as you moved from Cypress to Playwright ? https://github.com/3liz/lizmap-web-client/blob/master/tests/qgis-projects/tests/form_filter.md (maybe with the help of echoproxy ?)

Done bae34404e31314cd1ae6bd3f7a118921710359e7

tests/end2end/playwright/form-filter.spec.js
```diff
@@ -0,0 +1,106 @@
+// @ts-check
+import { test, expect } from '@playwright/test';
+import { gotoMap } from './globals';
+
+test.describe('Attribute table', () => {
```
Gustry, line 5:
```suggestion
test.describe('Form filter', () => {
```

Gustry (commented):
Thanks for the removal of the manual test ... :)

mdouchin:
This PR changes the behaviour of the form filter. Previously, all unique values from the text field (with the autocompletion activated) were downloaded once at startup. Now, every time the user types letters, a new request is triggered with the given filter and the `LIKE` and fetches all corresponding data.

It is more dynamic and better than before, but I think we should test it against a big dataset to evaluate the performance. `LIKE` clauses are known to have poor performances. We could promote the use of the `pg_trgm` PostgreSQL extension (and add index as described here https://docs.lizmap.com/next/en/publish/configuration/spatial_search.html#optimisation)

nboisteault:
> This PR changes the behaviour of the form filter. Previously, all unique values from the text field (with the autocompletion activated) were downloaded once at startup. Now, every time the user types letters, a new request is triggered with the given filter and the `LIKE` and fetches all corresponding data.
> 
> It is more dynamic and better than before, but I think we should test it against a big dataset to evaluate the performance. `LIKE` clauses are known to have poor performances. We could promote the use of the `pg_trgm` PostgreSQL extension (and add index as described here https://docs.lizmap.com/next/en/publish/configuration/spatial_search.html#optimisation)

I've tested with 1 100 000 lines inserted with:

```sql
INSERT INTO tests_projects.form_filter (label, geom)
SELECT
 md5(random()::text),
 null
FROM generate_series(1, 1000000);
```

Request made to get values when typing 'abc' returns in 800ms. 'abcd' or 'abcde' in 300ms.

mdouchin:
> Request made to get values when typing 'abc' returns in 800ms. 'abcd' or 'abcde' in 300ms.

great. Have you tried by using the following index

```sql
CREATE EXTENSION IF NOT EXISTS pg_trgm;
DROP INDEX IF EXISTS test_index;
CREATE INDEX test_index ON tests_projects.form_filter USING GIN (label gin_trgm_ops);
```

nboisteault:
> > Request made to get values when typing 'abc' returns in 800ms. 'abcd' or 'abcde' in 300ms.
> 
> great. Have you tried by using the following index
> 
> ```sql
> CREATE EXTENSION IF NOT EXISTS pg_trgm;
> DROP INDEX IF EXISTS test_index;
> CREATE INDEX test_index ON tests_projects.form_filter USING GIN (label gin_trgm_ops);
> ```

Now request made to get values when typing 'abc' returns in 500ms. 'abcd' or 'abcde' in 100ms.

assets/src/legacy/filter.js
```diff
@@ -392,48 +392,57 @@ var lizLayerFilterTool = function () {
...
+                $("#filter-field-order" + String(field_item.order)).append(html);
+                addFieldEvents(field_item);
+
+                $("#liz-filter-field-text" + lizMap.cleanName(field_item.title)).autocomplete({
+                    source: function (request, response) {
+                        const autocompleteFilter = `"${field}" ILIKE '%${request.term}%' `;
+                        if (!globalThis['filterConfigData'].filter) {
+                            sdata.filter = autocompleteFilter;
+                        } else {
+                            sdata.filter = globalThis['filterConfigData'].filter + ' AND ' + autocompleteFilter;
```
mdouchin, line 419: The generated filter can contain OR terms. We should use parenthesis to ensure the correct logic will be used. Spaces are used around them to please QGIS :
```suggestion
                            sdata.filter = ' ( ' + globalThis['filterConfigData'].filter + ' ) AND ' + autocompleteFilter;
```

mdouchin (changes requested):

PR #5058, opened 2024-12-02
Tests - Use URLSearchParams constructor to build querystrings

Gustry:
@rldhont A follow up from the PR #5053

* Lines are way shorter, no need to have horizontal scrolling
* Parameters are automatically escaped, like `EXP_Filter: '"gid" = 3'`
* Better with Git, easier to see which parameter is updated later


rldhont (approved):
Good catch!

PR #5059, opened 2024-12-02
Tests - Remove warnings from Javascript

Gustry:
In my Javascript IDE, I have plenty of warnings like : 

image


rldhont:
@Gustry `lizmap-feature-toolbar` is a custom element unknown by IDE. The way to fix is to extend the list of available HTML element in Lizmap Web Client project.

Gustry:
Yes, I noticed, there are ways to overpass this, but not for now.

PR #5079, opened 2024-12-04
[Feature] Circular geometry measurement on draw

mind84:
**ADDED FEATURES:**
 
- Add measure tooltip on circles with radius and area measurement
- Initialize measure tooltips on features loaded from `local storage`

**NOTES:**

- Circles area is calculated using the `fromCirlce` method of Openlayers Polygon [class](https://openlayers.org/en/latest/apidoc/module-ol_geom_Polygon.html).
There is another (more verbose) way to do it that involves `circular` method but requires to convert center and "last" coordinates in EPSG:4326.
Given the same polygonal approximation (128 sides), the two methods return the same output so I decided to use the less verbose one.

- For circles, the "segment" tooltip (radius) is no longer displayed near mouse pointer but it's displayed in the interior point of the feature, along with the area measure.  

Feature preview:

[circle_measurement.webm](https://github.com/user-attachments/assets/ea505e4d-87d9-4e5b-b999-a5e872fade23)

Funded by Etra

### DIFF
--- a/assets/src/modules/Digitizing.js
+++ b/assets/src/modules/Digitizing.js
@@ -480,13 +480,7 @@ export class Digitizing {
                     geom.set('totalOverlay', Array.from(this._measureTooltips).pop()[1], true);
                     geom.on('change', (e) => {
                         const geom = e.target;
-                        if (geom instanceof Polygon) {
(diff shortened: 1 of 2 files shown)


rldhont:
Thanks @mind84, can you provide a print screen or a gif ?

PR #5080, opened 2024-12-04
Tests on the landingpage with XPath and opacity

Gustry:
I can confirm that reverting PR #5011 solves the 500.

The purpose was not to merge this PR, but to understand why the E2E was OK on PR #5011.
While there is a Cypress test about the landing page :
https://github.com/3liz/lizmap-web-client/blob/release_3_8/tests/end2end/cypress/integration/projects_homepage-ghaction.js

I can confirm there is a cache issue.

I tried to empty the redis database : 

```
➜ docker exec -it lizmaprelease_3_8_test_redis redis-cli flushall
OK
```

PR #5011, opened 2024-11-27
include layer opacity info for embedded layers

mind84:
Backport for #4964

Funded by Faunalia

PR #5083, opened 2024-12-05
[Bugfix] Fix null xml QgisProject property during projects loading

mind84:
Refers to #5011 and #5080

Funded by Faunalia


rldhont (approved):

Gustry (approved):
LGTM from the test side. It works now.
Thanks for debug @mind84

PR #5105, opened 2024-12-10
Display a generic error in case of action failure to the end-user

Gustry:
When **not** connected on the instance @mdouchin 

Does the action writer is using `raise` to show error to the end user ?

image


mdouchin (approved):
Thanks

rldhont (approved):

Issue #4898, opened 2024-10-22
How to create a drill down form with Lizmap 


arno974:
### What is the question? (in English)

Hello, 

For a lizmap project I need to report damages made to bus stop furniture (example : the bin has been destroyed). To create a new damage report, users have to click on a bus station and then select one of the furniture linked to this bus station. 

To only link the furniture related to this bus station, I must create a linked drop-down lists.

So I've three tables : 

- bus_station : geolocalized postgis table point
- furniture : PostgreSQL type with an FK link to bus_station
- damages : PostgreSQL type with an FK link to furniture and bus_station (a damage is linked to a a furniture and to a bus station).

I've configured the relationships in QGIS and set up the QGIS form id_arrets_bus and id_mobiliers fields. 

**QGIS form id_arrets_bus id_mobiliers parameter**
![image](https://github.com/user-attachments/assets/4af740ab-dc18-465b-80cb-a62f5ad3da94)

**QGIS form id_mobiliers field parameter**
![image](https://github.com/user-attachments/assets/2d4b3536-47c4-469b-ac9c-00255aace5f6)

In QGIS, the furniture is correctly filtered according to the selected stop.
![image](https://github.com/user-attachments/assets/233ca7c5-ec8f-4d69-bd58-5e7564f03e12)

But in Lizmap, the drop down list remains empty
![image](https://github.com/user-attachments/assets/5194a3e3-fd0d-4749-9fed-3b12df0edc83)

I have read that Lizmap is able to use QGIS Expression and filters but it doesn't seems to work in my case.

Have you ever had to create nested drop-down lists? If so, could you please explain me how to do it ?

Thanks,

Arnaud

### Versions, safeguards, check summary etc

**Versions :**

* Lizmap Web Client : 3.7.11
* Lizmap plugin : 4.4.2
* QGIS Desktop : 3.34.4
* Lizmap.com : Yes
* QGIS Server : 3.34.11
* Py-QGIS-Server : 1.9.1
* QGIS Server plugin DataPlotly : Not installed
* QGIS Server plugin atlasprint : 3.4.1
* QGIS Server plugin cadastre : 1.20.0
* QGIS Server plugin lizmap_server : 2.11.2
* QGIS Server plugin wfsOutputExtension : 1.8.2

<details>
<summary>List of Lizmap Web Client modules :</summary>
<br/>
* Version Lizmap Web Client 3.8 needed
</details>

<details>
<summary>List of safeguards :</summary>
<br/>
[...]

PR #5107, opened 2024-12-11
Fix edition form child with relation

nworr:
Fix edition form when lauchind children creation from attriute table button : 
if there's a relation between parent id (in a select) and an other control, the value weren't updated.

the `select` is now not set as `disabled` but hidden and a disabled input is added ( look alike a select)
edition_child_parent_relation

Also mini fix on attribute table : the children table didn't work in some case.

Ticket : #4898

Funded by 3Liz


mdouchin (approved):

Gustry:
CC @mind84 related to your work on this side

PR #5121, opened 2024-12-12
Editing - Show the previous form when the user cancel current editing

mdouchin:
Sometimes, the user can forget that the editing form has been previously opened to create or edit a feature.
Some other tools can have been used, such as the Layers legend panel, which hides the form.

This PR displays the editing form back before showing the confirmation message, so that the user can see the previous form in its current state and make the correct decision.

Funded by AMP Métropole

### DIFF
--- a/assets/src/legacy/edition.js
+++ b/assets/src/legacy/edition.js
@@ -713,6 +713,16 @@ var lizEdition = function() {
     }
 
     /**
+     * Sleep function
+     * @param {number} sleepDuration Duration to wait (milliseconds)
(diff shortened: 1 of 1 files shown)


mdouchin:
Not sure about the addition of async...

mdouchin:
cc @Gustry

rldhont (approved):

rldhont:
@nboisteault can you give us your point of view ?

mdouchin:
close as a duplicate of #5145

PR #5125, opened 2024-12-12
Dataviz: use a plotlyjs custom bundle to reduce file size

nboisteault:
Update plotlyjs to v2.35.2

Funded by 3Liz


Gustry (commented):
Which version of Plotly was it before ?

PR #5135, opened 2024-12-17
Fix: order popups following layers order

nboisteault:
Ticket : 3249 (paid support)

Funded by 3Liz, Biotope


nboisteault:
Manual backports for release_3_8 and release_3_7

PR #5137, opened 2024-12-20
PHP: reading GetFeature GeoJSON as a stream with JSON-Machine

rldhont:
JSON-Machine is a very easy to use and memory efficient drop-in replacement for inefficient iteration of big JSON files or streams for PHP https://github.com/halaxa/json-machine

In some part of Lizmap PHP code, GetFeature GeoJSON are read in an inefficient way.


mdouchin:
I just tested this PR with the Land-use project data (the one visible here https://demo.lizmap.com/lizmap/index.php/view/map?repository=features&project=land_use)
It works flawlessly.

PR #5139, opened 2024-12-30
Draw: add split tool

nboisteault:
Funded by 3Liz


nboisteault:
Manually backported to 3.9

Issue #5146, opened 2025-01-02
FeaturesTable additions


neo-garaix:
### Is your feature request related to a problem? Please describe

→ In order to add a module based on FeaturesTable, we will need to add and modify some behavior 

### Describe the solution you'd like

- [x] Show the geometry of the item selected by highlighting it
- [x] Center the view on the selected item
--
- [x] Add an event when an item is highlighted
~- [ ] Add an event for both next/before buttons that sends feature IDs~
--
- [x] Use the attributeChangedCallback to detect when the expressionFilter attribute is modified, then we reload FeaturesTable
--
- [x] Remove the popup opened by the click and detect it through an attribute (popup-displayed) #5194 

(cc: @mdouchin )

Gustry:
I think this can be closed, thanks @neo-garaix

PR #5153, opened 2025-01-03
Features Table : add an event when an item is highlighted

neo-garaix:
Improvements for the `<lizmap-features-table>` web component:

* Now, when an item is highlighted, an event `features.table.item.highlighted` is sent with the item's `feature_id`.

Funded by Andromède Océanologie https://www.andromede-ocean.com/

Ticket : #5146


mdouchin (approved):
Thanks

rldhont (approved):

PR #5166, opened 2025-01-08
Activate given line in featuresTable

nboisteault:


mdouchin (changes requested):
This is the simplest way to go indeed.

We could invert the logic, and if possible:


* add a getter and a setter to the property `activeItemFeatureId`. The setter will be responsible for fetching the feature popup HTML content and display it. If `activeItemFeatureId` is null, the popup will be hidden.
```javascript
  get activeItemFeatureId() {
  }
  set activeItemFeatureId(value) {
  }
```
* move some logic from `onItemClick` to a new `togglePopup` which will decide if a popup should be displayed (feature ID set) or hidden (feature ID is null)
* create a new method `displayPopup` which will be called by `togglePopup` 
* When listening to the click, only set the property `activeItemFeatureId` and let the setter do the job by using `togglePopup`
* On render, after template is rendered, call `togglePopup` which will display a popup if needed

Just thinking out loud here, not tested it

PR #5168, opened 2025-01-08
Use QJazz in tests

dmarteau:
QJazz QGIS server in tests.


tests/Makefile
```diff
@@ -44,32 +46,24 @@ env:
...
+	echo "Plugins are auto installed at qgis worker startup"
 
 show-qgis-server-versions:
 	docker run \
 	    --rm -i \
-	    --entrypoint qgisserver \
-	    3liz/qgis-map-server:${LZMQGSRVVERSION} \
-	    --version
+	    --entrypoint qgis-server-rpc \
+	    3liz/qgis-services:${LZMQGSRVVERSION} \
```
Gustry, line 55:
```suggestion
	    3liz/qjazz:${LZMQGSRVVERSION} \
```

QJazz or services ?

Same line L66

tests/docker-conf/qgis-server.toml
```diff
@@ -0,0 +1,39 @@
+#
+# Worker backend configuration
+#
+
+[worker]
+name = "Qgis"
```
Gustry, line 6:
```suggestion
name = "QGIS"
```

Gustry (approved):
I tried it locally, it works.
Still an issue on CI for now.

Will tags such as `3.34`, `3.40` will be available on Docker hubs later ?

tests/Makefile
dmarteau: qjazz

Gustry: So the repo https://hub.docker.com/r/3liz/qgis-services will be removed ?

dmarteau: yes

Gustry (changes requested):
* Needed QGIS Server plugins are not installed in the current state of this PR. That's why the CI is not working. It was working locally, because my plugins were already existing before.
* Can you remove the script `tests/add_server_plugins.sh` ?
* In `tests/add_server_plugins.sh`, we need to use to the custom repository for QGIS plugins. Please do the change
* Will tags such as `3.34`, `3.40` will be available on Docker hubs ? https://hub.docker.com/r/3liz/qjazz/tags

Gustry:
For testing purpose, I have pressed the rebased button, to have latest commits from master.

dmarteau:
> Is-it not missing the mount for QGIS plugins in `map` container ? But it doesn't seem to be enough to have QGIS server plugins installed, even when rebuilding/starting containers.

Rules for plugins are the following:

1. Plugins location should be mounted somewhere
2. Wether or not plugin are installed at startup depends on configuration settings ('auto' vs 'manual').
3. Allowed plugins must be explicitely declared in configuration settings (see `tests/docker-conf/qgis-server.toml`)

dmarteau:
There is tests checking `OnlineResource`: they seems to be wrong to me.

The qjazz proxy handle standard `Forwarded` headers as defined in [RFC 7239](https://datatracker.ietf.org/doc/html/rfc7239) and legacy `X-Forwarded-Host` and `X-Forwarded-Proto`. 

The path returned in the service response may be changed with the appropriate `X-Qgis-Service-Url` header  (See the documentation for the [`QGIS_SERVER_SERVICE_URL` environment variable](https://docs.qgis.org/3.34/en/docs/server_manual/config.html#environment-variables).

dmarteau:
Found that QGIS does not stick to standard and handle headers in case sensitive way.

dmarteau:
I cannot check what's wrong on failed test, i need at least the request that failed if this is is a request to the QGIS service.

dmarteau:
@Gustry : please check the timeout error on Playwright test, with proper rights there is no issue locally with the  url I can guess from the  details.

Gustry:
Something is very slow in this PR.
At the end of the CI, there will be some artifacts generated : https://github.com/3liz/lizmap-web-client/actions/runs/12785872522?pr=5168, especially all QGIS servers logs, from the container `lizmapBRANCH_test_qgis`.
Should we also save logs from `lizmapBRANCH_test_qgis_rpc` container ?

tests/docker-compose.yml
```diff
@@ -67,34 +67,41 @@ services:
...
     depends_on:
       - lizmap
+  # Qgis RCP backend
+  map-rpc:
+    container_name: "lizmap${LZMBRANCH}_test_qgis_rpc"
+    image: 3liz/qjazz:${LZMQGSRVVERSION}
+    command: ["qjazz-rpc", "serve", "-C", "/srv/etc/qgis-server.toml"]
+    environment:
+      CONF_LOGGING__LEVEL: debug
+      CONF_WORKER__NUM_PROCESSES: 4
```
Gustry, line 77:
```suggestion
      CONF_LOGGING_LEVEL: debug
      CONF_WORKER_NUM_PROCESSES: 4
```
Is-it expected ?

Gustry (commented):
Is-it a typo ?

tests/docker-compose.yml
dmarteau: Do not change this, this is a standard mapping from env variables for structured configuration schema, double underscore '__' is section separator. This is automaticalled parsed and used by framework like Pydantic in Python or Rust configuration parser.

dmarteau:
> Something is very slow in this PR. At the end of the CI, there will be some artifacts generated : https://github.com/3liz/lizmap-web-client/actions/runs/12785872522?pr=5168, especially all QGIS servers logs, from the container `lizmapBRANCH_test_qgis`. Should we also save logs from `lizmapBRANCH_test_qgis_rpc` container ?

Yes, this where we log QGIS opérations.

dmarteau:
Where can I find logs related to failed tests ?

Gustry:
Can you rebase on top of master, I have merged #5203 

All logs are in the artifacts, for example, for the last run, after 194 minutes :(, artifacts are on the page summary  https://github.com/3liz/lizmap-web-client/actions/runs/12785872522?pr=5168 at the bottom, direct link to ZIP : 

https://github.com/3liz/lizmap-web-client/actions/runs/12785872522/artifacts/2434513285

The timestamp in the QGIS Server logs are very slow.

> Yes, this where we log QGIS opérations.

So can you add in your PR `.github/workflows/e2e_tests.yml` L220: 

```
docker logs lizmap${{ env.LZMBRANCH }}_test_qgis_rpc &> /tmp/e2e/docker/qgis-server-rpc.log | true
```

dmarteau:
> So can you add in your PR `.github/workflows/e2e_tests.yml` L220:
> 
> ```
> docker logs lizmap${{ env.LZMBRANCH }}_test_qgis_rpc &> /tmp/e2e/docker/qgis-server-rpc.log | true
> ```

I don't understand where I should exacty set this line: shouldn't it  be in a job ?  It seems wrong to me jush to add the line add the end of the file.

tests/add_server_plugins.sh
```diff
@@ -1,33 +0,0 @@
...
-  rm -rf /srv/plugins/.cache_qgis_plugin_manager
-fi
-if [ -f /srv/plugins/sources.list ]; then
-  rm -f /srv/plugins/sources.list
-fi
-
-echo "QGIS Server Lizmap and WfsOutputExtension plugins"
-echo "Unstable from https://packages.3liz.org"
-# qgis-plugin-manager init
-echo "https://packages.3liz.org/pub/server-plugins-repository/unstable/plugins.[VERSION].xml" > /tmp/sources-plugin-manager.list
```
Gustry, line 16: You must use this repository for plugins, we use dev version, and not stable version of `Lizmap server`.

tests/docker-compose.yml
```diff
@@ -67,29 +67,40 @@ services:
       - ${LZMWEBPORT}:80
     depends_on:
       - lizmap
+  # Qgis RCP backend
```
Gustry, line 70:
```suggestion
  # QGIS RPC backend
```

Gustry (approved):
Thanks, tests are finally passing.
Last thing to change is the plugin repository to change.

> Will tags such as 3.34, 3.40 will be available on Docker hubs ? https://hub.docker.com/r/3liz/qjazz/tags

So you mean when a git tag will be made on `qjazz` repository, it will be available on docker hub ?

tests/add_server_plugins.sh
dmarteau: Fixed using dedicated sources file in `tests/docker-conf/unstable-plugin-sources.list` and settings the `QGIS_PLUGIN_MANAGER_SOURCES_FILE` env variable for the `map-rpc` service.

Gustry:
@dmarteau Discussion we got with @rldhont and @mdouchin :

Do you think it's possible to add your work in another docker-compose file ? with inheritance ?
We would need to keep the `tests/add_server_plugins.sh`, make a copy of `tests/docker-conf/phpfpm/lizmapConfig.ini.php` and would be nearly able to run both stacks Py-QGIS-Server and QJazz ?

dmarteau:
> Do you think it's possible to add your work in another docker-compose file ? with inheritance ? We would need to keep the `tests/add_server_plugins.sh`, make a copy of `tests/docker-conf/phpfpm/lizmapConfig.ini.php` and would be nearly able to run both stacks Py-QGIS-Server and QJazz ?

I don't see the benefit unless making things  uneccessary complicated. Just keep that 
branch to draft and sync with master until we officially move to QJazz as stable backend since the backend for the LWC tests has only to be an acceptable OWS/QGIS server proxy.

dmarteau:
Waiting for https://github.com/3liz/qgis-lizmap-server-plugin/pull/104 to land on unstable repository before updating.

Gustry:
> Waiting for https://github.com/3liz/qgis-lizmap-server-plugin/pull/104 to land on unstable repository before updating.

You just need to push on gitlab ;-)

dmarteau:
@Gustry  end2end tests fails with the  following error: 
````
An error occurred trying to start process '/usr/bin/bash' with working directory '/home/runner/work/lizmap-web-client/lizmap-web-client/end2end'. No such file or directory
```

Doesn't seems related to Qjazz, but seems to be a misconfiguration issue (wrong path) ?

dmarteau:
@Gustry   ZMQGSRVVERSION is wrong in test:  3.34,  should `be qgis-ltr-eager|qgis-release-eager`

dmarteau:
> @Gustry LZMQGSRVVERSION is wrong in test: 3.34, should `be qgis-ltr-eager|qgis-release-eager`

Fixed qgis-server value in matrix configuration.

dmarteau:
@Gustry tests failed for reasons not related to *Qjazz*

Gustry:
Yes, we know the status of the current CI on the master branch, there are some ongoing work.
But `playwright/editing-value-relation.spec.js`, it's the first time I see it failing. (even if it's a new one since yesterday).

dmarteau:
After rebasing from master, playwright tests are continuously failing.

dmarteau:
@Gustry BLEEDING EDGE tests are completely failing because the plugin source url is invalid:

```
2025-05-07T13:52:42.475Z»   [32]»   INFO»   Installing plugins in /srv/qgis-server-plugins
Downloading https://packages.3liz.org/pub/server-plugins-repository/unstable/plugins.3.42.xml…
    HTTP Error 404: Not Found
2025-05-07T13:52:44.278Z»   [32]»   ERROR»  Cannot update plugins index, cancelling installation...
...
```

And thus the lizmap server plugin is not installed....

AMHA tests should not run at all if the Lizmap server plugin is not installed.

PR #5169, opened 2025-01-10
Github Action PHPStan Matrix

rldhont:
Defined Github Action PHPStan Matrix with PHP version:
* 7.4
* 8.0
* 8.1
* 8.2
* 8.3

Instead of only 8.2


Gustry (approved):
I'm not full sure, but let's try and see.

Issue #5106, opened 2024-12-10
[Bug]: Form filter, Locate by layer, Attribute table – status 500 error – Lizmap 3.8.3


aeduard:
### What is the bug? (in English)

I think it is a problem on my side, not a real bug, but I can't figure out the cause of the error. If the issue does not fit here at github I will repost on discourse.
When trying to use Form filter, Locate by layer, Attribute table on a PostGIS layer (either by connecting with service file or a basic connection) I get the following error on browser console: _Failed to load resource: the server responded with a status of 500 (Internal jelix error)_. Layers are working as expected beside these tools, eg. pop-up.
If I copy the data to a geopackage, there are no errors and everything works as expected.

### Steps to reproduce the issue

Everything is set as described in [documentation](https://docs.lizmap.com/3.8/en/publish/lizmap_plugin/form_filtering.html), namely WFS published in qgis server settings and layer id is also published in layer properties.
If I'm right these tools use direct connection between lizmap and postgresql, maybe there is the source of my problem, while the connection through qgis server works without issues.

 The issue does not produce any lizmap error log.

### Versions, safeguards, check summary etc

**Versions :**

* Lizmap Web Client : 3.8.3 - commit 8da71cbec https://github.com/3liz/lizmap-web-client/commit/8da71cbec
* Lizmap plugin : 4.4.6
* QGIS Desktop : 3.34.10
* QGIS Server : 3.34.12
* Py-QGIS-Server : not used
* QGIS Server plugin lizmap_server : 2.11.2

<details>
<summary>List of Lizmap Web Client modules :</summary>
<br/>
* Version Lizmap Web Client 3.8 needed
</details>

<details>
<summary>List of safeguards :</summary>
<br/>
* Mode : normal<br/>
* Allow parent folder : no<br/>
* Prevent other drive : yes<br/>
* Prevent PG service : no<br/>
* Prevent PG Auth DB : no<br/>
* Force PG user&pass : yes<br/>
* Prevent ECW : yes<br/>
</details>

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 24.04

### Browsers

Firefox, Microsoft Edge

### Browsers version

Firefox 133.0.3, Edge Version 131.0.2903.86

### Relevant log output

```shell
[...]

PR #5182, opened 2025-01-13
Add error log panel in admin

Gustry:
Logs are important, this panel should be more advertised and used. Some upgrades should be done, in this PR and later.

Some review on the log panel :
* Increase from 50 to 200 lines, can we increase more ? Or we should a way to scroll more up if necessary, later.
* Rename "Log" to "Admin log", following #3831
* Adding a text area for the "error log", https://github.com/3liz/lizmap-web-client/issues/5106#issuecomment-2536554681. I didn't add the "Erase button" for this log.
* The new text area is hidden or not according to the flag "sensitive data flag".
image

### DIFF
--- a/lizmap/modules/admin/controllers/logs.classic.php
+++ b/lizmap/modules/admin/controllers/logs.classic.php
@@ -45,21 +45,70 @@ class logsCtrl extends jController
         $conditions = jDao::createConditions();
         $detailNumber = $dao->countBy($conditions);
 
-        // Get last error log
-        $errorLog = \Lizmap\App\FileTools::tail(jApp::logPath('lizmap-admin.log'), 50);
+        // Number of lines for logs
+        $maxLinesToFetch = 200;
+
+        // Get last admin log
+        $lizmapLogPath = jApp::logPath('lizmap-admin.log');
+        $lizmapLog = \Lizmap\App\FileTools::tail($lizmapLogPath, $maxLinesToFetch);
+        $lizmapLogTextArea = $this->logLinesDisplayTextArea($lizmapLog);
+
+        $errorLogDisplay = !\lizmap::getServices()->hideSensitiveProperties();
+        $errorLogPath = jApp::logPath('errors.log');
+        $errorLog = '';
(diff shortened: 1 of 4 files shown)


nworr (commented):
Nice addition, the `tail [..] if ` part is redundant , it can be moved to a zone ?

Gustry:
You mean a PHP function to determine the number of line AND a zone for the `*.tpl` ?

Gustry:
@nworr I have added a PHP function for the height of the text area.

I must admit a bit a lazy to create a new class extending jZone, just to generate these 2 "small" textarea, is-it what I'm supposed to do ?

Just some note for myself for testing : 

```php
jLog::log('A log in lizmapadmin', 'lizmapadmin');
jLog::log('A log in errors', 'error');
jLog::log('A log in warning', 'warning');
jLog::log('A log in notice', 'notice');
jLog::log('A log in default');
```

lizmap/modules/admin/controllers/logs.classic.php
```diff
@@ -45,15 +45,36 @@ public function index()
...
-        $errorLog = \Lizmap\App\FileTools::tail(jApp::logPath('lizmap-admin.log'), 50);
+        // Number of lines for logs
+        $maxLinesToFetch = 200;
+
+        // Get last admin log
+        $lizmapLogPath = jApp::logPath('lizmap-admin.log');
+        $lizmapLog = \Lizmap\App\FileTools::tail($lizmapLogPath, $maxLinesToFetch);
+        $lizmapLogTextArea = $this->logLinesDisplayTextArea($lizmapLog);
+
+        $errorLogDisplay = \lizmap::getServices()->hideSensitiveProperties();
```
laurentj, line 56: wrong assignement to $errorLogDisplay. hideSensitiveProperties() indicate if we hide sensitive properties. So, if yes, we must **not** show errors log.

`$errorLogDisplay = ! \lizmap::getServices()->hideSensitiveProperties();`

laurentj (changes requested):

PR #5185, opened 2025-01-13
Add X-Request-Id in request to QGIS Server headers

rldhont:
`X-Request-Id` correlates HTTP requests between a client and server. It will help to correlate logs between Lizmap Web client and QGIS Server. `X-Request-Id` is supported by Py-QGIS-Server.


rldhont:
Wait for https://github.com/3liz/lizmap-web-client/pull/5206

nworr (approved):

dmarteau:
Note: Py-qgis-server for now, does not take into account x-requesti-id but qjazz does.   Nevertheless, this is a good idea to set it in all cases.

PR #5188, opened 2025-01-13
Update assets/packages-lock.json after build

Gustry:
When building LWC, we have this file updated @nboisteault, is-it normal ?


nboisteault:
I don't have these changes executing `npm run build` w/ nodejs V20.5.1 and npm 9.8.0. Which versions do you have?

PR #5193, opened 2025-01-14
improve setLayerOpacity : split embedded and normal layer search with xpath

nworr:
Ticket : #

Funded by 3liz

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/QgisProject.php
+++ b/lizmap/modules/lizmap/lib/Project/QgisProject.php
@@ -377,32 +377,64 @@ class QgisProject
      */
     protected function setLayerOpacity(ProjectConfig $cfg)
     {
+        /**
+         * NOTE : searching for embedded layers with opacity is time consuming :
+         * need to loop over all layers to find embedded, then peform xpath on new Project files
+         * so : 1st loop to find all layers with opacity (non embedded)
+         * 2nd loop will find embedded layers and parse project file searching for opacity
+         */
         $layers = $this->layers;
-        // loop through project layers to get embedded ones as well
-        foreach ($layers as $layer) {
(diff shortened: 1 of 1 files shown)


Gustry:
@mind84 It's linked to #4925 and #5011

We got some 500 errors with "big" projects, this PR is trying to fix for your information.

lizmap/modules/lizmap/lib/Project/QgisProject.php
```diff
@@ -378,33 +378,54 @@ protected function setShortNames(ProjectConfig $cfg)
...
                     $layerCfg->opacity = $opacity;
                 }
             }
         }
+
+        // embedded layers contains id
+        foreach ($layers as $layer) {
+            $xpathEmbedded = "//maplayer[@id='{$layer['id']}']";
+            $embeddedLayer = $this->xpathQuery($xpathEmbedded);
+            if ($embeddedLayer) {
```
Gustry, line 404: Maybe it's only me, but I prefer when we negate checks like this.

```
if not something:
     continue
```

so we can keep code on the left side of the editor and not moving the logic to the right. And we are sure there isn't code executer under the the `if`.

My 2 cents, feel free to skip.

mdouchin (changes requested):
Thanks a lot ! 
Please add a big uppercase comment which describe why you do it this way, and which reminds future devs that this part can impact performance a lot.

By the way, have you seen elsewhere this kind of heavy calls for other embed layer properties ?

lizmap/modules/lizmap/lib/Project/QgisProject.php
laurentj: agree

nworr:
@mdouchin 

`readEditionLayers ` , `readEditionForms` and `readLayersLabeledFieldsConfig` also search for embedd layers, but it's not a wide search

mdouchin (approved):

PR #5194, opened 2025-01-14
lizmap-features-table - Improve the logic of the feature detail opening

mdouchin:
Improve the way the web component `<lizmap-features-table>` displays the popup of a feature when the user clicks on its title.

This adds a new way to load the component with a highlighted feature at startup, 
by adding a new optional attribute `data-active-item-feature-id`. 

If this attribute is set to a valid feature ID, the table will be hidden and the corresponding feature popup visible.

image

For example:

* By setting the attribute in the HTML element definition:

  ```html
  <lizmap-features-table
    sortingorder="asc" sortingfield="libsquart" 
    withgeometry="1" 
    expressionfilter="quartmno = 'MC'" 
    uniquefield="id" 
    data-center-to-highlighted-feature-geometry="true" 
    data-show-highlighted-feature-geometry="true" 
    data-active-item-feature-id="18"
  >
  <lizmap-features-table>
  ```

* By setting the attribute in Javascript:
  ```javascript
  // Get the element (the selector must probably be more specific.)
  const elt = document.querySelector('lizmap-features-table')
  // Change its attribute data-active-item-feature-id which will display the feature details
  elt.setAttribute('data-active-item-feature-id', 18);
  // Note here that you can also use the dataset shortcut
  elt.dataset.activeItemFeatureId = 23;

  // Move back to the list of features
  // Note here that you can also use the dataset shortcut
  elt.dataset.activeItemFeatureId = null
  ```

Ticket : #5146

Funded by 3liz

### DIFF
--- a/assets/src/components/FeaturesTable.js
+++ b/assets/src/components/FeaturesTable.js
@@ -25,9 +25,10 @@ import GeoJSON from 'ol/format/GeoJSON.js';
  *                        withGeometry="1" expressionFilter="quartmno = 'HO'"
  *                        uniqueField="id" layerId="subdistrict_24ceec66_e7fe_46a2_b57a_af5c50389649"
  *                        layerTitle="child sub-districts"
- *                        (optionnal) data-show-highlighted-feature-geometry="true"
- *                        (optionnal) data-center-to-highlighted-feature-geometry="true"
+ *                        (optional) data-show-highlighted-feature-geometry="true"
+ *                        (optional) data-center-to-highlighted-feature-geometry="true"
  *                        (optional) data-max-features="100"
+ *                        (optional) data-active-item-feature-id="5"
  *                        >
  *      <lizmap-field data-alias="District's name" data-description="Label of district's name">
  *         "libsquart"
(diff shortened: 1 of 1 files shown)


assets/src/components/FeaturesTable.js
```diff
@@ -186,113 +187,194 @@ export default class FeaturesTable extends HTMLElement {
...
+        return feature;
+    }
 
-        // Titles based on active status
-        const activeItemTitle = `${this.openPopup ? lizDict['featuresTable.item.active.hover']: ''}`;
-        const defaultItemTitle = `${this.openPopup ? lizDict['featuresTable.item.hover'] + '.': ''} ${this.itemsDraggable == 'yes' ? lizDict['featuresTable.item.draggable.hover'] + '.' : ''}`;
+    /**
+     * Toggle the display of the active feature details
+     *
+     * @param {number|null} featureId Feature ID
```
Gustry, line 207:
```suggestion
```

assets/src/components/FeaturesTable.js
```diff
@@ -186,113 +187,193 @@ export default class FeaturesTable extends HTMLElement {
...
+        // This is important to be able to create a new element
+        // with its attribute data-active-item-feature-id set to a feature ID
+        if (!this.features.length) return;
+
+        // Get the value store in the dataset attribute
+        const activeItemFeatureId = this.dataset.activeItemFeatureId;
+        // console.log(`Toggle. activeItemFeatureId = ${activeItemFeatureId}`);
+
+        // Depending on its value (empty or not)
+        // We should display the feature details
```
rldhont, line 219:
```suggestion
        // Depending on its value (empty or string null)
        // We should display the feature details
```

PR #5131, opened 2024-12-16
Draw: add button to rotate geometries

nboisteault:
Add a new button in the drawing toolbar to allow rotating the selected geometry.

Using https://viglino.github.io/ol-ext/examples/interaction/map.interaction.transform.html
Thanks @Viglino

https://github.com/user-attachments/assets/9d84bc2c-aa06-40e4-a485-00d72d5cac5e

Funded by 3Liz

PR #5195, opened 2025-01-14
Merge fallback locales in lizDict

Gustry:
New PR like #5139 or #5131 adding new strings in `lizDict` are missing translations when using LWC in another language than English, until we do the update with [Lizmap locales process](https://github.com/3liz/lizmap-locales).

I'm clearly not sure if it's the purpose in this file to add the fallback, or if it must be done elsewhere, upstream etc.

### DIFF
--- a/lizmap/modules/view/controllers/translate.classic.php
+++ b/lizmap/modules/view/controllers/translate.classic.php
@@ -3,7 +3,7 @@
 use Lizmap\App\LocalesLoader;
 
 /**
- * Service to provide translation dictionnary.
+ * Service to provide translation dictionary.
  *
  * @author    3liz
  * @copyright 2011-2022 3liz
@@ -15,7 +15,7 @@ use Lizmap\App\LocalesLoader;
 class translateCtrl extends jController
 {
     /**
-     * Get text/javascript containing all translation for the dictionnary.
+     * Get text/javascript containing all translation for the dictionary.
      *
      * @urlparam string $lang Language. Ex: fr_FR (optional)
      *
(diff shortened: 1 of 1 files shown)


laurentj (approved):
LGTM. LocalesLoader is using low level api of JLocale, so, no, we cannot backport it in upstream.

Note that this patch cannot be backported to lizmap 3.8 as 3.8 does not use low level api of jLocale, but use its own parser that is not very reliable (and does not take advantage of cache used by jLocale)

Gustry:
I'm not fully sure about the last talks we got in the chat about this PR.
Indeed, the best should be on "build" time, to add all missing strings in each languages, to English.
But then, it means the ZIP would fallback on English by default anyway, despite the setting in INI file. Or this PR should be done only if the fallback is different than `en_US` to avoid loading two files.

Side note, this dictionary file doesn't seem "too big", like 200 lines `lizmap/modules/view/locales/en_US/dictionnary.UTF-8.properties`.

lizmap/modules/view/controllers/translate.classic.php
```diff
@@ -36,6 +36,8 @@ public function index()
         }
 
         $data = LocalesLoader::getLocalesFrom('view~dictionnary', $lang);
+        $fallback = LocalesLoader::getLocalesFrom('view~dictionnary', \jApp::config()->fallbackLocale);
```
laurentj, line 39: Please enclose the new lines into this test : `if (strpos(\jApp::config()->jResponseHtml['plugins'], 'debugbar') !== false) {`. So it will be executed only in the "development mode".

PR #5204, opened 2025-01-15
Tests - Start Page Object Model for Playwright

Gustry:
Adding POM : https://playwright.dev/docs/pom

Superseeds #5167

### DIFF
--- a/tests/end2end/playwright/attribute-table.spec.js
+++ b/tests/end2end/playwright/attribute-table.spec.js
@@ -1,5 +1,6 @@
 // @ts-check
 import { test, expect } from '@playwright/test';
+import { ProjectPage } from './pages/project';
 import { gotoMap } from './globals';
 
 test.describe('Attribute table', () => {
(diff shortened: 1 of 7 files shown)


Gustry:
Tests should be OK now.

@mind84 you review is also welcome (as I can add you in the GH review list).

With this, we avoid "dangerous", not convenient string for coding. Using variables and functions is way more convinient in IDE with autocompletion. It also gives better maintenance for later, when a UI is updated.

If merged, please add relevant new locator and function into this class.

Let's see later how we do with the "legacy" `gotoMap` function.

rldhont (changes requested):
Please use `*.js` file instead of `*.ts`.
You can add `// @ts-check` at the first line to keep the `*.ts` file checking

Gustry:
@rldhont Thanks, it's done

rldhont (approved):
LGTM!

nboisteault (approved):
Thanks @Gustry

PR #5225, opened 2025-01-16
Tests - Extends base class for testing admin pages and landing page

Gustry:
Follow up from #5204

Add some base classes to test the admin side, or the landing page. I have checked a little bit with Cypress tests as well, for later.

I will stop here about Playrwight Object Model, it was to start to use this, for better maintenance.

I'm just not sure why we need a "new" object, in the test itself.


nboisteault (approved):

nworr (approved):
POM is great !! 

thanks

PR #5227, opened 2025-01-17
Tests - Add a private repository

Gustry:
CC @nboisteault @rldhont @nworr @mdouchin @rldhont


Gustry:
Yes I know, that's why I didn't have to update the `.gitignore` file. But you need to add it manually every-time when you start the docker compose stack, to be recognized as a Lizmap repository, this is not convenient...

Lizmap admin panel → Repositories → Add new repo

nboisteault (approved):

PR #5238, opened 2025-01-20
Tests - Move HTTP requests metadata from Cypress to Playwright

Gustry:


Gustry:
I want to start to flag tests which are "read only" (no file deletion, no Create/Update/Delete on the DB), so we can run them in parallele soon.

nboisteault (approved):

Issue #4863, opened 2024-10-14
[Bug]: popup - iframe stop working on LWC 3.8


josemvm:
### What is the bug? (in English)

LWC 3.7.11
popup shows iframe in second table
![imagem](https://github.com/user-attachments/assets/e91e5528-cc3b-4f93-a4be-343fc97c9bec)

LWC 3.8.3-pre
popup doesn´t show iframe in second table
![imagem](https://github.com/user-attachments/assets/558c4278-85b4-42ce-a847-9a0f3abaeb55)

LWC 3.7.11 and LWV 3.8.3-pre
data - everything works fine
![imagem](https://github.com/user-attachments/assets/45326f3d-986f-498f-aa52-e5a00fb940c7)

### Steps to reproduce the issue

try beteween these versions

### Versions, safeguards, check summary etc

**Versions :**

* Lizmap Web Client : 3.8.3-pre.7994
* Lizmap plugin : 4.4.2
* QGIS Desktop : 3.34.11
* QGIS Server : 3.34.11
* Py-QGIS-Server : not used
* QGIS Server plugin atlasprint : 3.4.1
* QGIS Server plugin lizmap_server : 2.11.0
* QGIS Server plugin wfsOutputExtension : 1.8.2

<details>
<summary>List of Lizmap Web Client modules :</summary>
<br/>
* multiauth : 1.2.2
</details>

<details>
<summary>List of safeguards :</summary>
<br/>
* Mode : normal
* Allow parent folder : no
* Prevent other drive : no
* Prevent PG service : no
* Prevent PG Auth DB : no
* Force PG user&pass : no
* Prevent ECW : no
</details>

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 22.04

### Browsers

Firefox

### Browsers version

131.0.2

josemvm:
this is the console output:

![imagem](https://github.com/user-attachments/assets/4fd0c792-a23d-4dca-84f7-1d2bbff6275d)

strange output... it's the same permissions to show the document in table (data tool) and as i said before on lwc 3.7 all works fine for popup tool and also for data tool

josemvm:
https://github.com/3liz/lizmap-web-client/blob/fd79e9701f32ff63ba39390f65738b02af3628d1/lizmap/modules/view/controllers/media.classic.php#L106

```
[...]

PR #5255, opened 2025-01-22
Allow same-origin for iframe

Gustry:
https://github.com/3liz/lizmap-web-client/issues/4863#issuecomment-2604626722

CC @josemvm I don't know if it fixes your problem.
Maybe you can try to add it in **QGIS** (not the JS)

This kind of addition looks like end-less, we will always encounter new users with different `sandbox` usages : https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#sandbox

I'm wondering if setAttribute is the correct method to use. Shouldn't it be like appendAttributes to not change existing attributes ?

```js
            if (node.nodeName === 'IFRAME') {
                let sandbox = '';
                if (node.getAttribute('sandbox')) {
                    sandbox = node.getAttribute('sandbox');
                }
                if (!sandbox.includes("allow-scripts")){sandbox += ' allow-scripts'}
                if (!sandbox.includes("allow-forms")){sandbox += ' allow-forms'}
                if (!sandbox.includes("allow-same-origin")){sandbox += ' allow-same-origin'}
                node.setAttribute('sandbox', sandbox);
            }
```
?

Paid support 3155

### DIFF
--- a/assets/src/modules/Utils.js
+++ b/assets/src/modules/Utils.js
@@ -211,7 +211,7 @@ export default class Utils {
     static sanitizeGFIContent(content) {
         DOMPurify.addHook('afterSanitizeAttributes', node => {
             if (node.nodeName === 'IFRAME') {
-                node.setAttribute('sandbox','allow-scripts allow-forms');
+                node.setAttribute('sandbox','allow-scripts allow-forms allow-same-origin');
             }
         });
         return DOMPurify.sanitize(content, {


Gustry:
I'm wondering if `setAttribute` is the correct method to use. Shouldn't it be like `appendAttributes` to not change existing attributes ?

josemvm:
the same error in the console...

Gustry:
> the same error in the console...

Please check the HTML output for the iframe, in the webbrowser. To be sure that the code is executed, if you are not building JS etc for instance

Gustry:
As you can see, my changes in the PR is not taken into account in your screenshot.
Changing LWC source code is not enough, as you need to "compile" the code I think ...

Maybe by adding the `allow-same-origin` within your webbrowser, by double clicking on the HTML node is enough (and maybe to trigger the refresh).

josemvm:
ok, but it doesn't change either, the pdf document is still not displayed

rldhont:
> When the embedded document has the same origin as the embedding page, it is strongly discouraged to use both allow-scripts and allow-same-origin, as that lets the embedded document remove the sandbox attribute — making it no more secure than not using the sandbox attribute at all.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

PR #5260, opened 2025-01-23
Tests about XSS - Switch from iframe to object

Gustry:
Linked to https://github.com/3liz/lizmap-web-client/issues/4863#issuecomment-2610250397

It's not working for me on, my local, maybe because of unsecure link (instead of proper HTTPS ?)
I don't have any relevent log, just the message displayed inside

### DIFF
--- a/tests/qgis-projects/tests/xss.qgs
+++ b/tests/qgis-projects/tests/xss.qgs
@@ -593,7 +593,12 @@ def my_form_open(dialog, layer, feature):
       <dataDefinedFieldProperties></dataDefinedFieldProperties>
       <widgets></widgets>
       <previewExpression>"ID"</previewExpression>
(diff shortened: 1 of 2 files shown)


Gustry:
image

Antoviscomi:
> Linked to [#4863 (comment)](https://github.com/3liz/lizmap-web-client/issues/4863#issuecomment-2610250397)
> 
> It's not working for me on, my local, maybe because of unsecure link (instead of proper HTTPS ?) I don't have any relevent log, just the message displayed inside

In my  test also doesn't works, in my opinion while `DOMPurify.sanitize` doesn't allow `type=pdf/application`. I tried to allow pdf and same origins via .htaccess server-side on Apache but none results, `DOMPurify.sanitize` ignore/sanitize server instructions. So it seems  we not have chances of solutions at the present

PR #5269, opened 2025-01-27
Form filter - Also search for NULL values when empty string is checked

mdouchin:
In Lizmap form filter, you can use a field containing several values by configuring a "Unique values" field type.
Lizmap show a list of checkboxes which allow the user to check which values should be used for the filter. For example:

image

In the example above, the checkbox with no label corresponds to the features which have a `NULL` value for this field.

This modification fixes the previous behaviour which triggered a filter `"acces" IN ( '' )` by adding a test on the `NULL` value. The new filter is the correct : ` ( "acces" IN ( '' ) OR acces IS NULL ) ` 

Funded by Destination Bretagne Sud - Golfe du Morbihan https://destination-bretagnesud.bzh/

### DIFF
--- a/assets/src/legacy/filter.js
+++ b/assets/src/legacy/filter.js
@@ -653,6 +653,9 @@ var lizLayerFilterTool = function () {
                 var filter = null;
                 var field = field_item['field'];
                 if (clist.length) {
+                    // Check if the value '' is present. If so, we should also search for NULL
+                    let hasEmptyValue = false;
+
                     // If there is a separator in the field values, and we need
                     // to explode the values into single items, we need to use
                     // LIKE statements joined with OR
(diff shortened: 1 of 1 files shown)


mdouchin:
cc @Gustry FYI

nboisteault:
@mdouchin I don't think there will be a new Lizmap 3.7 release, so the backport tag is unnecessary cc @Gustry 
Could you add a test?

Gustry:
For some tests, the best would be some basic JavaScript function to be able to use JSUnit.

To be inspired from Lizmap server to format SQL query, according to a list of parameters : 

* Test https://github.com/3liz/qgis-lizmap-server-plugin/blob/master/test/test_filter_by_polygon.py#L238
* Static function https://github.com/3liz/qgis-lizmap-server-plugin/blob/master/lizmap_server/filter_by_polygon.py#L433

I know it's not easy in this "legacy" JS file, just the idea

Gustry:
> @mdouchin I don't think there will be a new Lizmap 3.7 release, so the backport tag is unnecessary cc @Gustry

Yes, according to the initial planning https://github.com/3liz/lizmap-web-client/wiki/Versions#diagram

We decided to have **less** overlap between branches. As 3.8.0 has been released in September 2024 and we are already at 3.8.4, we considered 3.8.X to be stable enough and to not maintain an too many branches in parallel. 3.7 should be already EOL since december, we added one extra release in January.

image

mdouchin:
@rldhont  @nboisteault OK to merge ?

NB: I removed to backport-3.7 tag

nboisteault (approved):

rldhont (approved):

PR #5287, opened 2025-01-29
PHP - Increase minimum version to 8.2

Gustry:
PHP - Increase minimum version to 8.2


Gustry:
Linked to #5434 about the version of PHP inside the PHP-CS-Fixer container

Gustry:
* [] The two Dockerfiles in the repo must be upgraded as well : `dev/` and `docker/`.

Issue #3963, opened 2023-11-14
[Question]: Print user information


jserieye:
### What is the question?

I use the code below to retrieve information about the logged in user
 
```javascript
var login = $('#info-user-login').text();
console.log(login );
``` 
 
For the login, lastname and firstname variables it works.
By I have no return for email, phone, organization, postcode, city, comment, ...

The goal is to be able to reuse these variables to personalize layouts, for example with the organization's logo.

### Versions

**Versions :**

* Lizmap Web Client : 3.6.6-pre.6247
* Lizmap plugin : 3.18.1
* QGIS Desktop : 3.28.11
* QGIS Server : 3.28.10
* Py-QGIS-Server : 1.8.8
* QGIS Server plugin atlasprint : 3.3.2
* QGIS Server plugin cadastre : 1.18.0
* QGIS Server plugin lizmap_server : 2.8.0
* QGIS Server plugin wfsOutputExtension : 1.8.0

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### QGIS server version, only if the section above doesn't mention the QGIS Server version

3.28.10

### Operating system

Ubuntu

### Browsers

Chrome

### Browsers version

Chrome

Gustry:
Kind of duplicate of a question asked last Friday #3960

Issue #3960, opened 2023-11-08
[Question]:  How to send user email to Postgresql ?


nde34:
### What is the question?

Hello,

I'm trying to do this [ICI](https://docs.lizmap.com/3.5/fr/publish/customization/javascript.html) to get the user email transfered to postgresql when he save modification on the edition form.
```
var userFields = {
   login: 'your_lizmap_user_login_field',
   firstname: 'your_lizmap_user_firstname_field',
   lastname: 'your_lizmap_user_lastname_field',
   organization: 'your_lizmap_user_organization_field'
};
```

What's the name for the email variable ?

i tried

```
email:'your_lizmap_user_email_field'
mail:'your_lizmap_user_email_field'
```

Nothing is working.

Thanks

### Versions

**Versions :**

* Lizmap Web Client : 3.6.3
* Lizmap plugin : 3.17.0
* QGIS Desktop : 3.22.8
* QGIS Server : 3.22.15
* Py-QGIS-Server : 1.8.7
* QGIS Server plugin atlasprint : 3.3.2
* QGIS Server plugin cadastre : 1.17.3
* QGIS Server plugin lizmap_server : 2.7.1
* QGIS Server plugin wfsOutputExtension : 1.7.1

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu

### Browsers

Chrome

### Browsers version

Chrome

Gustry:
By checking what the JavaScript is doing, it's easy to check what field can be retrieved.

![image](https://github.com/3liz/lizmap-web-client/assets/1609292/24f56624-bd3b-4688-87bc-70bffcb153e3)

So I would say no, it's not possible from the web-browser side, email is not provided in the HTML.
I would do it with a PostgreSQL trigger, only on the server side with the table container all user information.

nde34:
@Gustry 
Yes that's what i did thank you
It could be better if it is possible to retrieve email just like other informations from the web-browser side, so it's not needed to create triggers

PR #5288, opened 2025-01-29
Header - Add authenticated user organization in hidden fields

mdouchin:
This allows a JavaScript script to get the current user organization and use it in the editing form.

Ticket : #

Funded by Ville d'Avignon https://cartes.mairie-avignon.com


mdouchin:
cc @sigavignon

Gustry:
Tests added ;-)

Linked to #3963 and #3960

mdouchin:
@Gustry thanks a lot for the tests. Do you think we should also add other fields (city, postcode, etc.) ?

Gustry:
If more are added, I would prefer something cleaner (and revert your change in the first commit), using data attributes ?

I'm curious what other websites do ? Do they expose all "sensible" data into the frontend ?

Gustry:
Way more comfortable to use

rldhont (approved):

mdouchin:
ok for merging, and agree with you regarding API

PR #5294, opened 2025-01-29
Improve ESLint

neo-garaix:
## Improvement of ESLint

* Improved ESLint configuration to new standards
* All `assets` files except `legacy` are scanned
* Added ESLint to the `lint.yml` workflow
* Applied ESLint and fixed issues

Funded by 3Liz


assets/eslint.config.mjs
Gustry, line 1: Should we move this file one folder up, at the root, or should have another file for the `tests` folder ?

Gustry:
It would be great if this can be added to `.pre-commit-config.yaml` if it's not too long (and other commands as well)

Note for myself later, have a look to depandabot

assets/eslint.config.mjs
neo-garaix: I think it's better to move it up instead of having another so we can have a single config file to manage

assets/package.json
```diff
@@ -6,13 +6,12 @@
   "scripts": {
     "build": "webpack --config webpack.prod.js",
     "watch": "webpack --watch --config webpack.dev.js",
-    "pretest": "eslint src",
-    "pretest-fix": "eslint --fix src"
+    "pretest": "npx eslint src --ignore-pattern src/legacy",
+    "pretest-fix": "npx eslint --fix src --ignore-pattern src/legacy"
```
Gustry, line 10: I think this change is linked to my previous question https://github.com/3liz/lizmap-web-client/pull/5294#discussion_r1934263498

As discussed this morning, it would make more sense to merge all `package.json` at the root, to have these `scripts` shared among all the source code (tests included). And playwright and jsunit should be merge in the `devDependancies` instead in the core part.

CC @nboisteault

Gustry:
Done in #5334

PR #5295, opened 2025-01-29
Tests e2e playwright : Lint tests and POM

rldhont:
Linting some playwright files

### DIFF
--- a/tests/end2end/playwright/globals.js
+++ b/tests/end2end/playwright/globals.js
@@ -1,7 +1,21 @@
 // @ts-check
 const { expect } = require('@playwright/test');
-import { Page } from '@playwright/test';
 
+/**
+ * Playwright Page
+ * @typedef {import('@playwright/test').Page} Page
+ */
+
+/**
+ * Integer
+ * @typedef {number} int
+ */
+
+/**
+ * Expect no errors in the map page
+ * @param {Page} page The page object
+ * @param {boolean} checkLayerTreeView Checking  that tree view contains layers
+ */
 async function NoErrors(page, checkLayerTreeView = true) {
     // No error
     await expect(page.locator('p.error-msg')).toHaveCount(0);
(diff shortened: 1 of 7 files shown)


tests/end2end/playwright/requests-metadata.spec.js
```diff
@@ -1,13 +1,18 @@
...
+
 const url = 'index.php/view/app/metadata';
 
 /**
  * Check for a JSON response about the metadata
- * @param {Response} response The response object
+ * @param {APIResponse} response The response object
  *
- * @return {JSON} The JSON response
+ * @return {Promise<any>} The JSON response
```
Gustry, line 15:
```suggestion
 * @return {Promise<JSON>} The JSON response
```
?

tests/end2end/playwright/globals.js
```diff
@@ -99,11 +110,10 @@ export async function reloadMap(page, check = true) {
 
 
 /**
- * editedFeatureIds function
  * Get the last IDs when saving features as JSON
  * @param {Page} page The page object
  *
- * @return {JSON} The JSON response, like {"id":"31"}
+ * @return {Promise<any>} The JSON response, like {"id":"31"}
```
Gustry, line 116: Same question

Gustry (approved):
LGTM after my comment, with `Any`, we are loosing some information ?
Can you share your command ?

tests/end2end/playwright/requests-metadata.spec.js
rldhont: `JSON.parse` returns `any` and not `JSON`. It could be an object, an array, a numer or a string.

tests/end2end/playwright/globals.js
rldhont: Same response, `JSON.parse` returns `any` and not `JSON`. It could be an object, an array, a number or a string.

PR #5303, opened 2025-01-30
Restructuration package.json

neo-garaix:
## Changes how package.json are structured in LWC

* Now we have a main `package.json` with **workspaces**
* Others `package.json` are linked to the main one
* Only one `npm install` is required to install all dependencies at the same location
* Scripts like `npm run js:test` in `LWC/tests/js-units` can be executed in `LWC/` with `npm run js:test --workspace=tests/js:test`
* Updated the `import` path for js unit tests
* Adapted CI

Funded by 3Liz


assets/package.json
```diff
@@ -5,14 +5,10 @@
   "type": "module",
   "scripts": {
     "build": "webpack --config webpack.prod.js",
-    "watch": "webpack --watch --config webpack.dev.js",
-    "pretest": "eslint src",
-    "pretest-fix": "eslint --fix src"
+    "watch": "webpack --watch --config webpack.dev.js"
```
Gustry, line 8: While you are here, can you add the jsdoc command ? Or should we add it in the parent package .json ?

Gustry: Can you tell me which `version` is used when doing the `jsdoc` ? this file or the one in the root ? (1.0.0) In other works, can you remove all non necerrasy keys from this JSON file here, and keep only in the top one ?

Gustry (commented):
Nice cleanup !

* Can you edit the `INSTALL.md` about building JS ?
* We need to look the `make build` command, to add the option `npm install --production` to not install dev dependancies in production packages

nboisteault (approved):

assets/package.json
neo-garaix: `version` is 3.10.0, I added the `jsdoc` command in the `package.json` at the root.

assets/package.json
Gustry, line 1: I think this file can be removed, same as the other one in `tests/js-units/package.json` and `tests/end2end/package.json` ?

neo-garaix: We need those files because script with `npm run` use them for example : 
- `assets/package.json` => npm run watch (webpack needs it) 
- `tests/js-units/package.json` => To run js:units tests, tests need the path to `assets/...` and it's done through worskpaces with multiple package.json communicating together
- same for `tests/end2end/package.json`, Playwright and Cypress need it

PR #5338, opened 2025-02-03
Tests - Update tests about axis orientation to use POM and TRUE/true

Gustry:
Tests - Update tests about axis orientation to use POM and TRUE/true


Gustry:
@rldhont Does it look good to you ?

I will disable "flaky" tests in another PR. I can really lead to unexpected merge ...

PR #5334, opened 2025-02-03
Update eslint

Gustry:
Rebased work from @neo-garaix about #5294

PR #5339, opened 2025-02-03
Update ESLint config to 9.19.0

Gustry:
Backport from #5334 with @neo-garaix


Gustry:
A little bit lazzy to do the backport ...

PR #5340, opened 2025-02-03
Tests e2e Playwright base layers: mocking tiles

rldhont:
To speed up end2end playwright base layers tests, OSM and GeoPF tiles has been saved and mocked.

We have save the real tiles because, these tests checking that tiles are well drawn in OL map.


tests/end2end/playwright/base-layers.spec.js
```diff
@@ -1,7 +1,13 @@
 // @ts-check
+import * as path from 'path';
+import * as fs from 'fs/promises'
+import { existsSync } from 'node:fs';
 import { test, expect } from '@playwright/test';
 import { gotoMap } from './globals';
 
+// To update MOCK files
```
Gustry, line 8:
```suggestion
// To update OSM tiles in the mock directory
// IMPORTANT, this must not be set to `true` while commiting, on GitHub. Set to `false`.
```

Gustry (approved):
Just to be clear, minor comment, otherwise LGTM

tests/end2end/playwright/base-layers.spec.js
rldhont: I have updated the comment. can you check it ?

Gustry: Ok for me

rldhont:
@Gustry I have added a commit with default tiles.

PR #5226, opened 2025-01-16
Tests e2e Playwright : Store logs in Project POM

rldhont:
Follow up #5204

Add a way to store and check logs in Project page.

PR #5351, opened 2025-02-04
Tests - Check client side logs

Gustry:
Superseeds #5226 from @rldhont 

I'm not a fan of this new parameter, because then we end with similar parameters in `gotoMap` ...

But if want to check for client side logs, better to do it step by step, without to many projects.

I was thinking about Playwright tags, with `testinfo` https://playwright.dev/docs/api/class-testinfo#test-info-tags


rldhont:
@Gustry I was thinking to remove expecting browser logs to be empty in open method, because it was in #5226 a simple test.

I was also thinking to provide method to get logs filtered by types, and to checked logs in test not in POM.

PR #5357, opened 2025-02-04
GH Action - Add workflow check concurrency, eslint only in assets/src for now

Gustry:


rldhont:
It seems `legacy` directory has to be ignored. For example, with `--ignore-pattern assets/src/legacy/*`, isn't it ?

Gustry:
> It seems legacy directory has to be ignored. For example, with --ignore-pattern "assets/src/legacy/*", isn't it ?

It was the configuration file which was not well formed.

PR #5362, opened 2025-02-05
Limit line length to 400 in JS

Gustry:
Until now, we got some lines up to 700 characters long.

This is not the best : 

* Difficult to see the difference **within** one line diff
* Difficult to track within GIT, (for instance, within this PR, it's difficult to see if I broke something)

I set to 400, I hope temporary. Before going further, I would like your idea. Optimum is 120, 130 (to have two files side-by-side, without softwrap). Also in GitHub/GitLab UI, lines are 120 characters.
If a line is more than 130, it's means that the code should be refactored, to introduce an intermediate variable.

Keen to have feedbacks : @mind84 @nworr @mdouchin @nboisteault @rldhont


rldhont (approved):

mdouchin (approved):
Thanks, agreed we need to avoid this in the future

nworr (approved):
1000 % approved, let's start with 400

Gustry:
Thanks for all these approvals ! :D 

> 1000 % approved, let's start with 400

My idea was to **quickly** reduce to something "more human". like 200, or at least 300.

assets/src/components/MousePosition.js
```diff
@@ -46,7 +46,8 @@ export default class MousePosition extends HTMLElement {
         return html`
             <div class="mouse-position">
                 <div class="editable-position ${['dm', 'dms', 'mgrs'].includes(this._displayUnit) ? 'hide' : ''}">
-                    <input type="number" step="any" placeholder="longitude" @input=${(event) => this._lonInput = parseFloat(event.target.value)} @keydown=${(event) => { if (event.key === 'Enter') { this._centerToCoords(); } }} .value=${isNaN(lon) ? 0 : lon}><input type="number" step="any" placeholder="latitude" @input=${(event) => this._latInput = parseFloat(event.target.value)} @keydown=${(event) => { if (event.key === 'Enter') { this._centerToCoords(); } }} .value=${isNaN(lat) ? 0 : lat}>
+                    <input type="number" step="any" placeholder="longitude" @input=${(event) => this._lonInput = parseFloat(event.target.value)} @keydown=${(event) => { if (event.key === 'Enter') { this._centerToCoords(); } }} .value=${isNaN(lon) ? 0 : lon}>
```
nboisteault, line 49: Line break between `input`s might change the UI but I'm not sure so it has to be visually tested

Gustry: Ok, I have added two variables to split the line. Is that OK ?
Do you think it's ok in `assets/src/components/Snapping.js` or should I do the same ?

Gustry:
Snapping panel : 
image

Mouse position : 
image

@nboisteault Is that OK for you ?

nboisteault (approved):

Gustry:
Arf, exactly why longs line are not good. It makes conflicts very easy... ;-)

tests/end2end/playwright/legend.spec.js
```diff
@@ -1,6 +1,7 @@
 // @ts-check
 import { test, expect } from '@playwright/test';
 import { gotoMap } from './globals';
+import {base64svgLineLayer} from "assets/src/modules/state/SymbologyIcons";
```
rldhont, line 4: This change blocked end2end tests.

```
Error: require() of ES Module /home/runner/work/lizmap-web-client/lizmap-web-client/assets/src/modules/state/SymbologyIcons.js from /home/runner/work/lizmap-web-client/lizmap-web-client/tests/end2end/playwright/legend.spec.js not supported.
Instead change the require of SymbologyIcons.js in /home/runner/work/lizmap-web-client/lizmap-web-client/tests/end2end/playwright/legend.spec.js to a dynamic import() which is available in all CommonJS modules.
```

PR #5365, opened 2025-02-05
Tests - Switch to parametrized test instead of manual test for data editing

Gustry:
Tests - Switch to parametrized test instead of manual test for data editing

I will try to parametrize it more with other geom.


Gustry:
PR Ready

As discussed with @mdouchin, I think it would be nice to have a public method to retrieve the HTML content popup, for a specific layer ID and/or feature ID. It helps a lot in tests like this one, but also for user defined scripts.

Is the current locator OK ?

Gustry:
Using data-attributes make the testing way easier.

But this kind of test will break the day if we change the default popup content (other "auto" HTML layout, or from "auto" to "QGIS HTML maptip" for this layer for another testing purpose).
To make the test more resilient, and not taking care of the UI aspect (covered by another test), I guess it would be a nice to a "raw data representation", similar to : 

```html
<div class="lizmapPopupSingleFeature" data-layer-id="polygons_1e227060_6105_4a4a_92f2_11863ebe65f1" data-feature-id="5" class="lizmapPopupSingleFeature">
  <data data-layer-name="Friend name of the layer">
    <field alias="Friendly field name" name="field_name" raw_field_content="">Friend value 1</field>
  </data>
  COMPLEX DATA representation with bootstrap, tabs, etc ...
</div>
```

Just quickly thinking

PR #5266, opened 2025-01-27
API - Add a method to retrieve the QGIS Lizmap server endpoint URL

Gustry:
Linked to #5262

PR #5262, opened 2025-01-24
PHP: caching QGIS Server metadata

rldhont:
Adding [Kevinrob/guzzle-cache-middleware](https://github.com/Kevinrob/guzzle-cache-middleware), to defined cache strategy for caching QGIS Server metadata.

The main goal is to reduce the number of request to QGIS Server to get it's metadata. The QGIS Server metadata request's result will be stored during 60 seconds. During this time, no more QGIS Server metadata request will be performed.

Adding a new `jCache` profile : `requests`.

PR #5369, opened 2025-02-06
Update lizmapServices.class.php

Antoviscomi:
Causes of bad url construction such as http://myhost/cgi-bin/qgis_mapserv.fcgi/lizmap//server.json adding a double slash before of server.json and avoid the correct recognization of Lizmap server plugin on Master

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmapServices.class.php
+++ b/lizmap/modules/lizmap/classes/lizmapServices.class.php
@@ -627,7 +627,7 @@ class lizmapServices
     {
         if (empty($this->lizmapPluginAPIURL)) {
             // When the Lizmap API URL is not set, we use the WMS server URL only
-            // and we add '/lizmap'/ at then end
+            // and we add '/lizmap' at then end
             return rtrim($this->wmsServerURL, '/').'/lizmap';
         }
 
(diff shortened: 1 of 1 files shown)


Gustry:
It's due to my PR #5266 on 3.9 and master

lizmap/modules/lizmap/classes/lizmapServices.class.php
```diff
@@ -628,7 +628,7 @@ public function getHostLizmapAPI()
         if (empty($this->lizmapPluginAPIURL)) {
             // When the Lizmap API URL is not set, we use the WMS server URL only
             // and we add '/lizmap'/ at then end
```
Gustry, line 630:
```suggestion
            // and we add '/lizmap' at then end
```

Gustry (commented):
Thanks for your contribution.

We need to check with #5262

rldhont (approved):

lizmap/modules/lizmap/classes/lizmapServices.class.php
Antoviscomi: Sorry it seems to me I'd apply

PR #5376, opened 2025-02-06
Fix tests e2e playwright JS module

rldhont:
Introduced by https://github.com/3liz/lizmap-web-client/pull/5362, import of Lizmap JS module could not be done.

So playwright has to be a `module` package, `require` has to be replaced by `import` and `__dirname` defined at each file level if needed.


Gustry (approved):
LGTM, thanks !

nboisteault (approved):
LGTM

tests/end2end/playwright/globals.js
```diff
@@ -12,6 +14,34 @@ import { URLSearchParams } from 'url';
...
+ */
+export const __dirname = path.dirname(__filename);
+
+/**
+ * Get the auth storage state path
+ * @param {string} name The file name without extension
+ * @returns {string} The path to auth storage state path
+ */
+export function getAuthStorageStatePath(name) {
+    return path.join(__dirname, './.auth/' + name + '.json')
```
Gustry, line 42: Side question, should we prefer string templates ? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#syntax

@nboisteault ?

nboisteault: I prefer string templates when there are multiple line breaks. In this case I don't have any preference.

rldhont: Thanks @Gustry to pointed it. The right syntax is not string template but more than 2 parameters for `path.join` like `path.join(__dirname, '.auth', name + '.json')`

PR #5389, opened 2025-02-10
Tests - Use toHaveLength

Gustry:
Before, it was a "poor" debug info : 

```
Error: expect(received).toBe(expected) // Object.is equality
    Expected: 15
    Received: 14
      522 |         })
      523 |         let getPrintParams = await expectParametersToContain('Print requests 1', getPrintRequest.postData() ?? '', expectedParameters1)
    > 524 |         await expect(getPrintParams.size).toBe(15)
```

Which is not meaning ful, because we don't know what was in the list before. I expect `toHaveLength` to show the list content ?
Otherwise, we will add a flag in `expectParametersToContain` to check the size of parameters and to display them nicely if faulty

### DIFF
--- a/tests/end2end/playwright/globals.js
+++ b/tests/end2end/playwright/globals.js
@@ -193,7 +193,7 @@ export async function expectParametersToContain(title, parameters, expectedParam
 
     await expect(
         searchParams.size,
-        `Check "${title}" : Not enough parameters compared to expected!`
+        `Check "${title}" : Not enough parameters compared to expected :\nGot ${Array.from(searchParams.keys()).join(', ')}!`
     ).toBeGreaterThanOrEqual(Object.keys(expectedParameters).length)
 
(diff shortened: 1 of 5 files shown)


Gustry:
Better, but not enough : 


```
     Error: expect(received).toHaveLength(expected)

    Expected length: 15
    Received length: 14
    Received array:  ["SERVICE", "REQUEST", "VERSION", "FORMAT", "TRANSPARENT", "CRS", "DPI", "TEMPLATE", "map0:EXTENT", "map0:SCALE", …]

      79 |         })
      80 |         let getPrintParams = await expectParametersToContain('Print requests 1', getPrintRequest.postData() ?? '', expectedParameters1)
    > 81 |         await expect(Array.from(getPrintParams.keys())).toHaveLength(15)
```

it doesn't show all values by default

PR #5375, opened 2025-02-06
ESLint - Limit lines to 350

Gustry:
Follow up from #5362

* Decrease to 350
* Fix regression from #5362 about these new variables I have added (and already backported in 3.9 in #5371)

PR #5413, opened 2025-02-14
Refactoring lizmap-selection-tool

rldhont:
Try to update a five years old custom element

Linked to #5375 to have ESLint with limit lines to less than 400 why not 120 !

### DIFF
--- a/assets/src/components/SelectionTool.js
+++ b/assets/src/components/SelectionTool.js
@@ -21,86 +21,179 @@ export default class SelectionTool extends HTMLElement {
 
     connectedCallback() {
 
-        const isFilterDisabled = mainLizmap.selectionTool.selectedFeaturesCount === 0 && mainLizmap.selectionTool.filteredFeaturesCount === 0;
-        const mainTemplate = () => html`
+        const titleTemplate = () => html`
+        <h3>
+            <span class="title">
+                <button
+                    type="button"
+                    class="btn-selectiontool-clear btn btn-sm btn-error btn-link"
+                    title="${lizDict['toolbar.content.stop']}"
+                    @click=${() => mainLizmap.selectionTool.disable()}
+                    >×</button>
+                <span class="icon-star icon-white"></span>
+                <span class="text">&nbsp;${lizDict['selectiontool.toolbar.title']}&nbsp;</span>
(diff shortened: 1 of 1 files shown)


assets/src/components/SelectionTool.js
```diff
@@ -117,12 +210,38 @@ export default class SelectionTool extends HTMLElement {
...
+                    this
+                );
             },
-            ['selectionTool.newAddRemoveSelected', 'selectionTool.allFeatureTypeSelected', 'selection.changed', 'filteredFeatures.changed', 'selection.bufferValue']
+            [
+                'selectionTool.newAddRemoveSelected',
+                'selectionTool.allFeatureTypeSelected',
+                'selection.changed',
+                'filteredFeatures.changed',
+                'selection.bufferValue'
```
Gustry, line 228:
```suggestion
                'selection.bufferValue',
```

Neat-picking, but having a trailing comma is better, it avoids the next user to add one if he wants to add a line at the end of the list, and to be have the line `selection.bufferValue` updated, while he didn't change anything to this line. I would agree to set this as a rule for future contributions.

While you are here, I would make it alphabetic, but that's also my own opinion. There is a usually a control for that in an IDE.

assets/src/components/SelectionTool.js
```diff
@@ -117,12 +210,38 @@ export default class SelectionTool extends HTMLElement {
...
     disconnectedCallback() {
     }
+
+    get isFilterDisabled () {
+        return mainLizmap.selectionTool.selectedFeaturesCount === 0 &&
+            mainLizmap.selectionTool.filteredFeaturesCount === 0;
+    }
+
+    get results () {
+        return mainLizmap.selectionTool.selectedFeaturesCount > 1 ?
```
Gustry, line 242: I would prefer having some `if else` with returns inside. It looks difficult to see there are two if nested.

Gustry (approved):
I'm really neat picking on these 2 comments, it's way better, thanks.

I will add JSDoc later.

rldhont:
@Gustry I have just resolved your suggestions and added docstrings.

Gustry (approved):
LGTM again

PR #5304, opened 2025-01-30
API - Retrieve invalid layers between GetCapabilities en CFG file

Gustry:
I'm keen to have some feedbacks here.

For now, globally in Lizmap, if an error occurs, there is nothing displayed in the user interface. The only way is to open the developer tools, which is not a good advice, our admin end-users shouldn't open the web developer tools.

It would be nice, at the end of loading a map, a UI box is displayed if there were any "error" during loading the map ?

Just to keen to have ideas

Issue #4938, opened 2024-11-05
[Bug]: Is the actions (fire stations/hydrants/buildings) example working on LMWC 3.8?


gioman:
### What is the bug? (in English)

I tried to put to work this

https://docs.lizmap.com/current/en/publish/lizmap_plugin/actions.html

which was working fine up to 3.6. I updated the JSON file as described in the docs, so there is no warning message on project load. Anyway but trying to run the action to look for the closes fire stations or to select all buildings in a buffer around a specific fire hydrant do not give any results, but also no error.

By the way, also your demo here

https://demo.lizmap.com/lizmap/index.php/view/map?repository=features&project=fire_hydrant_actions

behave exactly the same.

Thanks in advance.

### Steps to reproduce the issue

See above.

### Versions, safeguards, check summary etc

**Versions :**

* Lizmap Web Client : 3.8.2
* Lizmap plugin : 4.4.4
* QGIS Desktop : 3.34.9
* QGIS Server : 3.34.11
* Py-QGIS-Server : not used
* QGIS Server plugin atlasprint : 3.4.1
* QGIS Server plugin lizmap_server : 2.11.0
* QGIS Server plugin wfsOutputExtension : 1.8.0

<details>
<summary>List of Lizmap Web Client modules :</summary>
<br/>
* lizmapdemo : 3.6.0a1
</details>

<details>
<summary>List of safeguards :</summary>
<br/>
* Mode : safe<br/>
* Allow parent folder : no<br/>
* Prevent other drive : yes<br/>
* Prevent PG service : yes<br/>
* Prevent PG Auth DB : yes<br/>
* Force PG user&pass : yes<br/>
* Prevent ECW : yes<br/>
</details>

### Check Lizmap plugin

- [X] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 22.04

### Browsers

Firefox, Chrome

### Browsers version

Latest updates.

SecondGIS:
Hello there,
I am still using LWC 3.6 since actions are currently not working in later versions ; as far as I understand from 3.6 documentation, there was no `map-center` parameter to be passed to `lizmap_get_data()` function in that version.
Anyone has any suggestion as to how to build that `map-center` parameter and use it in my SQL query inside `lizmap_get_data()` function with LWC 3.6 ?
Thanks for your help

wojtek19022:
[...]

PR #5424, opened 2025-02-18
Log all error and warning in the console, and display them

Gustry:
Linked to #4938 

Similar request in my PR #5304 : An error was raised in the console about invalid layer, but the admin user was not aware about it.

I think all `console.warn` and `console.error` should be reviewed, and we should use a kind of proxy to display in UI the error message, and a more detailed error message if the user is an admin.

Errors logged in the console are not visible, which make difficult for Lizmap users who doesn't know this feature and do not think about it.

I think only the "proxy" should do the `console.error`, and a corresponding Javascript error must be displayed with a collapsible div like : 

An error occured while running the action "name"

<details>
<summary>See more detail about the error</summary>
<br/>
Error: No fill, stroke, point, or text symbolizer properties in style: {"graphicName":"circle","pointRadius":6,"fill":true,"fillColor":"lightred","fillOpacity":0.3,"stroke":true,"strokeWidth":4,"strokeColor":"red","strokeOpacity":0.8}
</details>

PR totally broken for now, but I want some ideas before digging. I noticed there are different `addMessage`, there is also a `Message` component.

### DIFF
--- a/assets/src/components/Print.js
+++ b/assets/src/components/Print.js
@@ -408,8 +408,7 @@ export default class Print extends HTMLElement {
 
             document.querySelector('#message .print-in-progress button').click();
         }, (errorEvent) => {
-            console.error(errorEvent)
-            mainLizmap._lizmap3.addMessage(lizDict['print.error'], 'danger', true).addClass('print-error');
+            mainLizmap._lizmap3.addMessage(lizDict['print.error'], 'danger', true, errorEvent).addClass('print-error');
         });
     }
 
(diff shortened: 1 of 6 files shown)


Gustry:
I was quickly thinking : 

```mermaid
stateDiagram-v2
    file_a : File A.js
    file_b : File B.js
    log : Logger (according to level of message, current login, state...)
    console: Console
    ui: UI generic error message
    ui_admin: UI with connected as admin, more complete error message
    server: Log on the server, a new jLog client-side-errors.log
    file_a --> log
    file_b --> log
    log --> ui
    log --> ui_admin
    log --> server
    log --> console
```

Most of logger in desktop applications have this kind of setup.

PR #5470, opened 2025-02-25
Tests - Skip Lizmap warning when loading the map about outdated CFG if needed in tests

Gustry:
Tests - Skip Lizmap warning when loading the map about outdated CFG if needed in tests


Gustry:
PR approved by @rldhont

Issue #5452, opened 2025-02-21
[Bug]: Bug after reload page


Mavialle:
### What is the bug? (in English)

Hello,

This message follows ticket #5353, as we have identified the cause of the bug.

We recently updated Lizmap. Since then, a weird thing has happened on some interfaces : you connect to the interface for the first time, it works, and if you reload the page, nothing works. As if the page is not loading.

I noticed in the dev tools of Chrome that in network requests, a FETCH request named "service" was not fired when the page did not load. Any idea to solve the problem?

### Steps to reproduce the issue

1. Create a new QGIS project named "test.qgs" with QGIS Desktop 3.34.15
2. Open a layer from PostgreSQL created by : 

```sql
CREATE TABLE IF NOT EXISTS test.test_ligne
(
    gid serial NOT NULL,
    texte character varying,
    liste character varying,
    date date,
    file character varying,
    geom geometry(LineString,2154),
    CONSTRAINT test_ligne_pkey PRIMARY KEY (gid)
);

CREATE INDEX IF NOT EXISTS sidx_test_ligne_geom
    ON test.test_ligne USING gist
    (geom)
    TABLESPACE pg_default;
```

3. Open project proprieties > QGIS Server Menu > WFS tab > Check "published" for test_ligne
4. In QGIS, open the Lizmap plugin (version 4.4.7) and : 
     - in the layer menu, check "checked" for the test_ligne layer
     - Click the OK button at the bottom right to close the Lizmap plugin panel
     - If necessary, make the corrections reported in the plugin control tool using the “self-correction tools” menu.
5. Save the QGIS project
6. Put the qgis project and cfg file on Lizmap server
7. open the project on Lizmap : it's OK
8. Reload the page : the page loads poorly, for example, the layer no longer appears on the map or in the layer tree
9. In QGIS, open Lizmap plugin again, and in menu "edit layers" add test_ligne with the five options (create, without geom, etc), clcik on OK to close panel of edit layer and ok again to close Lizmap panel
10. Put the qgis project and cfg file on Lizmap server
11. open the project on Lizmap : it's OK
12. Reload the page: it's always OK

### Versions, safeguards, check summary etc

**Versions :**

* Lizmap Web Client : 3.8.5 - commit 16111314b https://github.com/3liz/lizmap-web-client/commit/16111314b
* Lizmap plugin : 4.4.7
* QGIS Desktop : 3.34.15
* QGIS Server : 3.34.15
[...]

PR #5471, opened 2025-02-25
[Fix JS] mainLizmap may not be ready when configs are loaded

rldhont:
mainLizmap may not be ready when configs are loaded, so we have to wait until it is ready to listen to `configsloaded` event.

Fixes https://github.com/3liz/lizmap-web-client/issues/5452


Gustry (approved):
LGTM

I think you have fixed some strings in the meantime

PR #5478, opened 2025-02-26
Admin - Display if the installation is not complete according to versions in installed modules

Gustry:
It's checking the version in `project.xml` against all "core Lizmap" modules thanks to the `installer.ini.php`.

image

It's to avoid some possible migrations issues

It will be manually backported to 3.9 and 3.8 due to Bootstrap 5.

### DIFF
--- a/lizmap/modules/lizmap/lib/Server/Server.php
+++ b/lizmap/modules/lizmap/lib/Server/Server.php
@@ -225,7 +225,7 @@ class Server
         $data = array();
         $modules = new ModulesChecker();
 
-        foreach ($modules->getList(false) as $info) {
+        foreach ($modules->getList(false, false, true) as $info) {
             $data[$info->slug] = array(
                 'version' => $info->version,
                 'core' => $info->isCore,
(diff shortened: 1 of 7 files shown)


lizmap/modules/lizmap/lib/Server/Server.php
```diff
@@ -269,11 +269,15 @@ private function getLizmapMetadata()
...
         $projectInfos = AppInfos::load();
         // Version
+        $lizmapVersion = $projectInfos->version;
         $data['info'] = array();
-        $data['info']['version'] = $projectInfos->version;
+        $data['info']['version'] = $lizmapVersion;
         $data['info']['date'] = $projectInfos->versionDate;
         $data['info']['commit'] = \jApp::config()->commitSha;
 
+        $modules = new ModulesChecker();
```
nworr, line 278: it is not necessary to call `ModulesChecker` in the class `Server`, 

If we want the `installation_complete` info in the json returned by `index.php/view/app/metadata`
, we can call the class in the controller `view/app` [controller](https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/view/controllers/app.classic.php)

nworr (commented):
Nice Addition !

rldhont (approved):

Gustry:
Note for myself, on snap servers : 

```bash
2025-03-07 16:43:11	176.138.102.239	error	Found module "admin" with version "3.10.0-pre", expected version "3.10.0-pre.8697".
```

Issue #4984, opened 2024-11-15
[Bug]: group as layer (type raster) - print doesn´t respect the opacity


josemvm:
### What is the bug? (in English)

if the user define an opacity different from 100%, the printer tool doesn´t take this option and the layout is printig with opacity 100%

### Steps to reproduce the issue

try to define an opacity different from 100% for group as layer (type raster) and try to print a layout

### Versions, safeguards, check summary etc

**Versions :**

* Lizmap Web Client : 3.8.4-pre.8065
* Lizmap plugin : 4.4.4
* QGIS Desktop : 3.34.11
* QGIS Server : 3.34.12
* Py-QGIS-Server : not used
* QGIS Server plugin atlasprint : 3.4.1
* QGIS Server plugin lizmap_server : 2.11.2
* QGIS Server plugin wfsOutputExtension : 1.8.2

<details>
<summary>List of Lizmap Web Client modules :</summary>
<br/>
* multiauth : 1.2.2
</details>

<details>
<summary>List of safeguards :</summary>
<br/>
* Mode : normal<br/>
* Allow parent folder : no<br/>
* Prevent other drive : no<br/>
* Prevent PG service : no<br/>
* Prevent PG Auth DB : no<br/>
* Force PG user&pass : no<br/>
* Prevent ECW : no<br/>
</details>

### Check Lizmap plugin

- [ ] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 22.04

### Browsers

Firefox

### Browsers version

132.0.2

josemvm:
the same for opacity set at group level in the layer tree (vector and raster layers)

josemvm:
hi @mind84 

> Print layout is the main reason for this pr so I would say yes.
I haven't tested the "group as layer" but individual raster layers work fine
Please let me know if https://github.com/3liz/lizmap-web-client/issues/4984 is still valid after this pr.

please can you try to fix it?

many thanks in advance

josemvm:
> group as layer and the same for opacity set at group level in the layer tree (vector and raster layers)

hi @mind84 can you confirm this issue on your development branch?

Gustry:
Please try with new version 3.8.6 : https://github.com/3liz/lizmap-web-client/releases/tag/3.8.6

josemvm:
very, very nice!
thanks a lot @mind84 (Faunalia, @pcav)

PR #5457, opened 2025-02-22
JS refactoring digitizing module and component

rldhont:
Refactoring with 120 max-len line

PR #5485, opened 2025-02-28
[Bugfix] Consider the opacity of groups for printing

mind84:
Now the `total opacity` is considered when printing, instead of the individual layer opacity.

By `total opacity` I mean the product of opacity of the individual layer with the opacity of the all parent groups (hierarchically),  `groupAsLayer` layers included.

Ticket : #4984

Backport 3.8, 3.9

Funded by Faunalia


assets/src/modules/state/Layer.js
```diff
@@ -552,6 +552,19 @@ export class LayerItemState extends EventDispatcher {
         }
         return this._visibility;
     }
+
+    /**
+     * Calculate total opacity by including also all parent gropus opacity values
```
Gustry, line 557:
```suggestion
     * Calculate total opacity by including also all parent groups opacity values
```

tests/end2end/playwright/print.spec.js
```diff
@@ -337,6 +337,127 @@ test.describe('Print', () => {
     });
 });
 
+test.describe('Print opacities', () => {
+    test.beforeEach(async ({ page }) => {
+        const url = '/index.php/view/map/?repository=testsrepository&project=group_as_layer_opacity';
+        await gotoMap(url, page);
```
Gustry, line 343: This `gotoMap` is deprecated, as "new" tests must use `POM` https://playwright.dev/docs/pom. Look in "playwright/pages"
You can add some locators about the "print" box, such as the scale, the buttons...

Gustry (commented):
Thanks @mind84.
Just some comments on the code style, as you could see, there some refactoring about POM, eslint, stylelint and coding rules in the past few weeks.

What is left to do to close #4984 ?

mind84:
> What is left to do to close https://github.com/3liz/lizmap-web-client/issues/4984 ?

I thinks that's it. 
I did not tag the PR with the `fixes` keyword because I'm not really sure this is really what is needed to solve the issue (but probably yes)

tests/end2end/playwright/print.spec.js
Gustry: Please also add the tag `@readonly` to your new test.

mind84:
Hi @Gustry,

I added the new test. Let me know if its ok, suggestions are appreciated!

Thanks!

tests/end2end/playwright/print.spec.js
```diff
@@ -337,6 +338,132 @@ test.describe('Print', () => {
...
+            // set opacity of `Group_opacity` (group as layer) layer to 60%
+            await page.getByTestId('Group_opacity').hover();
+            await page.getByTestId('Group_opacity').locator('i').nth(1).click();
+            await page.getByRole('link', { name: '60' }).click();
+
+            // opacity notes:
+            // the `raster` layer already has 80% opacity (inherited from QGIS project)
+            // setting the opacity to 60% on the `Group_opacity` causes:
+            // - `raster` layer to have a final total opacity of 48% (0.8*0.6) -> OPACITIES PARAM = 255*0.48 ~= 122
+            // - `single_wms_polygons` layer to have a final total opacity of 60% (1*0.6) _> OPACITIES PARAM = 255*0.6 = 153
```
Gustry, line 364: Sorry, nitpicking, but we have reduced max lines length to 120, see for instance https://github.com/3liz/lizmap-web-client/pull/5457

It's a long work in progress, but new code should apply this rule. Can you split this line, and also others L378, L399...

tests/end2end/playwright/print.spec.js
```diff
@@ -337,6 +338,132 @@ test.describe('Print', () => {
...
+
+        test('Layers in group with opacity', async ({ page }) => {
+            const printPage = new PrintPage(page, 'group_as_layer_opacity');
+            await printPage.open();
+            await printPage.openPrintPanel();
+            await printPage.setPrintScale('50000');
+
+            // set opacity of `Group_a` group opacity to 60%
+            await page.getByTestId('Group_a').locator('div').filter({ hasText: 'Group_a' }).first().hover();
+            await page.locator('.icon-info-sign').first().click();
```
Gustry, line 410: It would be a great move if these two lines can be moved to the `project` page, with a method `openLayerInfo(layerName)`.

So then we can just do 
```js
openLayerInfo('Group_a');
openLayerInfo('single_wms_points_group');
openLayerInfo('single_wms_polygons_group_as_layer');
```

You can look at existing method like `openAttributeTable`.

tests/end2end/playwright/print.spec.js
mind84: Yes of course

Is there a `linter` o some automatic tool to launch for line length checking?

tests/end2end/playwright/print.spec.js
mind84: Yep, I thought about that too but the `locator` is different for groups and and single layers. 
I try to build a function that take into account this difference

tests/end2end/playwright/print.spec.js
Gustry: I should install [pre-commit](https://pre-commit.com/) : 
https://github.com/3liz/lizmap-web-client/blob/master/CONTRIBUTING.md#pre-commit
It helps a lot.

But the current line-length is at [290](https://github.com/3liz/lizmap-web-client/blob/master/eslint.config.mjs#L78). We need to reduce step by step.

Set a visual margin in your IDE.

tests/end2end/playwright/print.spec.js
Gustry: Ok, I'm pretty sure with the "new" JS API in Lizmap that we can now if the label in the legend is a group or a single layer (cc @rldhont)

But in the meantime, why not `openLayerInfo()` and `openGroupLayerInfo` ?
Then it can be easily refactored in another step.

Because I know there a few tests opening this "Info" box, so it would help to migrate to POM.

mind84: > But in the meantime, why not openLayerInfo() and openGroupLayerInfo ?

So far I've managed to do it in just one function, LGTM in this way.

Thanks

Gustry (approved):
LGTM

Thanks for these updates

Gustry:
@mind84 You will need to take care of the 3.8 backport.

PR #5525, opened 2025-03-06
fix: In Guzzle, provided HTTP headers is an array

Gustry:
fix: In Guzzle, provided HTTP headers is an array

Superseeds #5510


Gustry:
Actually, except the docstring, this patch is valid for 3.8 as well. I can edit online the docstring on GitHub about the docstring, when the request is done by Guzzle (so without the nested array)

rldhont (approved):

PR #5529, opened 2025-03-06
Clean lizmap/www/assets/jelix to be up to date with jelix-www

rldhont:
* Clear `lizmap/www/assets/jelix`
* Copy `jelix-www` into `lizmap/www/assets/jelix`


lizmap/install/installer.php
```diff
@@ -4,4 +4,6 @@
 
 require __DIR__.'/../application.init.php';
 
+jFile::copyDirectoryContent(__DIR__.'/../vendor/jelix/jelix/lib/jelix-www', __DIR__.'/../www/assets/jelix');
```
laurentj, line 7: No. The jelix module installer already does this copy... 

I guess the main issue we have to fix is in the build step of the zip package, where we should copy the directory into the build directory.

laurentj (changes requested):

lizmap/install/installer.php
rldhont: Thanks I have removed this line

PR #5532, opened 2025-03-07
Speedup Lizmap Html page loading by adding preload link

rldhont:
Documentation about preload link https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/preload


nboisteault:
I'm not sure but it might block the HTML rendering.
It would be interesting to test against lighthouse to see if it improves its score.

josemvm:
hi @rldhont
sorry, i don't know if the problem is related to this issue, but in lwc 3.8 there is a strong and visible delay in loading project images after cleanning the lizmap cache

rldhont:
> I'm not sure but it might block the HTML rendering. It would be interesting to test against lighthouse to see if it improves its score.

From https://stackoverflow.com/questions/36641137/how-exactly-does-link-rel-preload-work
> For example, the application can use the preload keyword to initiate early, high-priority, and non-render-blocking fetch of a CSS resource that can then be applied by the application at appropriate time:

lizmap/app/responses/AbstractLizmapHtmlResponse.php
```diff
@@ -16,6 +16,24 @@ class AbstractLizmapHtmlResponse extends jResponseHtml
 {
     protected $CSPPropName = 'mapCSPHeader';
 
+    protected function outputJsScriptTag($fileUrl, $scriptParams)
```
nworr, line 19: These functions will _implicitly_ add a `preload` tag for each resource loaded with `addJSLink`/`addCssLink`, no ?

PR #5534, opened 2025-03-07
Install boostrap file (css, js) using composer

nworr:
Ticket : #

Funded by


rldhont:
The same could be done for release_3_9 because v2.3.2 is availble too https://github.com/twbs/bootstrap/releases/tag/v2.3.2

nworr:
> The same could be done for release_3_9 because v2.3.2 is availble too https://github.com/twbs/bootstrap/releases/tag/v2.3.2

I checked the content of the package, minified version of js/css file or not included and must be build.

PR #5536, opened 2025-03-07
fix: The version of LWC has a commit number in the version name, drop it for comparing to modules version

Gustry:
Fix regression from #5478

On Snap servers, the version is `3.10.0-pre.8697` while module versions are like `3.10.0-pre`. This was raising the message from PR #5478

AS this is for 3.8 branch, I will do another refactoring by moving version functions to this new file in another PR targeting only 3.9.


nworr (approved):
Nice addition thanks !

PR #5538, opened 2025-03-10
fix: The version of LWC has a commit number in the version name, drop…

Gustry:
Backport #5536 
Fix regression from https://github.com/3liz/lizmap-web-client/pull/5478


Gustry:
Unrelated test failure

PR #5542, opened 2025-03-10
Tests - Remove DB user and password from QGS files, use only the PG s…

Gustry:
Remove `user` and `password` from QGS files in `tests` folder. These projects are deployed in different endpoints such as 

https://sandbox.lizmap.com/lizmap_3_7
https://sandbox.lizmap.com/lizmap_3_8/
https://sandbox.lizmap.com/lizmap_3_9/

with different PG services, edited on runtime.

@mind84 It seems to mainly come from your QGIS Desktop, can you check and edit your connection in QGIS to drop the user and password. Thanks.

It will be manually backported to 3.9 and 3.8

### DIFF
--- a/tests/qgis-projects/tests/children_in_relation_div.qgs
+++ b/tests/qgis-projects/tests/children_in_relation_div.qgs
@@ -22,7 +22,7 @@
     <customproperties>
       <Option/>
     </customproperties>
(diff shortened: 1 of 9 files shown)


mind84:
@Gustry,
sorry for the inconvenience.

I double check in my QGIS, this is coming from an old configuration, now the connection is clean.

Thanks

PR #5544, opened 2025-03-11
[Bugfix] Wrong measure when both angle and dimension constraints are set on drawing tool

mind84:
When both dimension and angle constraints are set in the drawing tool, the resulting `linestring` length measurement is incorrect.

measurement_!

measurement_2

TEST NOTICE:
@Gustry,

I assumed that the `draw.spec.js` test will be refactored to be POM compliant, so for now I just added a new subtest in the "traditional" way. It makes more sense to me to refactor all the code at once, so I think it's best to keep it in a separate PR.

Backport 3.8, 3.9

Funded by Etra

### DIFF
--- a/tests/end2end/playwright/draw.spec.js
+++ b/tests/end2end/playwright/draw.spec.js
@@ -1,6 +1,7 @@
 // @ts-check
 import { test, expect } from '@playwright/test';
 import { gotoMap, reloadMap } from './globals';
+import {DrawPage} from "./pages/drawpage";
 
 test.describe('Draw', () => {
 
@@ -299,7 +300,7 @@ test.describe('Draw', () => {
         await page.locator('#draw button.digitizing-toggle-measure').click();
         await expect(page.locator('.ol-tooltip.ol-tooltip-static')).toBeVisible();
         await expect(page.locator('.ol-tooltip.ol-tooltip-static')).toHaveText('3.3 km34.27 km2');
-    })
+    });
 
     test('From local storage', async ({ page }) => {
         const the_json = '[{"type":"Polygon","color":"#000000","coords":[[[764321.0416656,6290805.935670358],[767628.3399468632,6290805.935670358],[767628.3399468632,6295105.423436],[764321.0416656,6295105.423436],[764321.0416656,6290805.935670358],[764321.0416656,6290805.935670358]]]}]';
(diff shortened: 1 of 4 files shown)


Gustry:
Thanks for the fix.

IMHO, I don't see any drawbacks to not move to POM right now for newly added tests. It makes tests way easier to read.

I have changed, manually, many `await page.locator('#newOlMap').click();` etc to `clickOnMap` from POM. Better to do it now according to me.

It is similar to Cypress. No new Cypress tests were added for one or two years now, but there are still so many : https://github.com/3liz/lizmap-web-client/tree/master/tests/end2end/cypress/integration because it's not "fun" to migrate them so it will takes a few months/years to get rid off them... :/

Some playwright files have a single POM test among a lot of "legacy" playwright test, we do not mind. Or if you want to create a separate file like `tests/end2end/playwright/draw-with-constraints.spec.js`, it's OK.

tests/end2end/playwright/draw.spec.js
```diff
@@ -301,6 +302,57 @@ test.describe('Draw', () => {
         await expect(page.locator('.ol-tooltip.ol-tooltip-static')).toHaveText('3.3 km34.27 km2');
     })
 
+    test('Measure - length and angle constraints', async ({ page }) => {
```
Gustry, line 305: New tests should also be tagged `@readonly` : https://github.com/3liz/lizmap-web-client/tree/master/tests#writing-tests

tests/end2end/playwright/draw.spec.js
```diff
@@ -301,6 +302,57 @@ test.describe('Draw', () => {
...
+        await drawProject.toggleMeasure();
+        // set constraint
+        await drawProject.setMeasureConstraint('distance','1500');
+
+        // draw a single fixed length line
+        await drawProject.clickOnMap(450,75);
+        await drawProject.dblClickOnMap(480,115);
+
+        // check final length measure
+        await expect(page.locator('.ol-tooltip.ol-tooltip-static')).toBeVisible();
```
Gustry, line 320: It seems the test just below `Circular geometry measure` and others have the same locator. It would make sense to have these locators in project POM ? Same as `.ol-tooltip.ol-tooltip-static` below ?
Like mapAnnotationToolTipStatic and measure or better name.

This will help later for migration.

nth can be let here.

tests/end2end/playwright/pages/drawpage.js
```diff
@@ -0,0 +1,99 @@
...
+         */
+        const listener = (dialog) => {
+            dialog.accept();
+            this.page.off('dialog',listener);
+        }
+        this.page.on('dialog', listener);
+
+        await this.drawPanel.locator('.digitizing-all').click();
+    }
+}
```
Gustry, line 99: Did you install `pre-commit` ? This end of line should be fixed by pre-commit config.

Gustry (approved):
Thanks, it's way better with POM, it makes the reading of tests better, to understand and review.

LGTM after my comments

tests/end2end/playwright/draw.spec.js
mind84: Yes indeed, thanks

tests/end2end/playwright/draw.spec.js
```diff
@@ -299,7 +300,62 @@ test.describe('Draw', () => {
...
+            await drawProject.deleteAllDrawings();
+
+            // add angle constraint
+            await drawProject.setMeasureConstraint('angle','37');
+            // draw a feature without finalize it
+            await drawProject.clickOnMap(395,246);
+            await drawProject.clickOnMap(423,263);
+
+            // check measure tooltips before finalize. The first should contains degrees info
+            await expect(page.locator('.ol-tooltip.ol-tooltip-measure').nth(0)).toHaveText('1.5 km37°');
```
Gustry, line 349: As I suggested in the previous comment, this one as well.

mind84:
@Gustry,

as note, if the PR is ok, I have to manually backport it to 3.8 (same issue solved by https://github.com/3liz/lizmap-web-client/pull/5463)

Thanks

mind84:
Hi @rldhont,

 > but your e2e test seems flacky

On my test environment runs smoothly, any suggestion for fixing it? Maybe move the test in a separated `describe` section?

Gustry:
> On my test environment runs smoothly, any suggestion for fixing it? Maybe move the test in a separated describe section?

I would say yes, without the `beforeEach`, just do not forget the `open()` after the page.

Can you rebase at the same time on master ?

PR #4266, opened 2024-03-04
New module presentation

mdouchin:
TODO

image

Funded by QAIR International https://www.qair.energy/

PR #5561, opened 2025-03-14
feat: Add presentation module to core

Gustry:
Revival of #4266

presentation_module

Because of the migration of Bootstrap 5, this will be manual backported to 3.10 master branch.

Some things need to be done before merging this PR : 

* [ ] E2E tests
* [ ] Core optional module (there are some `fixme` related to this)
* [ ] Namespace issue when loading a project

Funded by Qair https://www.qair.energy/


lizmap/modules/presentation/install/configure.php
```diff
@@ -0,0 +1,22 @@
...
+ * @see      http://3liz.com
+ *
+ * @license   Mozilla Public License : http://www.mozilla.org/MPL/
+ */
+class presentationModuleConfigurator extends Configurator
+{
+    public function getFilesToCopy()
+    {
+        return array(
+            '../www/css' => 'www:modules-assets/presentation/css',
```
laurentj, line 18: this is not the right path

laurentj (changes requested):

Gustry:
Due to lack of time/human resource for LWC 3.9 schedule, this will not land in LWC 3.9. It's postponed.

Note, the installation must be check, for instance if `publishers` exists in the DB

PR #5570, opened 2025-03-26
[Bugfix] JS Search: IGN address search uses map view instead of maxExtent

rldhont:
The IGN address search uses the map view instead of maxExtent. So if the user a zoom, he cannot get back to an other place contains by the map's max extent.

Funded by WPD Onshore France

### DIFF
--- a/assets/src/modules/Search.js
+++ b/assets/src/modules/Search.js
@@ -5,8 +5,6 @@
  * @license MPL-2.0
  */
 
-import { transformExtent } from 'ol/proj.js';
-
 /**
  * @class
  * @name Search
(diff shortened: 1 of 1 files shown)


assets/src/modules/Search.js
```diff
@@ -226,21 +224,31 @@ export default class Search {
...
-                        let text = '';
-                        let count = 0;
-                        for (const result of data.results) {
-                            var lab = result.fulltext.replace(labrex, '<strong class="highlight">$1</strong>');
-                            text += `<li><a href="#${result.x},${result.y},${result.x},${result.y}" data-wkt="POINT(${result.x} ${result.y})">${lab}</a></li>`;
-                            count++;
-                        }
-                        if (count == 0 || text == '') {
-                            text = '<li>' + lizDict['externalsearch.notfound'] + '</li>';
+                    $.get(service
```
nboisteault, line 227: Could you use `fetch` instead please?

nboisteault (approved):

assets/src/modules/Search.js
rldhont: I homogenize the code and fix the issue.

@nboisteault once it is merged and backported, could you provide a refactoring ?

PR #5571, opened 2025-03-26
Popup - Improve styling of Drag&Drop designed popup

mdouchin:
The new CSS will make the D&D popup displays very similar to the editing form

Before

image

After

image

Funded by Destination Bretagne Sud Golfe du Morbihan

PR #5586, opened 2025-03-31
Popup - Do not include CSS in the GetFeatureInfo response

Gustry:
A kind of follow up of previous PR #5571

In released branch, new versions from march 2025, versions 3.8.7, 3.9 beta 3 and master, we will have this temporary flag.


mdouchin (approved):

Gustry:
Just thinking, I think I will use `CSS_FRAMEWORK`, as this flag landed in March 2025 as well in the release 3.8 branch.

So if `CSS_FRAMEWORK` is absent, it means that the CSS is required, legacy behavior.

PR #5486, opened 2025-02-28
REST API : GET for repositories & projects

neo-garaix:
* New entrypoint `api.php`
* REST API
* We can navigate through the API thanks to swagger
* Swagger is implemented inside the test stack and accessible through `http://localhost:8133`
* `CORS` in nginx conf has been updated so we can test the API from the Swagger

## Swagger 

Capture d’écran du 2025-03-21 12-54-31

### Some commands : 

```bash
curl http://localhost:8130/api.php/admin/repositories
```
```json
{
    "error code": 401,
    "error message": "Unauthorized. Basic authentication is required to access this resource."
}
```

```bash
curl http://localhost:8130/api.php/admin/repositories --user admin:admin
```
```json
[
  {
    "key": "testsrepository",
    "label": "Tests repository",
    "path": "tests/"
  },
  {
    "key": "private",
    "label": "Private repository",
    "path": "CONFIDENTIAL/"
  },
  {
    "key": "badrepository",
    "label": "Repository with bad path",
    "path": "bad/"
  },
  {
    "key": "montpellier",
    "label": "Demo",
    "path": "demoqgis/"
  },
  {
    "key": "intranet",
    "label": "Demo - Intranet",
    "path": "demoqgis_intranet/"
  }
]
```

```bash
curl http://localhost:8130/api.php/admin/repositories/intranet --user admin:admin
```
```json
{
  "key": "intranet",
  "label": "Demo - Intranet",
  "path": "demoqgis_intranet/",
  "allowUserDefinedThemes": "",
  "accessControlAllowOrigin": "",
  "rightsGroup": {
    "lizmap.tools.edition.use": [
      "admins",
      "Intranet demos group",
      "lizadmins"
    ],
    "lizmap.repositories.view": [
      "admins",
      "Intranet demos group",
      "lizadmins"
    ],
    "lizmap.tools.loginFilteredLayers.override": [
      "admins",
      "Intranet demos group",
      "lizadmins"
    ],
    "lizmap.tools.displayGetCapabilitiesLinks": [
      "admins",
      "Intranet demos group",
      "lizadmins"
    ],
    "lizmap.tools.layer.export": [
      "admins",
      "Intranet demos group",
      "lizadmins"
    ]
  }
}
```

```bash
curl http://localhost:8130/api.php/admin/repositories/intr --user admin:admin
```

```json
{
    "error code": 404,
    "error message": "Repository not found. Please provide a valid repository."
}
```

```bash
[...]

PR #5593, opened 2025-03-31
[Backport release_3_8] REST API for projects

Gustry:
Backport #5486
From @neo-garaix

### DIFF
--- a/lizmap/modules/admin_api/controllers/project_rest.classic.php
+++ b/lizmap/modules/admin_api/controllers/project_rest.classic.php
@@ -1,0 +1,112 @@
+<?php
+
+/**
+ * @author    3liz.com
+ * @copyright 2011-2025 3Liz
+ *
+ * @see      https://3liz.com
+ *
+ * @license   https://www.mozilla.org/MPL/ Mozilla Public Licence
+ */
+
+use LizmapApi\Credentials;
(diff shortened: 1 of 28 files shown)


Gustry:
@nworr Can you check my latest commit ? Which was needed to add the 7.4 support into this branch.
Thanks

lizmap/modules/admin_api/controllers/project_rest.classic.php
```diff
@@ -0,0 +1,112 @@
...
+
+        if (!Credentials::handle()) {
+            return Error::setError($rep, 401);
+        }
+
+        try {
+            $repo = lizmap::getRepository($this->param('repo'));
+
+            if ($repo == null) {
+                throw new Exception("Valid repository is needed", 404);
```
nworr, line 38: Ok for me, 

We can also use `jHttp404NotFoundException`  : https://github.com/jelix/jelix/blob/959d8a9c721fd2f8720ec91102fefeb287e79ae9/lib/jelix/core/jHttpErrorException.class.php#L100

Gustry: This change needs to be done in the master branch, in the original PR #5486 
This PR is for 3.8.

PR #5608, opened 2025-04-02
[Feature] Layer export capabilities based on attribute layers configuration

mind84:
Layer export permissions are now defined taking into account not only the repository level permissions but also the export permissions of the individual layers defined at the attribute table level.

**SPECIFICATIONS**
Following https://github.com/3liz/lizmap-plugin/pull/625, two properties have been added to the layer attribute table configuration:
- `Export enabled`
- `Export allowed groups`

image

* if `Export enabled` is not defined in configuration, every users with export rights at the repository level have export rights (backward compatibility)
* if  `Export enabled` is not checked, no users have export rights even if rights is defined at the repository level
* if  `Export enabled`  is checked and :
  -  if  `Export allowed groups` is empty every users with export rights at the repository level have export rights
  - if `Export allowed groups`  is not empty, only users with export rights at the repository level and members of groups defined in this field have exports rights

Funded by Etra


mind84:
@Gustry 
probably e2e tests fail because of https://github.com/3liz/qgis-lizmap-server-plugin/commit/1b4a14c63f8ac3c5f51bee23678e8dafa820f9ae

Can you confirm?

I that case I'll fix them

Gustry:
Yes I noticed in PR #5609, but I didn't look at it.

Did you notice what was wrong ? Help are appreciable, in another PR ? ;-)

I was indeed curious about this 3 components : Lizmap server, Lizmap Deskop and LWC, all 3 of them are implied into the maptip CSS. Thanks.

mind84:
@Gustry 

> Did you notice what was wrong ? 

Yes, the actual input fields in the popup no longer have the `field` class but are rendered as in the edition form. This breaks some tests. There are some selectors that need to be changed but it shouldn't be too complicated. A chance to rewrite some tests in POM-style :D 

> Help are appreciable, in another PR 

Sure

Gustry:
>  A chance to rewrite some tests in POM-style :D

Yes indeed, it would be nicer, but feel free to just update the selector. You might pull many things at the same time ;-)

mind84:
https://github.com/3liz/lizmap-web-client/pull/5610

tests/end2end/playwright/pages/project.js
```diff
@@ -213,6 +213,47 @@ export class ProjectPage extends BasePage {
...
+                let bboxParameter = new RegExp(String.raw`BBOX=${bbox}`, "g");
+                expect(postData).toMatch(bboxParameter);
+            }
+        };
+
+        await this.page.getByRole('button', { name: 'Export' }).click();
+        await this.page.getByRole('button', { name: 'GeoJSON' }).click();
+
+        getFeatureRequestContains(await getFeatureRequestPromise);
+    }
```
rldhont, line 255:
```suggestion
    /**
     * launchExport function
     * @param {string} layer     - the layer to export
     * @param {string} format - the format to export
     * @returns {Promise<Request>} the request generated by the export
     */
    async launchExport(layer, format){
        const getFeatureRequestPromise = this.page.waitForRequest(
            request => request.method() === 'POST'
            && request.postData() != null
            && request.postData()?.includes('GetFeature') === true
        );

        await this.page.getByRole('button', { name: 'Export' }).click();
        await this.page.getByRole('button', { name: format }).click();

        return await getFeatureRequestPromise;
    }
```
I am in favor to do test outside POM.

tests/end2end/playwright/attribute-table.spec.js
```diff
@@ -56,19 +58,137 @@ test.describe('Attribute table data restricted to map extent', () => {
...
+            json.options['limitDataToBbox'] = 'True';
+            await route.fulfill({ response, json });
+        });
+
+        const project = new ProjectPage(page, 'enable_export');
+        await project.open();
+        await project.openAttributeTable('single_wms_points');
+
+        // launche export
+        await project.launchExport('single_wms_points','3.8308870374061232%2C43.5976501168972%2C3.9342777189191174%2C43.647116478434526')
```
rldhont, line 192:
```suggestion
        // launch export
        const getFeatureRequest = await project.launchExport('single_wms_points','GeoJSON');
                
        const expectedParameters = {
            'SERVICE': 'WFS',
            'REQUEST': 'GetFeature',
            'VERSION': '1.0.0',
            'FORMAT': 'GeoJSON',
            'BBOX': /3.8308\d+,43.597650\d+,3.934277\d+,43.6471\d+/,
        }
        const getFeatureParams = await expectParametersToContain('Export GeoJSON with BBOX', getFeatureRequest.postData() ?? '', expectedParameters)
        //await expect(getFeatureParams.size).toBe(14)
```
Use `expectParametersToContain` to check parameters.

rldhont (changes requested):
Some changes in tests

tests/end2end/playwright/attribute-table.spec.js
```diff
@@ -1,6 +1,8 @@
 // @ts-check
 import { test, expect } from '@playwright/test';
 import { ProjectPage } from './pages/project';
+import { expectParametersToContain, getAuthStorageStatePath } from './globals';
+import {AdminPage} from "./pages/admin";
```
rldhont, line 5:
```suggestion
import { AdminPage } from "./pages/admin";
```
Lint

rldhont (approved):
A last lint change and it is OK for me

rldhont:
@mind84 I have to check the JS Config classes before validating your PR.

rldhont:
@mind84  Even if the config is used by *legacy* code, can you enhance `AttributeLayerConfig` class from assets/src/modules/config/AttributeTable.js.

You have to add `'export_enabled': {type: 'boolean', default: true},` in `optionalProperties` and add the getter `exportEnabled` to `AttributeLayerConfig ` class

```js

    /**
     * The layer export is enabled
     * @type {boolean}
     */
    get exportEnabled() {
        return this._export_enabled;
    }
```

mind84:
>  can you enhance AttributeLayerConfig

Sure

lizmap/modules/lizmap/lib/Project/Project.php
```diff
@@ -2169,6 +2167,50 @@ public function getUpdatedConfig()
...
 
+    /**
+     * Updates attributeLayers configuration object with layers export permissions information.
+     *
+     * @param object   $configJson Project configuration object
+     * @param string[] $userGroups User groups
+     */
+    public function setLayersExportPermissions(&$configJson, $userGroups)
+    {
+        if (property_exists($configJson, 'attributeLayers')) {
```
Gustry, line 2178: Can you switch to `getPropertyOrDefault` ? It's the purpose of this method, to avoid writing twice the key.

And all other occurrences in the code you are adding `export_enabled` and `export_allowed_groups`.

Side note, but I'm in favor of exiting a function as soon as possible, when conditions are not met. (to avoid to wonder if some code are executed below a lonnnng block of code. If the property doesn't exist, I usually return from the function. I find this code clearer and more secure.
And it also deindent your code, so easier to keep code on the left and keep 120 chars.

tests/end2end/playwright/attribute-table.spec.js
```diff
@@ -56,19 +58,147 @@ test.describe('Attribute table data restricted to map extent', () => {
...
 
-        const getFeatureRequestPromise = page.waitForRequest(request => request.method() === 'POST' && request.postData() != null && request.postData()?.includes('GetFeature') === true);
-        // bbox in getFeature request for export
-        const getFeatureRequestContains = request => {
-            const postData = request.postData();
-            expect(postData).toContain('SERVICE=WFS');
-            expect(postData).toContain('REQUEST=GetFeature');
-            expect(postData).toMatch(/BBOX=3.8258070366989356%2C43.61961486194198%2C3.897611727269447%2C43.6544002918056/);
-        };
+test.describe('Layer export permissions', () => {
```
Gustry, line 64:
```suggestion
test.describe('Layer export permissions ACL', () => {
```

tests/end2end/playwright/attribute-table.spec.js
```diff
@@ -56,19 +58,147 @@ test.describe('Attribute table data restricted to map extent', () => {
...
+            let userContext;
+            if (login !== '__anonymous') {
+                userContext = await browser.newContext({storageState: getAuthStorageStatePath(login)});
+            } else {
+                userContext = await browser.newContext();
+            }
+            const userPage = await userContext.newPage();
+
+            // go to project page
+            const project = new ProjectPage(userPage, 'enable_export');
```
Gustry, line 159:
```suggestion
            const project = new ProjectPage(userPage, 'layer_export_acl');
```

Do you mind renaming the project ? Sorry nitpicking, in the source code, the word `ACL` is usually used. So it would be easier to find projects, E2E tests with this word I think later.

Similar to https://github.com/3liz/lizmap-web-client/blob/master/tests/qgis-projects/tests/project_acl.qgs

Or looking at keywords projects in the list https://sandbox.lizmap.com/lizmap_3_8/

tests/end2end/playwright/attribute-table.spec.js
```diff
@@ -56,19 +58,147 @@ test.describe('Attribute table data restricted to map extent', () => {
...
+
+            // reset layer export permissions
+            await adminPage.modifyRepository('testsrepository');
+            await adminPage.resetLayerExportPermission();
+
+            await adminPage.page.getByRole('button', { name: 'Save' }).click();
+        })
+    })
+
+    test('Layer export request', {
```
Gustry, line 177:
```suggestion
    test('Layer export request ACL', {
```

rldhont (approved):
Thanks @mind84

Gustry (commented):
Thanks for this feature @mind84 and [Etra](https://www.etraspa.it/)

PR #5613, opened 2025-04-08
[Backport release_3_9] [Test] Refactoring e2e tests due to popup html changes

rldhont:
Backport #5610
 **Authored by:** @mind84

PR #5610, opened 2025-04-03
[Test] Refactoring e2e tests due to popup html changes

mind84:
Updates some e2e tests following  https://github.com/3liz/qgis-lizmap-server-plugin/commit/1b4a14c63f8ac3c5f51bee23678e8dafa820f9ae

- `edition-form.spec.js` and `embedded-relations.spec.js` have been refactored to be POM compliant
- `n_to_m_relations_spec.js` and `webdav-upload.spec.js` have been partially refactored. These two tests requires a lot of effort to be completely refactored so for now I've just made some changes in locators to make them work properly with the new HTML template.
- 

Funded by Faunalia

PR #5615, opened 2025-04-09
[Backport release_3_8] [Bugfix] Tile WMS layer has no loading status updated

rldhont:
Backport #5612
 **Authored by:** @rldhont


Gustry:
Thanks for fixing the tests related to D&D as well !

Kind of backport from #5610 and #5613

PR #5305, opened 2025-01-30
Metadata - Add WFS link if the user has enough rights

mdouchin:
Ticket : #

Funded by

PR #5616, opened 2025-04-10
Move copy to clipboard button to its own template, to be able to reuse it

Gustry:
Move copy to clipboard button to its own template, to be able to reuse it

Keen to have feedbacks if it's the way to go @nworr and @laurentj 

Needed in another PR #5305 from @mdouchin 

Backports will be merged after a test on an HTTPS server (to have the clipboard API enabled)


laurentj (approved):

nworr (approved):
Looks good to me

rldhont (approved):
Thanks @Gustry

Gustry:
Hum, I though it was working, as JS is available in the `<head>` using the inspector and also present in the "networks" tab.
As the docker stack is not on HTTPS, we can't really try the clipboard API.

But I didn't notice there isn't any "console.warning" displayed when clicking.

Gustry:
Trying on snap.

On local, the JS is loaded.

PR #5619, opened 2025-04-14
Review on lizmap_search : remove unused parameter bbox, add debug if needed

Gustry:
* Remove unused `bbox` parameter
* If DEBUG=True in the URL, the SQL query is logged in the admin panel, just to see and debug

### DIFF
--- a/assets/src/modules/Search.js
+++ b/assets/src/modules/Search.js
@@ -99,7 +99,6 @@ export default class Search {
                     "repository": globalThis['lizUrls'].params.repository,
                     "project": globalThis['lizUrls'].params.project,
                     "query": $('#search-query').val(),
-                    "bbox": extent.toBBOX()
                 }
                 , (results) => {
                     var text = '';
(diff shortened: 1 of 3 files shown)


assets/src/modules/Search.js
```diff
@@ -99,7 +99,6 @@ export default class Search {
                     "repository": globalThis['lizUrls'].params.repository,
                     "project": globalThis['lizUrls'].params.project,
                     "query": $('#search-query').val(),
-                    "bbox": extent.toBBOX()
                 }
```
mdouchin, line 102: Ok to remove the bbox parameter.

I think we should also go further and remove the test of the intersection between the retrieved features bbox and the current extent, line 118 & 119 of the modified file

mdouchin (changes requested):
Thanks for tackling this issue.
I think we can remove a useless client-side test

Gustry:
image

PR https://github.com/3liz/lizmap-documentation/pull/270 in the documentation about this flag

mdouchin (approved):

PR #5622, opened 2025-04-14
Move copy to clipboard button to its own template, to be able to reus…

Gustry:
Backport #5616


Gustry:
Works on HTTPS protocol

PR #5624, opened 2025-04-15
Return correct HTTP code if the lizmap_search has failed

Gustry:
Lizmap search was always return HTTP code 200 even if the query has failed.

It returns now correct HTTP code, and logging in the admin panel at the same time in case of an exception.

Follow up from previous PR #5619

### DIFF
--- a/lizmap/modules/lizmap/controllers/searchFts.classic.php
+++ b/lizmap/modules/lizmap/controllers/searchFts.classic.php
@@ -1,6 +1,7 @@
 <?php
 
 use Lizmap\Project\UnknownLizmapProjectException;
+use Lizmap\Request\Proxy;
 
 /**
  * Php proxy to access database search.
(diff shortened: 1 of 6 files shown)


lizmap/modules/lizmap/controllers/searchFts.classic.php
```diff
@@ -62,20 +69,23 @@ public function get()
         try {
             $lproj = lizmap::getProject($repository.'~'.$project);
             if ($lproj == null) {
-                jLog::log('The lizmap project '.$project.' does not exist !', 'lizmapadmin');
+                jLog::log('The lizmap project "'.$project.'" does not exist.', 'lizmapadmin');
```
nworr, line 72: if `$lproj` is null and no `UnknownLizmapProjectException ` is thrown , it means a error from php while trying to get the project 
```suggestion
                jLog::log('Error while loading "'.$project.'" project', 'lizmapadmin');
```

lizmap/modules/lizmap/controllers/searchFts.classic.php
nworr, line 1: line 85 use static operator `$fts::getData(`

instead of `$fts->getData(`

PR #5628, opened 2025-04-15
Tests - Enable the Py-QGIS-Server API

Gustry:
Tests - Enable the Py-QGIS-Server API

This Docker stack is just for testing and E2E, so no problem to enable it by default


dmarteau (approved):
LGTM

PR #5635, opened 2025-04-16
Remove BBOX and proj name from the landing page

Gustry:
Read my comment https://github.com/3liz/lizmap-web-client/pull/4032#issuecomment-2056588757 exactly one year ago, I'm keen to have feedbacks.

These extents are not human-readable and are not designed to be shown like this IMHO. (especially on a landing page, for end users)

~I'm even thinking to backport to 3.8.~

The next step is to remove totally this hover div I think, but first removing these numbers.

Linked to #3558 cc @gioman

### DIFF
--- a/lizmap/modules/view/templates/view.tpl
+++ b/lizmap/modules/view/templates/view.tpl
@@ -16,6 +16,7 @@
       <div id="liz-project-{$mi->id}-{$p->id}" class="liz-project"
         data-lizmap-repository="{$mi->id}"
         data-lizmap-project="{$p->id}"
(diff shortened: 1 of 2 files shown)


mdouchin (approved):

rldhont (approved):

rldhont (approved):

rldhont:
These lines could be removed because the values is available through `dataset`.

```js
        var proj = projElem.dataset.lizmapProj;
        var bbox = projElem.dataset.lizmapBbox;
```

PR #5647, opened 2025-04-23
JS Digitizing - enhancing module, component and the way to use them

rldhont:
* Digitizing module: Is digitizing simple geometry ?
* JS Selection tool: digitizing simple geometry
* JS Digitizing component: simplify UI if only one tool available

lizmap-action-buttons

### DIFF
--- a/assets/src/modules/Digitizing.js
+++ b/assets/src/modules/Digitizing.js
@@ -392,7 +392,7 @@ export class Digitizing {
             minidockopened: (e) => {
                 if (e.id == 'measure') {
                     this.toolSelected = this._tools[0];
-                } else if (['draw', 'selectiontool', 'print'].includes(e.id)) {
+                } else if (['draw', 'print'].includes(e.id)) {
                     // Display draw for print redlining
                     this.context = e.id === 'print' ? 'draw' : e.id;
                 }
--- a/assets/src/modules/Digitizing.js
+++ b/assets/src/modules/Digitizing.js
@@ -122,6 +122,8 @@ export class Digitizing {
         // Set draw color to value in local storage if any or default (red)
         this._drawColor = localStorage.getItem(this._repoAndProjectString + '_drawColor') || '#ff0000';
 
+        this._singlePartGeometry = false;
+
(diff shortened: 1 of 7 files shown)


assets/src/modules/Digitizing.js
```diff
@@ -115,6 +115,8 @@ export class Digitizing {
         // Set draw color to value in local storage if any or default (red)
         this._drawColor = localStorage.getItem(this._repoAndProjectString + '_drawColor') || '#ff0000';
 
+        this._simpleGeometry = false;
```
Gustry, line 118:
```suggestion
        // Single part or multipart geometry
        this._singlePartGeometry = false;
```

I think I would stick to QGIS terminology "Single parts : Multipart" :  (and other occurences in the code below)

image

What do you think ?

rldhont: Done

Gustry:
I think keeping the "second" red dot over the map after drawing the first one doesn't play nice.
It leads to some strange behaviors.

https://github.com/user-attachments/assets/f6838713-73c6-4c56-998d-4556bae1996d

(Edit, we do not see the mouse on the video, but I'm moving it always, never static)

Can you hide the second red dot ? As the mouse needs to get back to the box to click "Launch", there is so a left over red dot sticking to the border of the div...

rldhont:
@Gustry ready for review

I would like your point of view about e2e tests, thanks.

Gustry:
The comment above is still valid but we can skip.

Second try, better now ! 

image

If only a single point geom is allowed for now in "actions", then these 4 buttons are not needed, what do you think ?
Or is-it technically allowed if someone make an action with another kind of geometry ?

rldhont:
> The comment above is still valid but we can skip.
> 
> Second try, better now !
> 
> image
> 
> If only a single point geom is allowed for now in "actions", then these 4 buttons are not needed, what do you think ? Or is-it technically allowed if someone make an action with another kind of geometry ?

I have plan to continue to enhance the component to activate or not the other tools. It is in the roadmap.

PR #5665, opened 2025-04-28
Docs - Trying to improve jsdoc for lizmap-features-table

Gustry:
@mdouchin Due to `.` in events name, this was not rendered in JSDoc.

Renaming these events, it makes the doc online working well : 

image

I wonder if some alias can be added to avoid a API breaking change ?
Keen to have some feedbacks. I can add a note in CHANGELOG.md, but this will be a little notice. I wish could avoid renaming.

According to feedbacks from @vlebert #4502


mdouchin:
Good catch ! Perhaps we could trigger old and new events to keep backward compatibility (for a while) ?

rldhont:
All the new events have `.` in state module.

Gustry:
See the doc now : https://docs.3liz.org/lizmap-web-client/js/module-FeaturesTable.html about lizmap-features-table

rldhont:
I am not in favor is this event naming.

PR #5667, opened 2025-04-29
Lizmap-features-table : Add layer ID in the signal highlighted

Gustry:
@vlebert and @mdouchin 

https://github.com/3liz/lizmap-web-client/pull/4502#issuecomment-2837866857

> However I think we miss the layer ID in addition to the feature ID. Otherwise we don't know which lizmap-features-table-container did trigger the signal and we may have different behaviour for different lizmap-features-table-container

I haven't tested, just a quick addition.

### DIFF
--- a/assets/src/components/FeaturesTable.js
+++ b/assets/src/components/FeaturesTable.js
@@ -178,10 +178,12 @@ export default class FeaturesTable extends HTMLElement {
          * When the table has been successfully displayed. The event carries the lizmap-features-table HTML element ID
          * @event features.table.rendered
          * @property {string} elementId HTML element ID
+         * @property {string} itemLayerId The layer ID of the selected item
          */
         mainEventDispatcher.dispatch({
             type: 'features.table.rendered',
-            elementId: this.id
+            elementId: this.id,
+            itemLayerId: this.layerId,
         });
 
     }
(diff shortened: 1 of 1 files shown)


assets/src/components/FeaturesTable.js
```diff
@@ -292,10 +292,12 @@ export default class FeaturesTable extends HTMLElement {
              * When the user has selected an item and highlighted it
              * @event features.table.item.highlighted
              * @property {string} itemFeatureId The feature ID of the selected item
+             * @property {string} itemLayerId The layer ID of the selected item
              */
             mainEventDispatcher.dispatch({
                 type: 'features.table.item.highlighted',
                 itemFeatureId: activeItemFeatureId,
+                itemLayerId: this.layerId,
             });
```
mdouchin, line 301: I would also always give the element ID, just in case (we can then retrieve its properties, etc.)
 
```suggestion
            mainEventDispatcher.dispatch({
                type: 'features.table.item.highlighted',
                elementId: this.id,
                itemFeatureId: activeItemFeatureId,
                itemLayerId: this.layerId,
            });
```

@Gustry please also add the `layerId` and the element ID it in the event `features.table.item.dragged`

mdouchin (changes requested):
thanks, lets also add it to the other event, and also add the element ID

mdouchin (approved):

PR #5660, opened 2025-04-25
UX - Remove keywords from the gray box on the landing page

Gustry:
Linked to #5635 #3558 #4032

PR #5669, opened 2025-04-29
Landing page - Only hide the gray box for now, until we remove it soon

Gustry:
Linked to #5635 #3558 #4032

Temporary until removing this div, which is used in div. #5660


mdouchin (approved):
Merci

gioman:
@Gustry thanks!

PR #5686, opened 2025-04-29
Fix tests in print when the project has been updated

Gustry:
Small follow up from #5647

It should fix tests on 3.40 in the `print.spec.js` file.

Clearly not sure about this is the way to go. Just fixing for now.


Gustry:
Tests about print are valid on both stacks now !

Gustry:
Unrelated test failure on QGIS 3.40.

Ready for review.

I decided to remove the check on the message.

The message is disappearing automatically.
I think we want to test the message, a dedicated test about `_lizmap3.addMessage` must be done.

PR #5653, opened 2025-04-25
Fix: When using quick search menu you can't zoom back to a feature after moving the map

nboisteault:
Fix: When using quick search menu you can't zoom back to a feature after moving the map

Fix #5564

Funded by 3liz

PR #5688, opened 2025-04-30
[Backport release-3_8] Fix: When using quick search menu you can't zoom back to a feature after moving the map

Gustry:
Backport #5653


Gustry:
#5691

PR #5689, opened 2025-04-30
Docs - Trying to improve jsdoc event for lizmap-features-table without deprecation

rldhont:
Enhancing https://github.com/3liz/lizmap-web-client/pull/5677 to keep event type.

### DIFF
--- a/assets/src/components/FeaturesTable.js
+++ b/assets/src/components/FeaturesTable.js
@@ -186,17 +186,16 @@ export default class FeaturesTable extends HTMLElement {
         /**
          * When the table has been successfully displayed. The event carries the lizmap-features-table HTML element ID
          * @event featuresTableRendered
-         * @property {string} elementId HTML element ID
+         * @property {string} type        features.table.rendered
+         * @property {string} elementId   HTML element ID
          * @property {string} itemLayerId The layer ID of the selected item
-         */
-        mainEventDispatcher.dispatch({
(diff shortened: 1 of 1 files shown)


rldhont:
This PR is finished because the event name for jsDoc is not the same as the type event used to listen event.

For exemple:
```
           /**
             * When the user has dropped an item in a new position
             * @event featuresTableItemDragged
             * @property {string} type          features.table.item.dragged
             * @property {string} itemFeatureId The vector feature ID
             * @property {string} itemOldLineId The original line ID before dropping the item
             * @property {string} itemNewLineId The new line ID after dropping the item in a new position
             * @example
             * lizMap.mainEventDispatcher.addListener((lizmapEvent) => {
             *         // Log the lizmap-features-table web component HTML element ID
             *         console.log(`lizmap-features-table item dragged, ID = ${lizmapEvent.elementId}`);
             *     },
             *     ['features.table.item.dragged']
             * );
             */
```

This `@event featuresTableItemDragged` is for JSDoc, but `@property {string} type          features.table.item.dragged` is for the JS. The event type value is the value to be used to listen for.

mdouchin:
After some thinking, I also would prefer not to change the new event wording. The dots are fine for me, and it would not add API breaks (I can remember already many JS codes using the events for selection, layer tree checked, etc.)
If PR #5690 logic is OK it is better than changing the API.

PR #5671, opened 2025-04-29
API: POST repository (create & edit)

neo-garaix:
* **Creation** of repositories with a **POST** request on `/admin/repositories/{repository_key}`
   * Using a `label` _query_ parameter
   * Using a `path` _query_ parameter
* **Edition** of groups rights on repositories with a **POST** request on `/admin/right/{method}/{repository_key}`
   * Using a `method` _path_ parameter that stands for `add` or `remove`
   * Using a `group` _query_ parameter for groups IDs
   * Using a `right` _query_ parameter for rights (like `lizmap.tools.edition.use`)
* Multiple **Errors code** added
* Updated Swagger

Funded by 3Liz

PR #5720, opened 2025-05-07
Improve API REST

neo-garaix:
_Following #5671_

* **Added** the `Error 409`
* Better way to **create** a repo
* **Added** end2end tests

## groups
* **/admin/groups**
   * **GET** : Get all groups

## paths
* **/admin/paths**
   * **GET** : Get all distinct paths

## repositories
* **/admin/repositories/_{repository_key}_**
   * **POST** : Create a specific repository
      * repository_key : The repository key to create a repository
      * label : The clean name of the repository
      * path : The path of the repository
      * allowUserDefinedThemes : Indicate if user defined themes are allowed

## rights
* **/admin/rights**
   * **GET** : Get all applicable rights on Lizmap Web Client
* **/admin/rights/_{repository_key}_**
   * **GET** : Get list rights of a specific repository
      * repository_key : The repository key to get the list
   * **POST** : Add a right to a group on a repository
      * repository_key : The repository that will be edited
      * group : The group that will be edited
      * right : The right that will be added
   * **DELETE** : Remove a right to a group on a repository
      * repository_key : The repository that will be edited
      * group : The group that will be edited
      * right : The right that will be removed

Funded by 3Liz

### DIFF
--- a/lizmap/modules/admin_api/controllers/repository_rest.classic.php
+++ b/lizmap/modules/admin_api/controllers/repository_rest.classic.php
@@ -9,11 +9,11 @@
  * @license   https://www.mozilla.org/MPL/ Mozilla Public Licence
  */
 
-use Lizmap\CliHelpers\RepositoryCreator;
 use Lizmap\Request\Proxy;
 use LizmapAdmin\RepositoryRightsService;
 use LizmapApi\Credentials;
 use LizmapApi\Error;
+use LizmapApi\RepoCreator;
 use LizmapApi\RestApiCtrl;
 use LizmapApi\Utils;
 
@@ -119,34 +119,13 @@ class repository_restCtrl extends RestApiCtrl
             return Error::setError($rep, 401);
         }
 
-        $repo = $this->param('repo');
-
-        if (lizmap::getRepository($repo)) {
-            return Error::setError($rep, 400, "The repository '{$repo}' already exists.");
-        }
-
-        return $this->createRepo($rep);
-    }
-
-    /**
-     * Creates a new repository with the specified parameters.
-     *
-     * @param object $rep the response object to populate with the repository creation result
-     *
-     * @return object the updated response object containing the creation status
-     *                and repository details if successful, or an error message if failed
-     */
-    public function createRepo($rep): object
-    {
-        $repoCreator = new RepositoryCreator();
-
         $key = $this->param('repo');
         $label = $this->param('label');
         $path = $this->param('path');
         $allowUserDefinedThemes = $this->param('allowUserDefinedThemes', null);
 
         try {
-            $isCreated = $repoCreator->create($key, $label, $path, $allowUserDefinedThemes);
+            $isCreated = RepoCreator::createRepository($key, $label, $path, $allowUserDefinedThemes);
 
             $rep->data = array(
                 'key' => $key,
@@ -160,9 +139,8 @@ class repository_restCtrl extends RestApiCtrl
                 201,
                 Proxy::getHttpStatusMsg(201),
             );
-
         } catch (Exception $e) {
-            return Error::setError($rep, 400, $e->getMessage());
+            return Error::setError($rep, $e->getCode(), $e->getMessage());
         }
 
         return $rep;
--- a/lizmap/modules/admin_api/controllers/repository_rights_rest.classic.php
+++ b/lizmap/modules/admin_api/controllers/repository_rights_rest.classic.php
@@ -11,11 +11,31 @@
 
 use LizmapApi\Credentials;
 use LizmapApi\Error;
+use LizmapApi\LizmapRights;
 use LizmapApi\RestApiCtrl;
 use LizmapApi\Utils;
 
 class repository_rights_restCtrl extends RestApiCtrl
 {
+    /**
+     * Retrieves all the rights used in Lizmap Web Client.
+     *
+     * @return object the response object containing data or an error message if credentials are invalid
+     */
+    public function get(): object
+    {
+        /** @var jResponseJson $rep */
+        $rep = $this->getResponse('json');
+
+        if (!Credentials::handle()) {
+            return Error::setError($rep, 401);
+        }
+
+        $rep->data = LizmapRights::getLWCRights();
+
+        return $rep;
+    }
+
     /**
      * Handles the edition of group rules on repositories.
      *
--- a/lizmap/modules/admin_api/controllers/repository_rights_rest.classic.php
+++ b/lizmap/modules/admin_api/controllers/repository_rights_rest.classic.php
@@ -9,6 +9,7 @@
  * @license   https://www.mozilla.org/MPL/ Mozilla Public Licence
  */
 
+use LizmapAdmin\RepositoryRightsService;
 use LizmapApi\Credentials;
 use LizmapApi\Error;
 use LizmapApi\LizmapRights;
@@ -19,8 +20,9 @@ class repository_rights_restCtrl extends RestApiCtrl
 {
     /**
      * Retrieves all the rights used in Lizmap Web Client.
+     * If 'repo' is defined, it retrieves rights on a specific repository.
      *
-     * @return object the response object containing data or an error message if credentials are invalid
+     * @return object The response object containing data or an error message
      */
     public function get(): object
     {
@@ -31,15 +33,42 @@ class repository_rights_restCtrl extends RestApiCtrl
             return Error::setError($rep, 401);
         }
 
-        $rep->data = LizmapRights::getLWCRights();
+        if ($this->param('repo')) {
+            try {
+                $rep->data = $this->getRepoRights();
+            } catch (Exception $e) {
+                return Error::setError($rep, $e->getCode(), $e->getMessage());
+            }
+        } else {
+            $rep->data = LizmapRights::getLWCRights();
+        }
 
         return $rep;
     }
 
     /**
-     * Handles the edition of group rules on repositories.
+     * Return rights to a specific repository.
      *
-     * @return object a JSON response object
+     * @return array Response array containing rights on a specific repository
+     *
+     * @throws Exception
+     */
+    protected function getRepoRights(): array
+    {
+        $repo = lizmap::getRepository($this->param('repo'));
+
+        if ($repo == null) {
+            throw new Exception("The repository doesn't exist !", 404);
+        }
+
+        return RepositoryRightsService::getRights($repo->getKey());
+    }
+
+    /**
+     * Adds a specific right for a group to a repository.
+     *
+     * @return object the response object containing the details of the operation,
+     *                or an error message if the operation fails
      */
     public function post(): object
     {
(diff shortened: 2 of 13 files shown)


tests/end2end/playwright/globals.js
```diff
@@ -334,12 +334,46 @@ export async function expectParametersToContain(title, parameters, expectedParam
...
+ * @param {import("playwright-core/types/types.js").APIRequestContext} request Request to use
+ * @param {string} url URL to do a POST request on
+ * @param {object} data parameters for the request
+ * @returns {Promise<import("playwright-core/types/types.js").APIResponse>} Response
+ */
+export async function requestPOSTWithAdminBasicAuth(request, url, data) {
+    return await request.post(url,
+        {
+            headers: {
+                authorization: 'Basic YWRtaW46YWRtaW4=' // admin:admin
```
Gustry, line 357: Can you factorize this variable, and use ` btoa("admin:admin")` ?

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -0,0 +1,126 @@
...
+            throw new \Exception('The repository path is not set.', 400);
+        }
+
+        $data = array(
+            'label' => $label,
+            'path' => $path,
+            'allowUserDefinedThemes' => $allowUserDefinedThemes,
+        );
+
+        $rootRepositories = \lizmap::getServices()->getRootRepositories();
```
Gustry, line 39: There is a setting in LWC admin panel to choose or not if we have a root folder.

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -0,0 +1,126 @@
...
+
+        // Testing a relative path and updating it if needed
+        try {
+            $path = self::testRelativePath($path, $rootRepositories);
+        } catch (\Exception $e) {
+            throw new \Exception($e->getMessage(), $e->getCode());
+        }
+
+        if (!file_exists($path) or !is_dir($path)) {
```
Gustry, line 48: I think it's the purpose of this request, to create the folder as well.

Maybe add a flag in the HTTP request, like `createDirectory` ?
If `createDirectory` is False, then indeed this check is valid.

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -0,0 +1,126 @@
...
+     *
+     * @param string $path     The path to validate and resolve. It can be a relative or absolute path.
+     * @param string $rootRepo the root repository directory used as a base for paths
+     *
+     * @return string returns the path
+     *
+     * @throws \Exception if the provided relative path is not authorized or cannot be resolved within
+     *                    the root repository
+     */
+    private static function testRelativePath(string $path, string $rootRepo): string
```
Gustry, line 82: Bonus if you can add some PHP unit on this function ?

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * Validates whether the given value represents a boolean true value.
+     *
+     * @param mixed $value The value to validate. Can be of any data type.
+     *
+     * @return bool returns true if the value represents a boolean true value, otherwise false
+     */
+    private static function isValidBooleanValue(mixed $value): bool
```
Gustry, line 111: Same, if you can add some PHP unit on this function ?

Gustry (commented):
First look.

Did you look at comment from @rldhont and me on #5671 ?

Maybe indeed we can restrict the login from the web interface, but a clear comment about that ? See `tests/end2end/playwright/requests-metadata.spec.js`, it is either `connected from context` or `basic auth`.

lizmap/modules/admin_api/lib/RepoCreator.php
rldhont: If the API mimic the User Interface, it cannot create Directory yet.

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -0,0 +1,126 @@
...
+     * @param string $allowUserDefinedThemes whether user-defined themes are allowed for the repository
+     *
+     * @return bool returns true if the repository is successfully created
+     *
+     * @throws \Exception if the repository creation fails
+     */
+    public static function createRepository(string $key, string $label, string $path, string $allowUserDefinedThemes): bool
+    {
+        if (\lizmap::getRepository($key)) {
+            throw new \Exception("The repository '{$key}' already exists.", 409);
```
rldhont, line 22: I invite you to defined an `ApiException` that extends `\Exception`, just like `UnknownLizmapProjectException`:

```php
<?php

namespace LizmapApi;

class ApiException extends \Exception {}

```

Then when you catch it you know that the code from `getCode()` is an HTTP code.

lizmap/modules/admin_api/controllers/repository_rights_rest.classic.php
```diff
@@ -9,76 +9,144 @@
...
         $right = $this->param('right');
 
-        $isValid = Utils::verifyVars($group, $right, $key);
-
-        if (!$isValid['bool']) {
-            return Error::setError($rep, 400, $isValid['message']);
+        try {
+            Utils::verifyVars($group, $right, $key);
+        } catch (Exception $e) {
+            return Error::setError($rep, $e->getCode(), $e->getMessage());
```
rldhont, line 132:
```suggestion
        try {
            Utils::verifyVars($group, $right, $key);
        } catch (ApiException $e) {
            return Error::setError($rep, $e->getCode(), $e->getMessage());
        } catch (Exception $e) {
            return Error::setError($rep, 500, $e->getMessage());
```

Once `LizmapApi\ApiException` is defined

rldhont (changes requested):

tests/api/openapi.yaml
```diff
@@ -566,6 +570,12 @@ paths:
           required: true
           schema:
             type: string
+        - name: createDirectory
+          in: query
+          description: Indicate if we want to create a directory depending in the path given
```
Gustry, line 575:
```suggestion
          description: Indicate if we want to create the directory on the file system
```

Just to be clear about the directory on file system, versus a "Lizmap repository".

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -41,43 +47,63 @@ public static function createRepository(string $key, string $label, string $path
...
-                400
+        if (!is_dir($path)) {
+            if ($createDirectory) {
+                $oldUmask = umask(0002);
+                $isCreated = mkdir($path, 0775, true);
+                umask($oldUmask);
+                if (!$isCreated) {
+                    throw new ApiException(
+                        "Unable to create directory '{$path}'".
+                        " Maybe you don't have enough permissions.",
```
Gustry, line 64:
```suggestion
                        " Maybe the process running Lizmap does not have enough permissions.",
```
I think it's not the connected user, but the process, usually `www-data` ?

And maybe a message should be send to jlog `error` : "The user "foo" tried to create the directory "/bar/..." but got the error ...." ?

Every-time an `Exception` is catch because this error were not expected, nice to keep some log about it.

lizmap/modules/admin_api/lib/RepoCreator.php
neo-garaix: I will check this !

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        $data = array(
+            'label' => $label,
+            'path' => $path,
+            'allowUserDefinedThemes' => $allowUserDefinedThemes,
+        );
+
+        $rootRepositories = \lizmap::getServices()->getRootRepositories();
+
```
Gustry, line 46: an important check must be added : 

if not $rootRepositories and $createDirectory then
     throw new ApiException('Impossible to create a directory if the Lizmap setting "root repositories" is not defined');


Indeed, the create directory on the filesystem must be allowed only and only if root repositories is defined.
We do not want users to create a folder in random folder /home/foo/bar for instance for security reason.

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -0,0 +1,122 @@
...
+        // Testing a relative path and updating it if needed
+        try {
+            $path = self::testRelativePath($path, $rootRepositories);
+        } catch (ApiException $e) {
+            throw new ApiException($e->getMessage(), $e->getCode());
+        }
+
+        if (!is_dir($path)) {
+            if ($createDirectory) {
```
Gustry, line 55: Maybe check also that $path is a **single** directory name, (no `../`, not `/...` ...). It must be only possible to create a folder on the OS **within** the $rootRepositories, linked to my comment above.

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -0,0 +1,122 @@
...
+
+class RepoCreator
+{
+    /**
+     * Creates a new repository with the given properties.
+     *
+     * @param string $key                    The unique identifier for the repository. Must not yet exist.
+     * @param string $label                  the display name of the repository
+     * @param string $path                   the path to the repository directory
+     * @param string $allowUserDefinedThemes whether user-defined themes are allowed for the repository
```
rldhont, line 13:
```suggestion
     * @param bool $allowUserDefinedThemes whether user-defined themes are allowed for the repository
```
It has to be a boolean, isn't it

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * @return bool returns true if the repository is successfully created
+     *
+     * @throws ApiException if the repository creation fails
+     */
+    public static function createRepository(
+        string $key,
+        string $label,
+        string $path,
+        string $allowUserDefinedThemes,
```
rldhont, line 24:
```suggestion
        bool $allowUserDefinedThemes,
```

lizmap/modules/admin_api/controllers/repository_rest.classic.php
```diff
@@ -113,56 +118,41 @@ protected function getRepoDetail($rep)
...
-     *                and repository details if successful, or an error message if failed
-     */
-    public function createRepo($rep): object
-    {
-        $repoCreator = new RepositoryCreator();
-
         $key = $this->param('repo');
         $label = $this->param('label');
         $path = $this->param('path');
         $allowUserDefinedThemes = $this->param('allowUserDefinedThemes', null);
```
rldhont, line 131:
```suggestion
        $allowUserDefinedThemes = Utils::isValidBooleanValue($this->param('allowUserDefinedThemes', null));
```
Check it is a boolean here

rldhont (changes requested):
Check `$allowUserDefinedThemes` is a boolean at the request level

lizmap/modules/admin_api/lib/RepoCreator.php
neo-garaix: We should unauthorize creation of directory through absolute path  because if rootRepo is set, we still could create some. It would not be fine.

lizmap/modules/admin_api/lib/RepoCreator.php
neo-garaix: Of course, I was thinking of it while reading your comment above.

neo-garaix: I think I will revamp the pathValidator method

lizmap/modules/admin_api/controllers/repository_rest.classic.php
```diff
@@ -113,56 +118,41 @@ protected function getRepoDetail($rep)
...
-    public function createRepo($rep): object
-    {
-        $repoCreator = new RepositoryCreator();
-
         $key = $this->param('repo');
         $label = $this->param('label');
         $path = $this->param('path');
-        $allowUserDefinedThemes = $this->param('allowUserDefinedThemes', null);
+        $allowUserDefinedThemes = Utils::isValidBooleanValue($this->param('allowUserDefinedThemes', false));
+        $createDirectory = Utils::isValidBooleanValue($this->param('createDirectory'));
```
Gustry, line 132: Not sure if it's accessible, but there is an helper `boolParam` https://jelix.org/reference/1.8/classes/jController.html#method_boolParam 

Just for your information. Fine as well.

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -0,0 +1,155 @@
...
+        $data = array(
+            'label' => $label,
+            'path' => $path,
+            'allowUserDefinedThemes' => $allowUserDefinedThemes,
+        );
+
+        $rootRepositories = \lizmap::getServices()->getRootRepositories();
+
+        if ($rootRepositories == '' and $createDirectory) {
+            throw new ApiException('The root repository is not set, and you want to create a directory.', 403);
```
Gustry, line 48:
```suggestion
            throw new ApiException('The root repository folder is not set on this instance, it is not possible to create a folder on the file system without this setting.', 403);
```
Just rephrasing ?

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        try {
+            $path = self::pathValidator($path, $rootRepositories);
+        } catch (ApiException $e) {
```
Gustry, line 53: I'm not sure, but you don't need to catch and resend as you are treating the same kind of exception `ApiException`. Up to you.

lizmap/modules/admin_api/lib/RepoCreator.php
```diff
@@ -0,0 +1,155 @@
...
+     * @param string $rootRepo the root repository directory used as a base for paths
+     *
+     * @return string Returns the path
+     *
+     * @throws ApiException If the path is incorrect
+     */
+    public static function pathValidator(string $path, string $rootRepo): string
+    {
+        if ($rootRepo == '') {  // createDirectory = false
+            if (!str_starts_with($path, '/')) {
```
Gustry, line 103: I guess PHP as a file utile module to check if it's root directory ? 

Some are running LWC on a Windows server : https://github.com/NaturalGIS/qgis_server_and_lizmap_on_windows

So the path might be `C:/` ?
Even if Windows server is not officially supported, this kind of string comparaison looks fragile to me. 

Just to keep note, we can check.

Gustry: Shouldn't we just check that $path match a regexp `\w` or `[a-zA-Z0-9_]` ?

lizmap/modules/admin_api/controllers/project_rest.classic.php
```diff
@@ -90,8 +95,10 @@ protected function getProjDetail($rep, $repo)
         try {
             $proj = $repo->getProject($this->param('proj'));
             $projInfos = $proj->getFirstQgisConfigLine();
-        } catch (Throwable $e) {
-            return Error::setError($rep, 404, $e->getMessage());
+        } catch (Exception $e) {
+            jLog::logEx($e, 'error');
+
+            return Error::setError($rep, $e->getCode(), $e->getMessage());
```
rldhont, line 101: Are you sure `getFirstQgisConfigLine()` will throw an `ApiException` or an `Exception` that provide an HTTP error code ?

lizmap/modules/admin_api/lib/RepoCreator.php
neo-garaix: Yeah, I replaced by some regexp

lizmap/modules/admin_api/controllers/project_rest.classic.php
neo-garaix: I think I forgot to change the `$e->getCode()` part. The code concerning XML reader can throw an `Exception` but not necessarily an HTTP one. I replaced it in https://github.com/3liz/lizmap-web-client/pull/5720/commits/a2b564f784415d33b1ac0b2030be2c61e6312d13.

rldhont (approved):
LGTM!

PR #5741, opened 2025-05-14
Metadata - Add WFS link if the user has enough rights

Gustry:
Backport #5305
Author @mdouchin


Gustry:
Waiting for the commit from #5742

PR #5743, opened 2025-05-15
Fix JS Attribte Table: format big data

rldhont:
Fixing regression in #5682
Adding end2end test

Funded by Terre de Provence Agglomération, Avignon, DVP Solar


Gustry:
Also fixing paid support 3489 and 3492

PR #5748, opened 2025-05-15
Fix identifiying NEU projection when root WMS Layer has no bounding boxes

rldhont:
With QGIS 3.40, the root WMS Layer has more easily no bounding boxes. To check if the project porjection is NEU and not ENU, Lizmap can use a child which could be a layer or a group.


assets/src/modules/Lizmap.js
```diff
@@ -76,10 +76,20 @@ export default class Lizmap {
                 // Add CRS:84 projection, same as EPSG:4326 but with ENU axis orientation
                 proj4.defs("CRS:84","+proj=longlat +datum=WGS84 +no_defs +type=crs");
                 register(proj4);
+                let wmsLayer = wmsCapabilities.Capability.Layer;
+                while (wmsLayer.BoundingBox === undefined) {
```
Gustry, line 80: I think I would add your PR description within the code as JS comment here. It might be easier to understand later (or better comment)

rldhont: Thanks @Gustry

PR #5749, opened 2025-05-15
JS initialize Lizmap Application from not legacy code

rldhont:
The Lizmap initialization was perform by the legacy code, but it now depends the mainLizmap singleton.
So to be sure that the mainLizmap is ready, it is more safe to perform Lizmap initialization after the mainLizmap singleton is ready.


Gustry:
LGTM according to the 3.40 test stack.

PR #5768, opened 2025-05-22
Tooltip: display layer style for points

nboisteault:
Also increase radius on hover

This option has to be checked in the Lizmap plugin.

Funded by CD30 Gard


rldhont:
@mdouchin would it need some configuration ?

mdouchin:
Yes, a way to setup a new property for each tooltip layer

```json
"displayLayerStyle": true
```

mdouchin:
@rldhont I just pushed back after rebasing from master. It seems ready to be merged, if you are OK.

PR #5770, opened 2025-05-23
Optimize fts query to use the index

nboisteault:
Funded by DVP SOLAR France paid support


rldhont (approved):
LGTM but I'd like to wait for @mdouchin review before merging.

PR #5819, opened 2025-06-13
JS: Requesting GetLegendGraphics after the map starting to be rendered

rldhont:
* Use multiple GetLegendGraphic request with single layer instead of one big POST request
* Add a Map State isReady setter
* Better load status changes for tile layers
* Reorganise the map constructor by moving some map changes at the end of the construction like updating map state properties and setting map state is ready
* Requesting Legend and Features after the map started to load

Funded by SMICA https://www.smica.fr/, Klein und Leber Gbr https://www.gisgeometer.de/

### DIFF
--- a/assets/src/index.js
+++ b/assets/src/index.js
@@ -54,6 +54,28 @@ const waitFor = async function waitFor(maxWait, sleepStep, f){
     return waitingTime;
 };
 
+const definedCustomElements = () => {
+    window.customElements.define('lizmap-geolocation', Geolocation);
+    window.customElements.define('lizmap-geolocation-survey', GeolocationSurvey);
+    window.customElements.define('lizmap-features-table', FeaturesTable);
+    window.customElements.define('lizmap-selection-tool', SelectionTool);
+    window.customElements.define('lizmap-selection-invert', SelectionInvert);
+    window.customElements.define('lizmap-snapping', Snapping);
+    window.customElements.define('lizmap-scaleline', Scaleline);
+    window.customElements.define('lizmap-mouse-position', MousePosition);
+    window.customElements.define('lizmap-digitizing', Digitizing);
+    window.customElements.define('lizmap-overviewmap', OverviewMap);
+    window.customElements.define('lizmap-feature-toolbar', FeatureToolbar);
+    window.customElements.define('lizmap-reverse-geom', ReverseGeom);
+    window.customElements.define('lizmap-paste-geom', PasteGeom);
+    window.customElements.define('lizmap-action-selector', ActionSelector);
+    window.customElements.define('lizmap-print', Print);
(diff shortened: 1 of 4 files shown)


nboisteault:
@rldhont I think this patch is way too big to be merged in release_3_8

rldhont:
It is a funded fix for release 3.8.11

Gustry:
> It is a funded fix for release 3.8.11

I totally understand your points, both @nboisteault and @rldhont.

But @rldhont , this shouldn't be a valid reason in the middle term.

* 3.8.0 has been released 10 months ago.
* 3.9.0 is around the corner, so it means that 3.8 is "nearly" end of life, because there is a 3 month overlap.
* In QGIS, **and** other softwares, some bugs, known in a released branch are only fixed in a newer version, the bug will remain in a older branch

The more we move to a final EOL status, the more only critical bugfix should be merged (or isolated features).

Just for information.

rldhont:
@nboisteault @Gustry @mdouchin all is OK now

Gustry (commented):
Tested locally, it works.
At the beginning, a little bit surprising this new behavior.

Can you check newly added `dispatch()` call to have appropriate JSDoc about new signals ?

gioman:
> a little bit surprising this new behavior

@Gustry how so?

Gustry:
> @Gustry how so?

All the legend is loading when the map is ready. So the legend area looks empty at the beginning.

I would be nice for the UX to have a small waiting GIF in the box before it shows the **root** of the legend.

@rldhont If you rebase, the 3.34 stack has been fixed on master.

rldhont:
Thanks @Gustry , still some errors in Cypress...

rldhont:
@Gustry I have fixed the new behavior you seen

Gustry:
It's indeed better.
Thanks.
Merging it

meyerlor:
woah, this 3.9 release is pure gold - dunno where to post it then here, but loading times are stunning. really good job done here and with all those other things which came with the release!

while measuring the loading times with the dev tools of firefox one thing catched my eye: Lizmap is sending a lot of GET requests for legends of layers where the legend is completely disabled via the qgis plugin (it does not affect the map rendering time, just a lot of requests are popping up in the dev tools). they look like:

https://test.gisgeometer.de/index.php/lizmap/service?repository=test&project=GUA&SERVICE=WMS&REQUEST=GetLegendGraphic&VERSION=1.3.0&FORMAT=application%2Fjson&LAYER=AS_Ahausen_Muehlenweg_PLAN&STYLES=default

just wanted to report it, not sure if there is any action required!

Gustry:
Thanks for your feedback. It has been forwarded. 😉

You can create a ticket if needed.
Do you mean this option ?
image

Better to discuss in a ticket.

PR #5845, opened 2025-06-19
Change Login event in log when QGIS performs authentification 

nworr:
An altenative to #5827 

When QGIS perform authentification, the `user-agent` is detected and the log event is changed to `qgis-login`

Due to some misspelled `lizmapLogConfig.ini.php` in old versions, i choose to not update the file using a upgrade process, thus i add a method to build default values

The behaviour of the  log page is now to hide qgis-login event, i add option to show them

Screenshot 2025-06-19 at 10-25-28 Administration

Funded by 3liz

### DIFF
--- a/lizmap/modules/admin/locales/en_US/admin.UTF-8.properties
+++ b/lizmap/modules/admin/locales/en_US/admin.UTF-8.properties
@@ -222,8 +222,8 @@ logs.export.button=Export
 logs.empty.button=Empty
 logs.empty.confirm=Are you sure to want to empty the log content?
 logs.empty.ok=The log table %s has been successfully emptied.
-log.qgis_login.hide=Hide QGIS Login events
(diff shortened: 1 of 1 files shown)


lizmap/modules/admin/locales/en_US/admin.UTF-8.properties
```diff
@@ -222,6 +222,8 @@ logs.export.button=Export
 logs.empty.button=Empty
 logs.empty.confirm=Are you sure to want to empty the log content?
 logs.empty.ok=The log table %s has been successfully emptied.
+log.qgis_login.hide=Hide QGIS Login events
+log.qgis_login.show=Show QGIS Login events
```
Gustry, line 226:
```suggestion
log.qgis_login.hide=Hide QGIS login events
log.qgis_login.show=Show QGIS login events
```

PR #5858, opened 2025-06-27
Fix media controllers to intensively using ETAG

rldhont:
* Fix ETAG for media files
* Fix etag for illustration files
* Fix ETAG for default illustration file
* Enhancing media controller tests


mdouchin (approved):
Not tested, but reading the code seems to correspond to the need
Thanks

lizmap/modules/view/controllers/media.classic.php
```diff
@@ -287,11 +287,34 @@ public function getMedia()
...
+                $etag = sha1($etag);
+            }
+
+            if ($etag !== '' && $rep->isValidCache(null, $etag)) {
+                return $rep;
+            }
+
+            if ($etag !== '') {
+                $this->setEtagCacheHeaders($rep, $etag);
+            }
```
laurentj, line 317: I think we could simplify code like this:

```suggestion
        } else  if ($this->canBeCached()) {
            // Etag header and cache control
            $etag = 'media';
            $etag .= '-'.$lrep->getKey().'~'.$lproj->getKey();
            $etag .= '-'.$path;
            $etag .= '-'.filemtime($finalPath);
            $etag = sha1($etag);

            if ($rep->isValidCache(null, $etag)) {
                return $rep;
            }

            $this->setEtagCacheHeaders($rep, $etag);
```

lizmap/modules/view/controllers/media.classic.php
```diff
@@ -287,11 +287,34 @@ public function getMedia()
             $rep->fileName = '';
             $rep->content = $content;
         }
+        // For HEAD request, we need to set the content length
+        $fileSize = filesize($finalPath);
+        $rep->addHttpHeader('Content-Length', $fileSize);
 
-        $rep->setExpires('+1 days');
+        // Set Allow header to GET and HEAD for all files
+        $rep->addHttpHeader('Allow', 'GET, HEAD');
```
laurentj, line 295: This header is only useful when you return an error 405 Method Not Allowed, after checking the HTTP method.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Allow

```
if ($_SERVER['REQUEST_METHOD'] != 'GET' && $_SERVER['REQUEST_METHOD'] != 'HEAD') {
    $rep = $this->getResponse(....);
    $rep->setHttpStatus(405, 'Method not Allowed');
    $rep->addHttpHeader('Allow', 'GET, HEAD');
    return $rep;
}
```

rldhont:
Thanks @laurentj I have fixed the code.

PR #5860, opened 2025-06-27
Tests - Update PHP-Unit version

Gustry:
Tests - Update PHP-Unit version


rldhont:
Fixed with https://github.com/3liz/lizmap-web-client/pull/6035

Issue #5728, opened 2025-05-09
[Bug]: ERROR:  relation "lizmap_search" does not exist


hoge6b01:
### What is the bug? (in English)

Since installing Lizmap Web Client 3.8.8 I've got the following ERROR log entry:

pg_query(): Query failed: ERROR:  relation "lizmap_search" does not exist
LINE 1: SELECT * FROM lizmap_search LIMIT 0;
                      ^	/var/www/lizmap-web-client/lizmap/vendor/jelix/jelix/lib/jelix/plugins/db/pgsql/pgsql.dbconnection.php	211

### Steps to reproduce the issue

Entry in the error.log

### Versions, safeguards, check summary etc

3.8.8

### Check Lizmap plugin

- [x] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 24.04.2

### Browsers

Firefox

### Browsers version

138.0.1 

### Relevant log output

```shell
pg_query(): Query failed: ERROR:  relation "lizmap_search" does not exist
LINE 1: SELECT * FROM lizmap_search LIMIT 0;
                      ^	/var/www/lizmap-web-client/lizmap/vendor/jelix/jelix/lib/jelix/plugins/db/pgsql/pgsql.dbconnection.php	211
```

katagen:
I have the same error, but it appears with the update 3.8.7

hoge6b01:
Might be. I updated from 3.8.6 to 3.8.8.

Gustry:
It might be linked to my two PR : #5619 #5624 and their backports in 3.8.8

Gustry:
Hum, these 2 PR doesn't to be linked to file `lizmap/modules/lizmap/classes/lizmapFtsSearch.listener.php`, from where the SQL query is executed.

Looks like to #2008 #2034

wojtek19022:
I encountered similar problem in latest 3.7 version. It appeared while i started using pg_service.conf to connect with database and diaplay its' layers.

Everything worked fine until i started exploring relations that i made (1:n and n:m) then it stopped working.

From logs I got information that lizmap doesn't have connection to display data from a relation... but it displayed these layers. So where is actually the problem for lizmap?? 

Everything was properly working on a QGIS desktop 3.34.15. I also user PostgreSQL 15 database.

Issue #1285, opened 2019-07-17
Annoying warning "...no such table: lizmap_search..."


nboisteault:
### What is the bug?
When loading LWC errors.log shows : 
`2019-07-17 10:07:27	10.0.2.2	warning	2019-07-17 10:07:27	[2]	SQLite3::query(): Unable to prepare statement: 1, no such table: lizmap_search	/lizmap_web_client/lib/jelix/plugins/db/sqlite3/sqlite3.dbconnection.php	112
`

### Steps to reproduce the issue

1. Load LWC
2. Look last line ine /var/log/errors.log
3. There is a warning like above

### Environment

 - Lizmap version: 3.2 and master
 - Lizmap admin information (users, logs…) stored into Sqlite (default)

dassau:
Hi, we have the same warning here, with the result that projects do not load at all. Is there a workaround available, that we can use?

infoble:
Idem for me with lizmap 3.3. This error does not seem to affect the application. Me my only problem is the lack of qgis server log but I do not think there is a link.

nboisteault:
@mdouchin @laurentj  We should replace this [query](https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/classes/lizmapFtsSearch.listener.php#L45) with this [one ('Slow and sure' part)](https://dba.stackexchange.com/a/86098/126216). But it only works with PostgreSQL. Should this search functionality work with SQLite ?

laurentj:
There is already all  we need into Jelix to see if the table exists, with jDbSchema. It does the query you indicate for pgsql, and similar queries for other type of databases.

I have a patch. But it annoys me, because doing such query at each requests is stupid.

I don't understand why this "search engine" is always enabled. It should be deactivated by default, and we should have a checkbox into the configuration panel to enable/disable it (so we should have a configuration parameter into lizmapConfig.ini.php). And of course, when the administrator wants to activate it, the table should be checked, and if it not there, it should be kept disabled (or we could disable the checkbox).

laurentj:
We should have also :

* an update script that adds this new parameter, with `true` if the table already exists
* the installer script should create this table if the parameter is set to true, or, in the configuration panel, it should create the table if the checkbox is checked.

nboisteault:
@laurentj @rldhont @mdouchin do we postpone the fix?

rldhont:
link to https://github.com/3liz/lizmap-web-client/issues/1968

[...]

PR #5873, opened 2025-06-30
Fix the way to check if the lizmap_search is available

rldhont:
Instead of doing a `SELECT * FROM lizmap_search LIMIT 0;` to check if `lizmap_search` table / view / materialized view exists in the `search_path`


Gustry:
#1285 and #5728

lizmap/modules/lizmap/classes/lizmapFtsSearch.listener.php
```diff
@@ -45,17 +45,31 @@ protected function checkLizmapFts()
...
+        ";
 
-        // Try to get data from lizmap_fts table / view / materialized view
         try {
-            $cnx->query('SELECT * FROM lizmap_search LIMIT 0;');
-            $cnx->commit();
-            $ok = true;
+            $res = $cnx->query($sql);
+            foreach ($res as $r) {
+                $ok = $r->lizmap_search_exists;
```
Gustry, line 69:
```suggestion
                return $r->lizmap_search_exists;
```

As we are in a foreach, it makes a little bit cleaner that we return ASAP, and doing nothing with the `$ok`

Gustry (approved):
Thanks !

_On my side_, I would not backport to 3.8, or with a delay like QGIS.

PR #5887, opened 2025-07-01
Fix ValueRelation EditWidget parsing

rldhont:
The ValueRelation EditWidget was not well parsed and default value not well keeped.

Other unit tests added

### DIFF
--- a/tests/units/classes/Project/Qgis/VectorLayerEditWidgetTest.php
+++ b/tests/units/classes/Project/Qgis/VectorLayerEditWidgetTest.php
@@ -807,4 +807,30 @@ class VectorLayerEditWidgetTest extends TestCase
             $this->assertSame($value, $editWidget->config->PropertyCollection[$prop], $prop);
         }
     }
+
+    public function testUniqueValuesFromXmlReader(): void
+    {
+        $xmlStr = '
+        <editWidget type="UniqueValues">
+          <config>
+            <Option type="Map">
+              <Option value="1" type="QString" name="Editable"/>
+            </Option>
+          </config>
+        </editWidget>
+        ';
+        $oXml = App\XmlTools::xmlReaderFromString($xmlStr);
+        $editWidget = VectorLayerEditWidget::fromXmlReader($oXml);
+
+        $this->assertEquals('UniqueValues', $editWidget->type);
+        $this->assertNotNull($editWidget->config);
+        $this->assertInstanceOf(EditWidget\UniqueValuesConfig::class, $editWidget->config);
+
+        $config = array(
+            'Editable' => true,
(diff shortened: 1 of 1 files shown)


mdouchin (approved):

lizmap/modules/lizmap/lib/Project/Qgis/Layer/EditWidget/ValueRelationConfig.php
```diff
@@ -79,6 +90,10 @@ protected function set(array $data): void
         if (array_key_exists('UseCompleter', $data)) {
             $data['UseCompleter'] = filter_var($data['UseCompleter'], FILTER_VALIDATE_BOOLEAN, FILTER_NULL_ON_FAILURE);
         }
+
```
nboisteault, line 93: Could it be removed?

nboisteault (approved):

nboisteault:
@rldhont I've tested and it works with my test project.
What happened? Did the XML change in .qgs files?
Old projects where not having bug so I guess the bug is only visible for new layers in QGIS 3.40?

rldhont:
> @rldhont I've tested and it works with my test project. What happened? Did the XML change in .qgs files? Old projects where not having bug so I guess the bug is only visible for new layers in QGIS 3.40?

Yes the XML has changed

lizmap/modules/lizmap/lib/Project/Qgis/Layer/EditWidget/ValueRelationConfig.php
rldhont: Yes, it was late yesterday in the train.

tests/units/classes/Project/Qgis/VectorLayerEditWidgetTest.php
```diff
@@ -300,4 +301,325 @@ public function testRangeFromXmlReader(): void
...
+            'NofColumns' => 1,
+            'OrderByValue' => true,
+            'UseCompleter' => false,
+            'Value' => 'es_nom_commun',
+        );
+        foreach ($config as $prop => $value) {
+            $this->assertSame($value, $editWidget->config->{$prop}, $prop);
+        }
+
+        // Old Value relation
```
Gustry, line 352: What is "Old" ?
Can you add a string like "before QGIS 3.40".

I would be nice to have it in other places.

This is for, in 2 or 3 years, to know if this code can be safely removed because QGIS 3.34 will not be supported and we can do some cleaning into the code about "OLD" tests.

Thanks

nboisteault:
@rldhont should not it be backported to 3.8?

Issue #3758, opened 2023-07-02
Add option *not* show NULL values in a locating panel dropdown


gioman:
To avoid:

![Screenshot 2023-06-11 134044](https://github.com/3liz/lizmap-web-client/assets/1951107/f856e29d-53c4-42a5-a742-11c7f4ec2638)

There have been some question/discussion about the fact that the default identify dialog does not shows NULL attributes, so I think that this also should apply (maybe as option) to the list/dropdown generated for a locating panel/dropdown as is not really uncommon to want to locate by an attribute that may have some NULL values.

PR #5898, opened 2025-07-03
Fix JS: LocateByLayer does not support null value because of DOMPurify.sanitize

rldhont:
`DOMPurify.sanitize` needs a string and `null` value does not have a `toString` method.

Funded by Cadageo

### DIFF
--- a/assets/src/modules/LocateByLayer.js
+++ b/assets/src/modules/LocateByLayer.js
@@ -291,10 +291,15 @@ export default class LocateByLayer {
             for (var j=0, featuresLen=features.length; j<featuresLen; j++) {
                 var featElement = features[j];
                 locate.features[featElement.id.toString()] = featElement;
-                if ( !('filterFieldName' in locate) )
+                if ( !('filterFieldName' in locate) ) {
+                    // add option
                     options += '<option value="' + featElement.id + '">'
-                        + DOMPurify.sanitize(featElement.properties[locate.fieldName].toString())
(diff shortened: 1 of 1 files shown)


assets/src/modules/LocateByLayer.js
```diff
@@ -291,10 +291,15 @@ export default class LocateByLayer {
...
-                if ( !('filterFieldName' in locate) )
+                if ( !('filterFieldName' in locate) ) {
+                    // add option
                     options += '<option value="' + featElement.id + '">'
-                        + DOMPurify.sanitize(featElement.properties[locate.fieldName].toString())
-                        + '</option>';
+                    // DOMPurify.sanitize needs a string
+                    options += featElement.properties[locate.fieldName] !== null ?
+                        DOMPurify.sanitize(featElement.properties[locate.fieldName].toString()) :
+                        'NULL'; // Default label for null value
```
rldhont, line 300: @nboisteault @Gustry @mdouchin : Empty string or `NULL` when the value is `null`.

josemvm:
maybe the same for the filter tool (?)

assets/src/modules/LocateByLayer.js
rldhont: To be iso as before DOMPurify, we have to display `null`. I will change it.

For an other version, we could add an option to do not add `null` to the list with some log. https://github.com/3liz/lizmap-web-client/issues/3758

PR #5908, opened 2025-07-04
Editing - Fix getting wrong feature when layer has a SQL filter

mdouchin:
When the PostgreSQL layer in QGIS has a datasource filter containing a OR, Lizmap does not get the correct feature when opening the editing form from the popup.

Adding parenthesis around the datasource SQL attribute fixes this issue

Funded by 3liz


mdouchin:
Il will add a PHP unit test and fix the current unit test (no need to add parenthesis in the context of multiple primary keys...)

mdouchin:
Unit test added

nboisteault (approved):

PR #5917, opened 2025-07-06
PHP: Well deprecate lizmapRepository class

rldhont:
* The `lizmapRepository` class now extends `\Lizmap\Project\Repository`.
* Replace lizmapRepository static properties usage by Repository static properties


laurentj:
Why don't you remove the class? It has been deprecated since 2021.

rldhont:
> Why don't you remove the class? It has been deprecated since 2021.

Because `lizmap::getRepository` returned `lizmapRepository` instead of `\Lizmap\Project\Repository` and because I saw it in a deprecated Jelix class.

But I can remove this class.

Issue #5944, opened 2025-07-11
Bing maps is deprecated and should be replaced by Azure maps


nboisteault:
### Is your feature request related to a problem? Please describe
Bing maps is deprecated, [free plan does not work since June 30, 2025](https://www.bingmapsportal.com/Announcement?redirect=True).
Though Bing maps can still be used until June 30, 2028 for users paying the service.

### Describe the solution you'd like
Add Azure maps to Lizmap

Gustry:
I already thought that Bing maps was not free and a paid API key was required : 

<img width="795" height="193" alt="Image" src="https://github.com/user-attachments/assets/8b948827-d2ce-460d-ac69-a5f49cb3ac45" />

nboisteault:
> I already thought that Bing maps was not free and a paid API key was required :

You could use it for free but the number of requests was limited.

PR #5945, opened 2025-07-11
E2e: remove bing maps test

nboisteault:
Bing maps is now deprecated: https://www.bingmapsportal.com/Announcement?redirect=True

Linked to https://github.com/3liz/lizmap-web-client/issues/5944


Gustry:
CC @mind84, what do you think ? As you created the PR #4150.

mind84:
@Gustry,

> what do you think ?

Sorry for the delay, yes of course the test should be removed :). Don't know if OL is going to support this, if so I can update Lizmap accordingly.

Cheers

PR #5957, opened 2025-07-18
Fix: zoom to point features at startup

nboisteault:
Use OL GeoJSON `readFeatures` method to get filtered features extent

Ticket :  paid support 3580

Funded by FMProjet


rldhont:
A linked test failed `[end2end] › playwright/startup.spec.js:7:5 › Startup › Zoom to features extent`

PR #5991, opened 2025-07-31
Use rspack instead of webpack

nboisteault:


rldhont:
@nboisteault do you superseeded https://github.com/3liz/lizmap-web-client/pull/5872 ?

PR #6006, opened 2025-08-25
Deprecate `run-docker`

dmarteau:
Simplify commands for local and CI tests.

### Changes

Deprecate `./run-docker` which is only a pass-through to make without really value-added and allows more fine grained  command execution.

PR #6007, opened 2025-08-26
Tests: fix restrict key for dumping database

rldhont:
To check if database is not changed by readonly end2end tests, the database is dumped. pg_dump will generate a random one as needed. Keys may contain only alphanumeric characters.


rldhont:
Fix #6006

dmarteau:
I'm just asking how this solve the diff test since the reference dump has not be regenerated with the correct key ?

rldhont:
> I'm just asking how this solve the diff test since the reference dump has not be regenerated with the correct key ?

The pg_dump command will modify a file in the source tracked by git. If the command failed, it does not update the file and the diff is empty. If the command run, it updates the file but if the database has not changed, the file does not change.

And because the restrict key is not valid, you have not have to change the sql to add `\restrict` and `\unrestrict`

PR #6008, opened 2025-08-26
Tests: PHP-Unit 11 for local like in Github workflows

rldhont:
In the Github wrokflows, the PHP-Unit version is 11.5.2 and 10 for local.

The default PHP version for tests is 8.2.


rldhont:
No backport because min version f PHP is 8.1 in release_3_10.

PR #6009, opened 2025-08-26
GH Action - E2E QGIS - BLEEDING_EDGE with QGIS 3.44

rldhont:
Update BLEEDING_EDGE E2E QGIS GH Action to QGIS 3.44 the last QGIS version.


dmarteau:
FYI: #5168 use QGIS 3.40 as ltr and 3.44 as release version: end2end tests pass with 3.40 but fail  with 3.44

rldhont:
> FYI: #5168 use QGIS 3.40 as ltr and 3.44 as release version: end2end tests pass with 3.40 but fail with 3.44

It is the reason of this PR ;-)

dmarteau (approved):

rldhont:
@dmarteau same error as in #5168 we need to investigate why we do not support 3.44.

rldhont:
I made some test locally with QGIS 3.44 and some failed tests does not failed locally. So I think that the test is not well write.

rldhont:
@dmarteau @mdouchin @nboisteault I have fix QGIS 3.44 support and I will backport it to 3.10 and 3.9

PR #6010, opened 2025-08-26
Tests e2e: enhancing tooltips tests

rldhont:
1. Check the response of the tooltips request
2. Add a test to checked if an error message is displayed in case of response error


dmarteau:
This "Too many requests" error from npm registry is becoming a real pain (and above all coming from an official tool `npm install`)

PR #6041, opened 2025-08-28
Tooltip - Add fields to the response needed for SLD rendering when needed

mdouchin:
This allows the `features` controller to request `lizmap_server` plugin for the fields used in the layer symbology and add them to the response so that OpenLayers can render the tooltip features with roughly the same symbology as inside QGIS.

Follow up #5768 (this new PR should probably be merged with #5768)

Funded by Département du Gard (France) https://sig.gard.fr/


mdouchin:
Commit 6b2fee5d5ed6839d5d8c40fc92357682a647180f integrated in #5768

PR #6049, opened 2025-08-29
Api - improve rights

mdouchin:
Funded by 3liz


rldhont (approved):
Thanks, LGTM !

You have to update the test about context, because it was not authorized.

PR #6064, opened 2025-09-16
Editing - Avoid checking valuerelation & relationreference fields as valid WFS typenames

mdouchin:
In the context of editing features, Lizmap Web Client backend should access data with WFS getfeature request even if the publisher has not checked them as published in the QGIS project properties. This allows to fill the combo boxes (or multiple checkboxes) based on "value relation" or "relation reference" edit types.

Funded by 3liz


rldhont (approved):

rldhont:
@mdouchin can be merged like that.

PR #6065, opened 2025-09-16
fix : login key for other entrypoint

nworr:
When the [webdav module](https://github.com/3liz/lizmap-webdav-module) is enabled, the Login event should be logged with a specific  key (see https://github.com/3liz/lizmap-web-client/pull/5942 ) 

But #5942 didn't work as expected (request object is not instanciated yet ?) 

Screenshot 2025-09-16 at 10-48-07 Administration

So the PR use `$_SERVER` vars to find the entrypoint

Screenshot 2025-09-16 at 10-48-17 Administration
 
Funded by 3Liz


mdouchin (approved):
Thx

rldhont (approved):

PR #6092, opened 2025-09-22
JS: Update minor version dependecies

rldhont:
The caret symbol, before ver number, means you are only allowed to update the package to the latest minor version.

To update dependencies, you can do:

```bash
rm -rf node_modules/*
rm package-lock.json
npm install
```


rldhont:
I have opened this PR https://github.com/flatgeobuf/flatgeobuf/pull/458 to be able to update flatgeobuf to 4.3

Issue #6085, opened 2025-09-21
[Bug]: /admin/qgis_projects crashes if QGS project has no last save data time ?


kaloyan13:
### What is the bug? (in English)

Opening /admin/qgis_projects gives "Error 500. A technical error has occured. Sorry for this trouble."

I managed to view the page by returing a default date string:
```python
public function getLastSaveDateTime(): string
    {
        #return $this->lastSaveDateTime;
        return $this->lastSaveDateTime || '0000-00-00 0:0:0';
    }
```

### Steps to reproduce the issue

1. Install release 3.9.2 in Ubuntu 24.04
2. Login with admin user and view /admin/qgis_projects

### Versions, safeguards, check summary etc

Lizmap Web Client - 3.9.2
QGIS Server - 3.44.3 - 1d1d67e9edd 
Py-QGIS-Server -	1.9.6 

Plugins:
atlasprint Help 	3.4.3
Lizmap server Help 	2.14.0
wfsOutputExtension Help 	1.8.3

### Check Lizmap plugin

- [x] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 24.04

### Browsers

Firefox

### Browsers version

Firefox 143 for Linux x64

### Relevant log output

```shell
2025-09-21 18:58:42	[0]	Lizmap\Project\QgisProject::getLastSaveDateTime(): Return value must be of type string, null returned	/var/www/lizmap-web-client-3.9.2/lizmap/modules/lizmap/lib/Project/QgisProject.php	359
```

kaloyan13:
After patching it I saw default projects are created with an old QGIS Desktop.

<img width="1330" height="165" alt="Image" src="https://github.com/user-attachments/assets/352d2525-3308-44e5-b710-5bcb525f137a" />

rldhont:
Fixed by #6113 and available in release 3.9.3

PR #6113, opened 2025-09-30
Provide default values for *very* old qgs files with missing attributes

nworr:
Old qgs file ( < 3.10 i assume) does'nt contain `saveDateTime` in the xml project tag leading to a fatal error (see #6085)

the pr add default value (empty string) if value is null
add also mini refacto 

Ticket : #6085

Funded by 3liz


rldhont (changes requested):
Can you update `Qgis\ProjectInfo` instead ?

rldhont:
Can you also add a test in `tests/units/classes/Project/Qgis/ProjectInfoTest.php` with `tests/units/classes/Project/Ressources/WMSInfotest.qgs` ?

PR #6140, opened 2025-10-08
Fix QgisForm upload processing and widget

rldhont:
Fixing these two issues:

* The upload processing by QgisForm update the value even if it does not change.
* The upload widget does not display the file if the file does not respect the config path.

And adding e2e tests.


mdouchin (approved):
🚀

PR #6145, opened 2025-10-13
request IP now provided by JelixContext for logging event

nworr:
When authentifaction is performed, a `AuthCanLogin` event is performed and if the `logIP`  configuration is `true` the IP must be added to the authentication log.

But  if the entrypoint doesn't instanciate the request Object (ie . the [Web DAV module](https://github.com/3liz/lizmap-webdav-module/blob/main/webdav/install/dav.php) we need to compute the IP using `$_SERVER` vars (like the jelix jClassicRequest Objet)

I choose to add the method to `JelixContext` because it can be usefull for others modules

Funded by Avignon


rldhont:
LGTM, thanks @nworr

PR #6107, opened 2025-09-25
Fix regression: exporter disappear from layer information sub dock

rldhont:
An exporter tool was available in the layer information sub dock up to version 3.8. It has desappear with the ability to disable the export of layer in attribute layer.

To get back this feature, the exporter is available for all layer publshed in WFS if no attribute table config is defined or if export in attribute table is not disabled for any layer in attribute table.

The exporter is disable if at least it has been disabled for one layer in attribute table config and it is not enabled for this layer.

Fixes #5857

PR #6147, opened 2025-10-13
Fix export layer from sub-dock

rldhont:
The export tool from sub-dock is still buggy when the layer has no attribute table config.

Funded by DEAL Martinique


rldhont:
Follow https://github.com/3liz/lizmap-web-client/pull/6107

PR #6225, opened 2025-11-18
Fix digitizing tool to accept decimal values for distance and angle 

meyerlor:
Changed parseInt() to parseFloat() in the distanceConstraint and angleConstraint setters to allow decimal input. Also added step="any" attribute to the HTML number inputs to enable decimal entry in the UI.

This fixes the issue where users could only enter integer values when using the digitizing measure toggle button.

@rldhont @mdouchin  I hope i do not step your toes with those PRs! I just thought when i see room for improvements, rather then opening endless tickets i try it myself! Please let me know when i'm overstepping or making things more difficult then ease things..!

### DIFF
--- a/assets/src/components/Digitizing.js
+++ b/assets/src/components/Digitizing.js
@@ -458,6 +458,7 @@ export default class Digitizing extends HTMLElement {
                         placeholder="${lizDict['digitizing.constraint.distance']}"
                         class="distance"
                         min="0"
+                        step="any"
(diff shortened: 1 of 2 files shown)


rldhont:
Thanks @meyerlor for this fix. To which version do you think it has to be backported ?

meyerlor:
would love to see it in 3.9 if that's possible? That's where my client was missing it :)

PR #6230, opened 2025-11-19
Tests e2e: migration from Cypress to Playwright attribte table

rldhont:
Move attribute table tests from cypress to playwright


rldhont:
@nboisteault @mdouchin ready for review

PR #6244, opened 2025-11-21
e2e: fix chrome version to 140 for Cypress in CI

nboisteault:
Funded by 3Liz

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -325,7 +325,7 @@ jobs:
         uses: cypress-io/github-action@v6.10.2
         with:
           install: false
-          browser: chrome
+          browser: chromium
           headed: true
           working-directory: tests/end2end
           spec: cypress/integration/*-ghaction.js


rldhont:
`Chromium 142`

rldhont:
We probably have to use https://github.com/browser-actions/setup-chrome as explain in https://github.com/cypress-io/github-action about using chrome-testing

.github/workflows/e2e_tests.yml
```diff
@@ -325,7 +325,7 @@ jobs:
         uses: cypress-io/github-action@v6.10.2
         with:
           install: false
-          browser: chrome
+          browser: chromium
```
rldhont:
```
      uses: browser-actions/setup-chrome@v2
        with:
          chrome-version: 140
      uses: cypress-io/github-action@v6.10.2
        with:
          install: false
          browser: chrome-for-testing
```

nboisteault: That's what I've tested just before https://github.com/3liz/lizmap-web-client/actions/runs/19568152240/job/56035070077?pr=6244#step:29:50

But I think that the `chrome-for-testing` parameter has been introduced after our Cypress version.

PR #6261, opened 2025-11-25
e2e: cypress w/ firefox browser in CI

nboisteault:
Cypress 9.7.0 does not work w/ Chrome. Let's use Firefox instead of Chrome

Funded by 3Liz


rldhont:
I close it because Cypress has been removed !

Issue #5852, opened 2025-06-23
[Bug]: LWC tries loading Legend graphics even if layer has no legend (disabled via plugin)


meyerlor:
### What is the bug? (in English)

Not really a bug  but i wanted to report it anyway, as mentioned here already: https://github.com/3liz/lizmap-web-client/pull/5819#issuecomment-2989949361

LWC is sending a lot of GET requests, even if the layer has it's Legend disabled. might be worth considering to skip those GET's ? 

No Problems arise from this besides a very full dev tool with my 750-layer project (650 of them with legend disabled)

![Image](https://github.com/user-attachments/assets/025998ca-8282-4e58-b159-2684300415f8)

### Steps to reproduce the issue

1. Disable Legend in LWC QGIS plugin
2. Load Project in webbrowser of choice, dev tools /network monitoring on
3. Many unnecessary GET requests apear

### Versions, safeguards, check summary etc

**Versions :**

* Lizmap Web Client : 3.9.0 - commit 4a0d5ea52 https://github.com/3liz/lizmap-web-client/commit/4a0d5ea52
* Lizmap plugin : 4.5.1
* QGIS Desktop : 3.40.7
* QGIS Server : 3.40.7
* Py-QGIS-Server : 1.9.4
* QGIS Server plugin atlasprint : 3.4.2
* QGIS Server plugin lizmap_server : 2.13.1
* QGIS Server plugin wfsOutputExtension : 1.8.3

<details>
<summary>List of Lizmap Web Client modules :</summary>
<br/>
* Version Lizmap Web Client 3.8 needed
</details>

<details>
<summary>List of safeguards :</summary>
<br/>
* Mode : normal<br/>
* Allow parent folder : no<br/>
* Prevent other drive : yes<br/>
* Prevent PG service : yes<br/>
* Prevent PG Auth DB : yes<br/>
* Force PG user&pass : yes<br/>
* Prevent ECW : yes<br/>
</details>

### Check Lizmap plugin

- [ ] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 24.04

### Browsers

Firefox

### Browsers version

120

### Relevant log output

```shell

```

meyerlor:
Closed with https://github.com/3liz/lizmap-web-client/pull/6279

PR #6279, opened 2025-11-28
Respect legend_image_option setting when loading symbology

meyerlor:
The legend_image_option configuration was only checked when displaying symbology in the UI, but symbology was still being loaded from the server even when set to 'disabled'. This resulted in unnecessary network requests.

Filter out layers with legendImageOption set to 'disabled' before loading symbology, both during initial load and when layer styles change.

Fixes unnecessary symbology loading for layers with disabled legends:
https://github.com/3liz/lizmap-web-client/issues/5852

### DIFF
--- a/assets/src/modules/Legend.js
+++ b/assets/src/modules/Legend.js
@@ -6,7 +6,7 @@
  */
 
 import { LayerTreeGroupState } from '../modules/state/LayerTree.js'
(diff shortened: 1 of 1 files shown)


meyerlor:
Current behaviour (especially for Rasters with pallete it created a huge load):
grafik


Fixed behaviour:
grafik

Just the hardcoded raster symbol is shown now, no palette legend

rldhont (approved):

PR #6280, opened 2025-11-28
[Bugfix] Editable features in attribute table

mind84:
Closely related to https://github.com/3liz/lizmap-web-client/pull/5479

In this case the `editableFeatures` method is called by the attribute table and, as with the aforementioned PR, if the table contains a lot of records, the risk of PHP crashes (exceeding the maximum execution time) increases.

Since `editableFeatures` method of `qgisVectorLayer` class already accepts extra WFS parameters (after https://github.com/3liz/lizmap-web-client/pull/5479), it may be convenient to pass filters from client to server to limit (or attempt to)  the number of records processed.

This solution does not completely eliminate the risk of processing a huge amount of data but, instead, introduces a basic control, if applicable. 

The best approach would be to do this control directly server side as "middleware" between the WFS response from QGIS server and the `editableFeature` response send by Lizmap to the client, but since this PR primarly concerns legacy attribute table code, we intended to temporarily patch this behavior while awaiting further attribute table code implementations that will be available in the future.

Funded by Faunalia

### DIFF
--- a/tests/end2end/playwright/editable-features.spec.js
+++ b/tests/end2end/playwright/editable-features.spec.js
@@ -1,0 +1,69 @@
+import { test, expect } from '@playwright/test';
+import { expect as requestExpect } from './fixtures/expect-request.js'
+import { expect as responseExpect } from './fixtures/expect-response.js'
+import {ProjectPage} from "./pages/project";
+import { getAuthStorageStatePath } from './globals';
+
+test.describe('Editable features',
+    {
+        tag: ['@readonly'],
+    },()=> {
+        test('Check attribute table editable features - anonymous', async ({ page }) => {
+            const project = new ProjectPage(page, 'attribute_table_editable_features');
+            await project.open();
+
+            const layerName = 'filter_layer_by_user';
+            let getFeatureRequestPromise = project.waitForGetFeatureRequest();
+            await project.openAttributeTable(layerName);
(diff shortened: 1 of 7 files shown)


mind84:
Backport 3.9 and 3.10

tests/end2end/playwright/editable-features.spec.js
```diff
@@ -0,0 +1,65 @@
...
+                'features': '"gid" = \'1\' OR "gid" = \'3\' OR "gid" = \'2\'',
+            };
+            await expectParametersToContain('Editable features on attribute table - anonymous', getEditableFeaturesRequest.postData() ?? '', expectedParameters);
+            const response = await getEditableFeaturesRequest.response();
+
+            // check response
+            expect(response?.ok()).toBeTruthy();
+            expect(response?.status()).toBe(200);
+            // check content-type header, content is streamed by the server
+            expect(response?.headers()['content-type']).toContain('application/octet-stream');
```
rldhont: The response for editable features request has to be `application/json`. The controller has to be enhanced with `$rep->mimeType = 'application/json';`

Once done the expect response fixture could be use.

mind84:
Fixed `filter-layer-by-polygon` test error

mind84:
Others tests failure appears to be unrelated with this PR

PR #6284, opened 2025-12-01
Fix: Prevent parent groups from being auto-checked

meyerlor:
…when only child groups are in theme

When a theme includes nested groups like "ParentGroup/ChildGroup", only the child group should be checked, not the parent.

Previously, the automatic parent propagation in the state management would check parent groups when their children were checked, even if the parent wasn't explicitly listed in the theme's checked-group-nodes.

This fix uses a two-pass approach:
1. Set all groups to their intended state (child groups auto-check parents)
2. Uncheck parent groups that aren't explicitly in the theme configuration

Sorry for not beeing carefull enough - when implementing @mdouchin suggestion regarding legend-nodes it also led to checked-group-nodes. I missed that checking a group node would propagate the parent group node checked as well - which is not wanted. So we need to stick for group-nodes to a two-pass approach i think!

### DIFF
--- a/assets/src/legacy/switcher-layers-actions.js
+++ b/assets/src/legacy/switcher-layers-actions.js
@@ -388,28 +388,69 @@ var lizLayerActionButtons = function() {
                         }
 
                         // STEP 2: Set groups checked/expanded state based on theme configuration
+                        // Use a two-pass approach to handle automatic parent checking:
+                        // - Pass 1: Set all groups (child groups will auto-check parents)
+                        // - Pass 2: Uncheck parent groups that shouldn't be checked
+
+                        // Helper function to check if group is in list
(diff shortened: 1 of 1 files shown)


rldhont (approved):

rldhont:
Hi @meyerlor I am working on e2e tests before merging your fix to be sure that nothing has been broken in the future.

meyerlor:
yes that is a very good idea - i think especially the https://github.com/3liz/lizmap-web-client/blob/master/tests/qgis-projects/tests/theme.qgs project needs some more themes which reflect all the possibilities. I can not change it with my QGIS without access to the datasource, otherwise i would have done so already! If you create/change that .qgs file, i can try to add the e2e tests!

rldhont:
I have tested it, and I will provided e2e tests.

Issue #6061, opened 2025-09-03
[Bug]: User set transparency does not work (LWC 3.9.1)


meyerlor:
### What is the bug? (in English)

Two issues here:
1) The infobutton - "popup" wont show the option to change the transparency of a layer:

<img width="714" height="674" alt="Image" src="https://github.com/user-attachments/assets/e9761b68-4355-4d26-88e4-6680e2d17ea1" />

2) The infobutton "popup" of a layer group shows the transparency selector, but it has no effect:

<img width="698" height="504" alt="Image" src="https://github.com/user-attachments/assets/755f38fc-47bd-4fbe-8cf8-f8e799be2e7b" />

### Steps to reproduce the issue

Try to set transparency for a layer(group) as logged in user is not possible

### Versions, safeguards, check summary etc

**Versions :**

* Lizmap Web Client : 3.9.1 - commit 48696fcf0 https://github.com/3liz/lizmap-web-client/commit/48696fcf0
* Lizmap plugin : 4.5.3
* QGIS Desktop : 3.40.9
* QGIS Server : 3.40.7
* Py-QGIS-Server : 1.9.4
* QGIS Server plugin atlasprint : 3.4.2
* QGIS Server plugin lizmap_server : 2.13.1
* QGIS Server plugin wfsOutputExtension : 1.8.3

<details>
<summary>List of Lizmap Web Client modules :</summary>
<br/>
* altiProfil : 0.5.10
* altiProfilAdmin : 0.5.10
</details>

<details>
<summary>List of safeguards :</summary>
<br/>
* Mode : normal<br/>
* Allow parent folder : no<br/>
* Prevent other drive : yes<br/>
* Prevent PG service : yes<br/>
* Prevent PG Auth DB : yes<br/>
* Force PG user&pass : yes<br/>
* Prevent ECW : yes<br/>
</details>

### Check Lizmap plugin

- [x] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 24

### Browsers

Firefox

### Browsers version

Firefox 125

### Relevant log output

```shell

```

guenterw:
Strange, they are displayed for me for individual layers with points, lines, and hatches (Lizmap 3.9.1 with QGIS Server 3.40.10, Py-QGIS Server)

[...]

PR #6351, opened 2025-12-13
Exclude basemaps from single WMS layer option

meyerlor:
Having local basemaps like Orthofotos in an Cloud-Optimized format (COG - JPEG tiles in a TIFF container) and setting the option "Load layers as single WMS" highly degrades rendering time of lizmap projects. Switching from "no basemap" or OSM to the DOP adds 2-3s of rendering time for the client.

Checking the qgis-server logs led me to the conclusion that the problem can not be here: rendering time for the DOP was well below 1s (300-500ms) so the high cost likely comes from converting the original JPEG to PNG (the format of the combined requests) - setting the file format of the DOP layer in the lizmap plugin to jpeg does not change anything, as the JPEG-based DOP is added to the WMS stack which is delivered in PNG.

I propose to always render basemaps as separate layers and exclude them from the single WMS layer stack to circumvent this problem and deliver DOP as JPEG and speed up rendering time by the factor 10 (300ms instead of ~3000).

I also updated tests accordingly.

I'm very open to other suggestions for this particular problem!

The as-is status/loading time:
grafik

Here the result of rendering times with the proposed solution:
grafik

### DIFF
--- a/assets/src/modules/SingleWMSLayer.js
+++ b/assets/src/modules/SingleWMSLayer.js
@@ -137,10 +137,13 @@ export default class SingleWMSLayer {
         this._orderedLayers = [];
 
         // construct base and map layers array
+        // Also ensure visibility is calculated for each layer before first render
         this._singleWMSLayerList.forEach((m,k) => {
             if (m instanceof BaseLayerState) {
                 this._baseLayers.push(k);
             } else if (m instanceof MapLayerState){
+                // Force visibility calculation in case it hasn't been computed yet
+                m.calculateVisibility();
                 this._mapLayers.push(k);
             }
         });
@@ -343,7 +346,8 @@ export default class SingleWMSLayer {
                     this._layerStyles.push(selectedBaseLayerState.wmsSelectedStyleName || "");
                 }
             } else if (currentLayerState instanceof MapLayerState){
-                // get item visibility
+                // Check if layer should be rendered using visibility
+                // visibility considers both the layer's checked state and parent group visibility
                 if(currentLayerState.visibility) {
                     this._layersName.push(currentLayerState.name);
                     this._layersWmsName.push(currentLayerState.wmsName);
--- a/assets/src/modules/config/Options.js
+++ b/assets/src/modules/config/Options.js
@@ -36,7 +36,8 @@ const optionalProperties = {
     'hideGroupCheckbox': { type: 'boolean', default: false },
     'activateFirstMapTheme': { type: 'boolean', default: false },
     'automatic_permalink': { type: 'boolean', default: false },
-    'wms_single_request_for_all_layers' : { type:'boolean', default: false }
+    'wms_single_request_for_all_layers' : { type:'boolean', default: false },
+    'exclude_basemaps_from_single_wms' : { type:'boolean', default: false }
 };
 
 /**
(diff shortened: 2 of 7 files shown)


meyerlor:
@mind84 thanks for [pointing me](https://github.com/3liz/lizmap-web-client/issues/6061#issuecomment-3670547208) to your [original work ](https://github.com/3liz/lizmap-web-client/pull/3947) - i read the conversation of that PR just now :)

If i understand it right, @mdouchin [suggested](https://github.com/3liz/lizmap-web-client/pull/3947#issuecomment-1790511813) to exclude not only external (wms) but also "local" baselayers from the "single wms"-stack - just like what i try to introduce here with this PR. You probably had good reasons why not to do that in the final version of your PR - would you mind to share your thoughts on that?

mind84:
Hi @meyerlor!

> You probably had good reasons

The original idea is to limit the number of request to the server as much as possible. In project with hundreds of layers it makes a huge difference. This is why all layers that could potentially end up in the single request have been included, including some particular types of baselayers. I don't know honestly whether this could be categorized as good reason or not, but at time seems to me reasonable enough :). That said, I haven't considered the DOP format in this scenario, and since only one baselayer can be displayed on the map at a time, regardless of the number of baselayers configured, it might be a valid option to leave all baselayers out of the single request. I'll take some time to think about any other reasons and let you know.

Cheers and happy new year!

meyerlor:
Haha alright! I experimentet quite a bit with file formats and ended up converting raw .tiff files (the format we usually get for DOP) into a single JPEG containered Cloud Optimized GeoTIFF:
C:\OSGeo4W\bin\gdal_translate.exe Input_VRT.vrt Output_raster.tif -a_srs EPSG:25832 -of COG -co BLOCKSIZE=512 -co COMPRESS=JPEG -co QUALITY=75 -co BIGTIFF=YES -co NUM_THREADS=ALL_CPUS -co OVERVIEW_RESAMPLING=AVERAGE -co OVERVIEW_COUNT=12
Which is quite reasonable in file size, quality and rendering speed. According to my (AI) research, QGIS server has to convert this JPEG to PNG which is slow. But it also suggested, using DEFLATE instead of JPEG to compress should be much faster when converting it to PNG for the WMS - but it apparently was same (or even a bit slower)..

Anyways, happy new year to you as well! Enjoy the holidays! :)

mind84:
Hi @meyerlor!

I think that in this specific case, you're seeing a performance improvement, but in other cases (e.g., vector baseLayers), performance could probably be a bit worse. Overall, considering the most common use cases, I'd say it probably makes sense to exclude all baselayers from the Single WMS request.

Cheers!

meyerlor:
Maybe we could introduce an additional toggle in the plugin "exclude basmap layer group from single WMS stack", so it is up to the admin to decide how basemaps should be handled? 

any thoughts about this @rldhont?

rldhont:
> Maybe we could introduce an additional toggle in the plugin "exclude basmap layer group from single WMS stack", so it is up to the admin to decide how basemaps should be handled?
> 
> any thoughts about this @rldhont?

Good idea!

meyerlor:
👍  ok, so i will create a PR for the plugin and change this one here to react according to the .cfg

gioman:
Sorry of hijack this conversation to make possibly a silly question: is this about getting a smooth experience when zooming in/out in a base map as like as you get when you open (for example) https://www.openstreetmap.org/ ?

meyerlor:
This would be a possible side effect. just to understand you right: when i have OSM as basemap, this runs _very_ smoothly. So do other WMS/XYZ-tiles based external services. Is this not the case with your projects?

Currently, if you toggle in the lizmap plugin the "Load layers as single WMS layer (Under "Map Options), local baselayers (like DOP) are added to this stack. As this stack is delivered in PNG, QGIS server needs to convert your DOP (in my case: .tiff) to PNG which takes ~3sec. So your whole map rendering is waiting until this DOP is converted. That is what my PR wants to circumvent: deliver the DOP seperately from the rest of the (vector) layers, so it can be delivered in the original .tiff or JPEG format, which will be much faster/smoother. 
Hope to have explained it precice/understandable enough, i'm no native english speaker :-)

meyerlor:
i think finally it's good - the tests were a bit hard to comply. The failing legacy e2e tests are not related to my PR (i think?)

rldhont:
Thanks @meyerlor

PR #6368, opened 2025-12-16
Some UI fixes

meyerlor:
This aims to fix 3 minor issues i found:

1. The tooltip for the digitizing tools when creating lines/polygons was hidden:
Tooltip-Digitizing

I'm a bit insecure regarding UI development - in 3.11 this tooltip fix works but if i change the map.css manually in 3.9.4 the tooltip still won't show. :(

The other two points affect the behaviour on mobile devices:

2. In Landscape mode, the "Baselayer" Dropdown menu was hidden/unreachable:
Basemap_portrait_copy

Fixed:
Baselayer_available

3. The menu on mobile devices is unnecessary broad:
Breite_legende

Improved:
fixed_legende

### DIFF
--- a/lizmap/www/assets/css/media.css
+++ b/lizmap/www/assets/css/media.css
@@ -264,8 +264,9 @@ only screen and (                   device-height <= 640px) {
     width: max-content; /* Automatic width based on content */
     min-width: 250px;
     max-width: 400px;
-    left: -100%; /* Completely outside viewport */
-    transition: left 0.3s cubic-bezier(0.77,0.2,0.05,1.0);
+    left: 0;
+    transform: translateX(-100%); /* Completely outside viewport */
+    transition: transform 0.3s cubic-bezier(0.77,0.2,0.05,1.0);
   }
 
   #mapmenu > div{
@@ -332,7 +333,7 @@ only screen and (                   device-height <= 640px) {
   }
 
   #menuToggle.opened ~ #mapmenu{
-    left: 0;
+    transform: translateX(0);
   }
 
   #menuToggle{


rldhont:
Thanks @meyerlor 

2 tests are failing with your changes :
* map-projects-switcher.spec.js `Switcher from map to map` because `#navbar button.btn.zoom-in` is probably not visible
* base-layers.spec.js `Base layers list` the project seems to not load as expected

For which version your fix is for ?

meyerlor:
@rldhont sorry, this is a bit more messy / complicated then i anticipated.

The fix for the missing tooltip tips when editing geometries is working in Lizmap 3.11.0-pre - but when I use the map.css / media.css in a 3.9.4 instance, the tooltip still won't show.

The missing baselayer - drop down menu shows up in 3.9.4 but not in 3.11

The menu broadness seems to work in both (although in 3.11. the menu tools are not in a list but in a  side-by-side stack:
grafik


I can work on it or just remove the pr - UI is kind of a minefield it seems :(

rldhont:
Hi @meyerlor, can we split this PR in two different one ?

I think the commit **Fix tooltip z-index in edition digitization tab** is for **LWC 3.10** because it fixes a **Bootstrap 5** issue in lizmap, and could not be used for LWC 3.9 because it is bootstrap 2.

The other commits is for LWC 3.10 and 3.9 because it is about the docks on mobile. It does not matter which bootstrap version is used. We have to fix how the menu is displayed on mobile with bootstrap 5 but it is not the purpose of your fix.

Are you agree ?

meyerlor:
yes sure thing - i will split it according to your suggestions !

meyerlor:
(or was your plan to split it yourself? feel free to reopen, i probably won't be able to do something here before the weekend)

rldhont:
Hi @meyerlor I have created the two PRs and will merged them.

meyerlor:
much appreciated!

PR #6370, opened 2025-12-18
[Bug] Error on add child element from parent attribute table

mind84:
Fixes syntax error in legacy edition code

Backport 3.9 and 3.10

Funded by Faunalia


rldhont:
Thanks @mind84

PR #6420, opened 2026-01-13
replaySubject notification

mind84:
Hi @mdouchin, @rldhont, @nboisteault,

this is only for demonstration purpose of what I meant with 

https://github.com/3liz/lizmap-web-client/pull/6414#issuecomment-3729514447

so you can evaluate whether it could be a valid solution or not.

This basically allows to "listen" to "event" already emitted and can replace the uiready event (covering the unfortunate cases when uiready/uicreated is emitted before the js custom is fully loaded).

You can also add the attached custom js in media to see how it works:

[test_ui_ready_subscriber.js](https://github.com/user-attachments/files/24592204/test_ui_ready_subscriber.js)

### DIFF
--- a/assets/src/index.js
+++ b/assets/src/index.js
@@ -26,6 +26,7 @@ import Treeview from './components/Treeview.js';
 import NavBar from './components/NavBar.js';
 import Tooltip from './components/Tooltip.js';
 import Message from './components/Message.js';
+import { ReplaySubject } from 'rxjs';
 
 import { mainLizmap, mainEventDispatcher } from './modules/Globals.js';
 import executeJSFromServer from './modules/ExecuteJSFromServer.js';
@@ -74,6 +75,10 @@ const definedCustomElements = () => {
     window.customElements.define('lizmap-navbar', NavBar);
     window.customElements.define('lizmap-tooltip', Tooltip);
     window.customElements.define('lizmap-message', Message);
+
+    // call the next() method on ReplaySubject.
+    // every subscribers of this ReplaySubject receive a notification
+    lizMap.uiReadySubscriber.next(1);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


rldhont:
Hi @mind84, thanks for pointing us this library https://rxjs.dev/

We have some legacy events based on OpenLayers 2 events class and we have an `EventDispatcher` as a singleton and used in `State` classes.

```js
/**
 * Export global objects needed by lizmap components and classes
 * Internal use only
 */

import Lizmap from './Lizmap.js';
import EventDispatcher from '../utils/EventDispatcher.js';

const mainEventDispatcher = new EventDispatcher();
const mainLizmap = new Lizmap(mainEventDispatcher);

export {
    mainLizmap,
    mainEventDispatcher
};
```

```js
/**
 * The main Lizmap definition
 * @class
 * @name Lizmap
 */
export default class Lizmap {

    constructor(eventDispatcher) {
        this._eventDispatcher = eventDispatcher;
...
    }

...

    /**
     * The main event dispatcher
     * @type {EventDispatcher}
     */
    get eventDispatcher() {
        return this._eventDispatcher;
    }

...

}
```

```js

/**
 * Class representing a layer item state: could be group, vector or raster layer
 * @class
 * @augments EventDispatcher
 */
export class LayerItemState extends EventDispatcher {
...
}
```

We have to remove the legacy events, but how can we reconcile the Reactive Extensions Library and our EventDispatcher ?

mdouchin:
Perhaps we could find a way to have the same behaviour as the librairy but without depending on it ? Not sure it is possible, but a native JS event dispatcher which fit our needs would be better IMHO, if possible

Some nice introduction to observables : https://sabrinabea.com/JavaScript/sytykjs/16-observables.html

mind84:
Hi @rldhont, @mdouchin,

it's definitely possible to replicate the behavior of `rxjs` lib without add it as dependency, probably by enhancing the `EventDispatcher` class. The goal, to me, is to remove the ol2 dependency but this requires a massive refactor.

I'll keep this PR as draft and try to implement the ReplaySubject behavior with native js. Probably it'll take some time :)

Thank you!

rldhont:
Hi @mind84, `rxjs` is great lib and could be usefull. If it could help to use `Observable`, we can use it.

PR #6428, opened 2026-01-14
Fix sync edition map and view map

rldhont:
The edition map is managed by OL2 and view map by OL10. The sync is not well performed.

Funded by EPF-HDF


rldhont:
@mdouchin can you test it ?

PR #6433, opened 2026-01-16
Remove type hint for message parameter in log method

### LINKED ISSUE #6431: [Bug]: PHP Fatal Error in Lizmap\Logger\Logger due to signature mismatch with psr/log: 2.*

A blocking PHP Fatal Error occurs in the 3.10 branch. The issue is caused by a signature mismatch between the Lizmap\Logger\Logger class and the Psr\Log\AbstractLogger interface, based on the dependencies defined in the current composer.json.``

Technical Details

lizmap/modules/lizmap/classes/Logger.php

Lizmap Version: 3.10.x (Alpha/Dev)

PHP Version: 8.2+

Dependency in composer.json: "psr/log": "2.*"

The Issue The composer.json file explicitly requires version 2.x of the PSR-3 logging library. In this version, the log method signature is: 

`public function log($level, $message, array $context = array())`

However, the implementation in Logger.php for the 3.10 branch uses PHP 8 Union Types and a return type hint: 

`public function log($level, string|\Stringable $message, array $context = array()): void`

Since Lizmap\Logger\Logger extends AbstractLogger, PHP throws a Fatal Error because the child class signature is more restrictive than (and thus incompatible with) the parent interface provided by psr/log v2.x.

`Error Message Fatal error: Declaration of Lizmap\Logger\Logger::log($level, string|\Stringable $message, array $context = array()): void must be compatible with Psr\Log\LoggerInterface::log($level, $message, array $context = array())`

What's the correct way to fix?

1. Update Requirements: Update composer.json to require "psr/log": "^3.0" to align the dependency with the new code signature?

2. Restore Compatibility: If compatibility with psr/log 2.x is intended for the 3.10 branch, the string|\Stringable type hint and the : void return type should be removed from Logger.php?

"I have temporarily fixed this locally by removing the type hints at line 117 of Logger.php, which restored the application's stability without needing a full composer update."

**Versions :**

* Lizmap Web Client : 3.10.0-alpha.1
* Lizmap plugin : 4.5.7
* QGIS Desktop : 3.40.14
* QGIS Server : 3.40.14
* Py-QGIS-Server : 1.9.6
* QGIS Server plugin DataPlotly : 4.3.2

<details>
<summary>List of Lizmap Web Client modules :</summary>
<br/>
* Version Lizmap Web Client 3.8 needed
</details>

<details>
<summary>List of safeguards :</summary>
<br/>
* Mode : normal<br/>
* Allow parent folder : yes<br/>
* Number of parent : 10 folder(s)<br/>
* Prevent other drive : no<br/>
* Prevent PG service : no<br/>
* Prevent PG Auth DB : no<br/>
* Force PG user&pass : no<br/>
* Prevent ECW : no<br/>
</details>

### Check Lizmap plugin

### Operating system

Ubuntu 23.04

### Browsers

Chrome

### Browsers version

Chrome 143

### Relevant log output

```shell

```

Antoviscomi:
Fixes #6431
PHP Fatal Error in Lizmap\Logger\Logger due to signature mismatch with psr/log: 2.*

Ticket : #

Funded by

### DIFF
--- a/lizmap/modules/lizmap/lib/Logger/Logger.php
+++ b/lizmap/modules/lizmap/lib/Logger/Logger.php
@@ -116,8 +116,7 @@ class Logger extends AbstractLogger
      *
      * @throws InvalidArgumentException
      */
-    public function log($level, $message, array $context = array())
-
+    public function log($level, $message, $context = array())
     {
         if (!in_array($level, self::LogLevels)) {
             throw new InvalidArgumentException('Invalid log level');


Antoviscomi:
@rldhont 
I'm facing an issue with the CI/Fixer. To avoid the Fatal Error caused by the psr/log: 2.* dependency, we must remove the native type hints string|\Stringable from the log method signature. However, the PHP-CS-Fixer is configured to enforce these types and keeps failing the build.

Since we cannot add the types without breaking the inheritance with PSR-Log v2, should we:

Exclude this file from the fixer rules?

Downgrade the strictness of the fixer for this specific method?

Update composer.json to psr/log: ^3.0 to allow the modern signature?

Antoviscomi:
@rldhont 
The DocBlock `@param string|\Stringable $message `correctly documents the PSR-3 v3.x intent, but the native type hint cannot be added to the code because it breaks inheritance with the `psr/log: 2.*` requirement in `composer.json`. The fixer is currently forcing a signature that is incompatible with the project's own dependencies. 
how to solve this?

lizmap/modules/lizmap/lib/Logger/Logger.php
```diff
@@ -110,11 +110,14 @@ public function isLevelHighEnough(string $level): bool
...
-     * @param mixed $level
+     * @param mixed              $level
+     * @param string|\Stringable $message
+     * @param array              $context
      *
      * @throws InvalidArgumentException
      */
-    public function log($level, string|\Stringable $message, array $context = array()): void
+    
+    public function log($level, $message, array $context = array())
```
rldhont:
```suggestion
    public function log($level, $message, array $context)
```
PHP-CS-fixer try to do this.

Antoviscomi (commented):
I readding the value to `$context`

rldhont:
> PHP-CS-Fixer requested some lint changes.

I have tested this and PHP-CS-Fixer is Ok
```php

    /**
     * Logs with an arbitrary level.
     *
     * @param mixed              $level
     * @param string|\Stringable $message
     * @param array              $context
     *
     * @return void
     *
     * @throws InvalidArgumentException
     */
    public function log($level, $message, $context = array())
    {
        if (!in_array($level, self::LogLevels)) {
            throw new InvalidArgumentException('Invalid log level');
        }

        // If the level is not high enough, return early.
        if ($this->isLevelHighEnough($level)) {
            return;
        }

        \jLog::log(
            $this->interpolate((string) $message, $context),
            $level
        );
    }
```

Antoviscomi:
> > PHP-CS-Fixer requested some lint changes.
> 
> I have tested this and PHP-CS-Fixer is Ok
> 
> ```
>     /**
>      * Logs with an arbitrary level.
>      *
>      * @param mixed              $level
>      * @param string|\Stringable $message
>      * @param array              $context
>      *
>      * @return void
>      *
>      * @throws InvalidArgumentException
>      */
>     public function log($level, $message, $context = array())
>     {
>         if (!in_array($level, self::LogLevels)) {
>             throw new InvalidArgumentException('Invalid log level');
>         }
> 
>         // If the level is not high enough, return early.
>         if ($this->isLevelHighEnough($level)) {
>             return;
>         }
> 
>         \jLog::log(
>             $this->interpolate((string) $message, $context),
>             $level
>         );
>     }
> ```

@rldhont thank you for your patience and suggestions, the check now it seems passed

Antoviscomi:
@rldhont there is the knowed failure on E2E QGIS 3.40 PG 14-3 PHP 8.2

rldhont:
> @rldhont there is the knowed failure on E2E QGIS 3.40 PG 14-3 PHP 8.2

The `WFS GetFeature random_points` is flacky. Even though the data source is an SHP file, QGIS Server does not always return the same result. It is probably due to virtualization and Docker.

Thanks to help fix this issue https://github.com/3liz/lizmap-web-client/issues/6431

rldhont (approved):

PR #6449, opened 2026-01-26
Patch event listeners for touch events to be non-passive, Fix scroll-blocking touch event listeners in OpenLayers legacy

Antoviscomi:
Added a patch to modify event listener behavior for touch events, ensuring non-passive listeners are used to maintain map interaction fluidly.

This PR addresses the performance violations and laggy interactions on touch devices caused by `non-passive` event listeners in the legacy `OpenLayers 2` components.

Chrome and other modern browsers flag these as` [Violation]` because they block the main thread while waiting for a potential `preventDefault()`. In a mapping context, this delay results in jerky zoom and pan movements.

Changes
Added a global `EventTarget` patch in `index.js`: This interceptor captures touchstart and touchmove events.

Explicitly set `passive: false`: While it sounds counter-intuitive, forcing passive: false tells the browser that we intend to use `preventDefault()` for map interactions, eliminating the 100-200ms "observation" delay the browser would otherwise impose.

Ensured compatibility: The patch handles both boolean and object-based `addEventListener` signatures, ensuring no breakage in legacy code or newer components.

Performance Impact
Interaction Latency: Reduced. Map panning and zooming on mobile/touch screens are now significantly smoother.

Console Violations: Mitigated the "Added `non-passive` event listener" warnings.

Load Time: During testing, the overall map initialization remains fast (approx. 640ms in dev environment).

Testing
Tested on Chrome  144.0.7559.96 with mobile emulation and physical touch devices.

Verified that map panning, zooming, and drawing tools (which require preventDefault) still function correctly.

Confirmed that Ctrl+Shift+R (hard refresh) is necessary to clear the legacy JS cache and apply the fix.

### DIFF
--- a/assets/src/index.js
+++ b/assets/src/index.js
@@ -35,7 +35,30 @@ import litHTMLDep from './dependencies/lit-html.js';
 import { proj4 } from 'proj4rs/proj4.js';
 
 /**
- *  waitFor function returns waiting time in milliseconds
+ * Patch to mitigate [Violation] "Added non-passive event listener" warnings.
+ * This ensures that OpenLayers 2 legacy touch events can still call preventDefault()
+ * while signaling to the browser that we are intentionally using non-passive listeners
+ * to maintain map interaction fluidly.
+ */
+(function patchPassiveEventListeners() {
+    const originalAddEventListener = EventTarget.prototype.addEventListener;
+    EventTarget.prototype.addEventListener = function(type, fn, options) {
+        let modifiedOptions = options;
+        if (type === 'touchstart' || type === 'touchmove') {
+            if (typeof options === 'object') {
+                // Force passive to false to ensure map panning/drawing works correctly
+                modifiedOptions = { ...options, passive: false };
+            } else {
+                // If options was a boolean (capture), convert to object
+                modifiedOptions = { capture: !!options, passive: false };
+            }
+        }
+        originalAddEventListener.call(this, type, fn, modifiedOptions);
+    };
+})();
+
+/**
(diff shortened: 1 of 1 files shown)


nboisteault:
@Antoviscomi OL2 map is only used for edition. Can you confirm that you watched lags only in edition mode?

Antoviscomi:
> @Antoviscomi OL2 map is only used for edition. Can you confirm that you watched lags only in edition mode?

Yes, @nboisteault , I can confirm that. 
While the lags were significantly more pronounced during edition tasks (where interaction is more intense), I consistently observed the `[Violation] Added non-passive event listener to a scroll-blocking touchstart event` warnings right at the initial map load, even before entering edition mode. As in classic browser view as in Touch emulation view mode by Chrome dev console

It appears that the OpenLayers 2 legacy core registers these listeners globally upon initialization. By patching these to be non-passive, I've seen two major benefits:

At Startup: The browser console is now clean of violation warnings, and the main thread is no longer throttled during the initial rendering.

In Edition: The interaction is finally fluid, as the browser doesn't have to wait for the scroll-blocking check on every touch event.

Furthermore, without the patch, those violations would multiply exponentially during editing. Since every map pan or vertex drag triggers continuous redraws and event processing, the scroll-blocking nature of the original listeners caused a massive overhead. Now, even with heavy interaction, the thread stays free and the 'incalculable' violations I previously saw are completely gone

In my current environment (running via Rspack/Apache), the load time is now a solid 670ms and the map feels much more responsive overall.

PR #6498, opened 2026-01-30
Edition with new draw tool

nboisteault:
Funded by 3Liz

### DIFF
--- a/assets/src/components/Digitizing.js
+++ b/assets/src/components/Digitizing.js
@@ -41,7 +41,7 @@ import '../images/svg/file-upload.svg';
  * It is mandatory and provide a way to use this element for different contexts.
  *
  * The other attributes are:
- *  active - Activate the element on load
+ *  active - Activate the element on load or when the attribute is added
  *  selected-tool - Start selected drawing tools one of DigitizingAvailableTools or available-tools
  *  available-tools - List of available drawing tools based on DigitizingAvailableTools
  *  save - Enable save capability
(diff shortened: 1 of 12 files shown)


meyerlor:
@nboisteault just beeing curious - if i understand your PR right, the digitizing tool will change. One of my remaining top-priorities to implement in LWC is to add some of the "advanced digitizing" tools QGIS offers. Namely:
- Drawing parallel lines in a set distance
- Trim/Extend Lines
- Draw a circle with a set radius
- Draw line segments with a set arc (like: 90° to the first drawn segment...)

So at the moment it does not make sense to start working on this until you are finished, right? 😄

nboisteault:
@meyerlor current digitizing tool in edition will be replaced by the `<lizmap-digitizing>` web component so you can begin a PR to extend it.

> Trim/Extend Lines

You can currently cut a line. You can also edit it to extend it. But maybe not in the way you want, please tell us more

> Draw line segments with a set arc (like: 90° to the first drawn segment...)

 You can do it with the constraints part which become visible when you activate the measure tool

meyerlor:
Ah, with digitizing tool you mean the drawing tool? I never used it much yet, will have to look more into it! (One main reason beeing that no snapping to existing features is available if i'm not wrong?)

Just for further clarification what i mean with trim/extend: This [QGIS tool](https://docs.qgis.org/3.40/en/docs/user_manual/working_with_vector/editing_geometry_attributes.html#trim-extend-feature) - so more of a "autoextend to the other selected line(segment)"-function. Or for the trim part: cut one line at the intersection with the other line.

Tomorrow i will play around more with the drawing tool :)

meyerlor:
Small update here: I'm working on it, but it's quite a big task - i got it working so far with all the tools (and some more) but still struggling to update/create the necessary tests.

Do you have any plans to migrate those other OL2 parts anytime (appart from editign/drawing/measuring) to OL6?

PR #6528, opened 2026-02-04
Fix Overlap on #lizmap-search Close Button and Items in map.css

Antoviscomi:
This PR addresses the overlap issue strictly within the search component. Specifically, it manages the spatial relationship between the ` #lizmap-search-close button` and the list items inside `#lizmap-search ul.items.`

Unlike previous attempts, this is a new, solid targeted approach: by applying `padding` to the search container itself, we ensure that the search result items never collide with the `search-specific close button`. **This fix is fully encapsulated and does not affect any other close buttons or absolute elements in the Lizmap interface (such as the attribute table).**

Key Changes
Container-Level Clearance: Added a structural `padding-top: 45px` to the `#lizmap-search container`. This creates a dedicated "Safe Zone" for the `absolute-positioned close button`, ensuring it never overlaps with the actual list content.

Structural Block Displacement: Forced the `strong` tag within search results to behave as a block element. This ensures that result titles occupy a solid, predictable space, making them reliable targets for E2E interaction.

Z-Index Hierarchy: Refined the stacking order between the search container (1040) and the close button (1041) to maintain clear layering.

### DIFF
--- a/lizmap/www/assets/css/map.css
+++ b/lizmap/www/assets/css/map.css
@@ -3229,13 +3229,10 @@ lizmap-mouse-position > div.coords-unit > select{
 #lizmap-search.open{
   display:block;
 }
-
+  
+ /* STRUCTURAL SHIFT: Forcing the strong tag as a block element ensures a solid, predictable layout for the result titles. */
+  
 #lizmap-search ul.items li > strong {
-  
-  /* STRUCTURAL SHIFT: Forcing the strong tag as a block element 
-     ensures a solid, predictable layout for the result titles.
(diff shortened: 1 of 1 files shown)


rldhont:
Thanks @Antoviscomi for your efforts on fixing this issue.

@mdouchin can you take a look ?

Antoviscomi:
> Thanks @Antoviscomi for your efforts on fixing this issue.
> 
> @mdouchin can you take a look ?

It's of course a pleasure @rldhont i hope this way can works smootly

PR #6595, opened 2026-02-21
fix: edition form – ValueMap and Checkbox show NULL when editing a feature

meyerlor:
## Problem

When editing an existing feature, Value Map dropdowns and Checkbox
widgets display NULL/no selection instead of the feature's actual value.

EMptyValues1

Two root causes:

1. **PHP 8.1+ PDO type changes** – PDO now returns native PHP types
   (int, float, bool) instead of strings by default
   (ATTR_STRINGIFY_FETCHES defaults to false). The strict `===`
   comparison in `fillSelect()` and the `!=` comparison in
   `jFormsControlCheckbox::setData()` both expect strings, so
   native-typed values never match.

2. **PostgreSQL numeric/decimal precision** – numeric columns with a
   defined scale (e.g. `numeric(10,2)`) return values with trailing
   zeros (`"0.40"`), while QGIS ValueMap keys are stored without them
   (`"0.4"`). The strict comparison therefore fails.

## Fix

In `QgisForm::setFormDataFromFields()`, normalize the DB value before
passing it to `setData()`:

- Cast non-string values to string
- Strip trailing decimal zeros from numeric strings

EMptyValues2

Backport to 3.9 would be highly appreciated!

### DIFF
--- a/lizmap/modules/lizmap/lib/Form/QgisForm.php
+++ b/lizmap/modules/lizmap/lib/Form/QgisForm.php
@@ -652,6 +652,13 @@ class QgisForm implements QgisFormControlsInterface
                         }
                     }
                 }
+                // Cast to string so that form widget comparisons work correctly.
+                // PDO in PHP 8.1+ returns native types (e.g. integers) instead of
+                // strings, which breaks strict === comparisons in fillSelect() and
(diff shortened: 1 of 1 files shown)


rldhont (approved):
Thanks @meyerlor

PR #6598, opened 2026-02-24
Add project announcements feature

meyerlor:
adds a project announcements feature that allows admins to show popup messages to map users when they open a project. Could be useful for things like maintenance notices, new data updates or feature announcements. To be honest, I'm not 100% sure if this feature is really necessary as a built-in thing. You could basically achieve something similar by just putting the announcement text on the landing page content. The main advantage here is the targeting (specific project/repository/group) and the "show only X times" logic - but whether this justifies the added complexity is a question for you to decide.

  What is included:
  - Admin UI to create/edit/delete announcements (under LizMap menu)
  - Targeting by repository, project and user groups
  - Configurable max display count per user (so the popup is not shown every time)
  - Frontend modal popup using the existing #lizmap-modal
  - DB schema for PostgreSQL and SQLite, with upgrade migrations

  Test plan

  - Fresh install: tables and ACL rights are created automatically
  - Upgrade from 3.9+: migration creates tables and assigns rights to admins
  - Create announcement in admin panel, verify it shows as modal on matching project
  - Verify group targeting works correctly
  - Verify max_display_count stops showing after X views
  - Test with anonymous users (sessionStorage fallback)

  ---
  Want me to create it like this, or should I adjust something / should i stop working on it?

This is a small demonstration:
Announcement

When recording it, i just saw that it might also be an idea to faciliate the "Warnings" PopUp functionallity and extend it a little bit? Just brainstorming here.


meyerlor:
unstale. not sure how to continue though

mind84:
@meyerlor,

I think this feature can be very useful.

@rldhont @mdouchin what do you think about it?

Cheers!

PR #6603, opened 2026-02-27
Auto-activate snapping via snap_on_start config

meyerlor:
## Summary
- Auto-activate snapping when editing starts if the layer has `snap_on_start: "True"` in the cfg
- Defers WFS snap requests until map tiles finish loading so QGIS Server doesnt get blocked on large projects
- Shows a info message for 7 seconds when snapping activtes
- Legacy configs without the key wont auto-activate, so backwards compatible

Selection tool changes moved to #6614 as requested by @rldhont. The companion plugin PR for the `snap_on_start` checkbox is here: https://github.com/3liz/lizmap-plugin/compare/master...meyerlor:lizmap-plugin:feature/snap-on-start

## Test plan
Existing snap E2E test has been updated. Could probably add a test for a layer without `snap_on_start` to verify it doesnt auto-activate, and maybe one for the deferred loading behaviour on larger projects.

### DIFF
--- a/assets/src/modules/Snapping.js
+++ b/assets/src/modules/Snapping.js
@@ -7,7 +7,7 @@
 
 import { mainEventDispatcher } from '../modules/Globals.js';
 import Edition from './Edition.js';
-import { MapRootState } from './state/MapLayer.js';
+import { MapLayerLoadStatus, MapRootState } from './state/MapLayer.js';
 import { TreeRootState } from './state/LayerTree.js';
 
 /**
@@ -39,6 +39,8 @@ export default class Snapping {
         this._snapEnabled = {};
         this._snapToggled = {};
         this._snapLayers = [];
+        this._snapOnStart = false;
+        this._pendingMapReadyListener = null;
 
         // Create layer to store snap features
         const snapLayer = new OpenLayers.Layer.Vector('snaplayer', {
@@ -136,6 +138,9 @@ export default class Snapping {
                                     'snap_segments_tolerance': editionLayerConfig.hasOwnProperty('snap_segments_tolerance') ? editionLayerConfig.snap_segments_tolerance : 10,
                                     'snap_intersections_tolerance': editionLayerConfig.hasOwnProperty('snap_intersections_tolerance') ? editionLayerConfig.snap_intersections_tolerance : 10
                                 };
+
+                                this._snapOnStart = editionLayerConfig.hasOwnProperty('snap_on_start')
+                                    && editionLayerConfig.snap_on_start === 'True';
                             }
                         }
                     }
@@ -161,6 +166,12 @@ export default class Snapping {
                         this._setSnapLayersVisibility,
                         ['layer.visibility.changed','group.visibility.changed']
                     );
+
+                    // Auto-activate snapping if configured (snap_on_start).
+                    // Legacy configs without the key do not auto-activate.
+                    if (this._snapOnStart) {
+                        this._activateWhenMapReady();
+                    }
                 }
             },
             'edition.formDisplayed'
@@ -169,6 +180,13 @@ export default class Snapping {
         // Clean snap when edition ends
         mainEventDispatcher.addListener(
             () => {
+                // Remove pending map-ready listener if still waiting
+                if (this._pendingMapReadyListener) {
+                    this._rootMapGroup.removeListener(
+                        this._pendingMapReadyListener, 'layer.load.status.changed'
+                    );
+                    this._pendingMapReadyListener = null;
+                }
                 this.active = false;
                 this._lizmap3.map.getLayersByName('snaplayer')[0].destroyFeatures();
                 this.config = undefined;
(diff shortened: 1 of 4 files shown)


mind84:
Hi @meyerlor!

One thing to consider about automatically enabling the snapping feature is that when activated, the system makes a **WFS** call for each active/visible layer in the map to retrieve the feature geometries. The call is repeated if the map extent is changed.
For large projects with many layers, this could initially slow down usability. We also have situations where we're asked to automatically enable snapping, and currently we do so with a simple custom js, case by case.

Probably the best way is to leave to the admin the choice by adding some kind of options to the Lizmap plugin but sounds to me a bit overkill.

What you think?

Cheers!

meyerlor:
Hey, @mind84 thanks for that feedback - I tested it just now on a project where i have 120+ layers and all of them are published as WFS - and you are right, it indeed slows rendering down quite a bit! But just to clarify, the WFS is only requested for layers where the edited layer can snap on to (& are visible), not all of them (or am i wrong here?).
Also, editing form takes a while longer to open. 

I tried something in the last commit: Fire the WFS request after map rendering is finished so QGIS server does not get blocked with concurring requests - basically mimicking what a user does. This made rendering times and general usability much smoother. Do you think this might be a path to go forward?

Plan B would be to introduce in the Edition Config of the plugin a simple checkbox "Enable Snapping by default" per layer - that would certainly doable if you would feel more comfortable with that?

Plan C would to implement both :D 

...And Plan D: Per project config via JS

meyerlor:
I think the 3.40 tests are quite flaky and failling is not related to the PR?

mind84:
Hi @meyerlor,

> But just to clarify, the WFS is only requested for layers where the edited layer can snap on to (& are visible), not all of them (or am i wrong here?)

You are right

> I tried something in the last commit

Seems good to me. To clarify, I don't have a strong opinion on that, to me it's just relevant that the "large project" scenario is taken into account, as it has been. That said:

- Probably if a project has more than 100 layer, not all of these layer are visible/needed while editing
- WFS calls are issued wit BBOX param, so requests do not involve all features
- Snapping 100+ layers comes at a cost, and generally, this cannot be avoided


> Plan B

While this could easily do the job, IMHO it increases the complexity of the config setup and I don't think it's worth it for a case like this.

Cheers!

mdouchin (changes requested):
Thanks for this proposal.

I am quite sure some users will have issues with big data if the snapping is automatically activated, but I agree this is better for most cases. I always imagine parcels of a big area (such as 1 000 000 features)...

No strong opinion, about this, perhaps the configuration option would be a good idea. I can propose a new checkbox in the plugin.

For the selection part (activate the rectangle), could you also display a message such as "The selection tool is activated, please draw a rectangle on the map to trigger the selection of features" (better wording needed...). I think you can use the `lizMap.addMessage` function to do so

meyerlor:
@mdouchin depending how busy you are - i have some time this weekend and could work on introducing a checkbox for the lizmap-plugin. 

The message when starting the selection tool sounds good - should i also add one for the snapping? When an editing session starts and snapping is enabled via config as default, a displayed message to inform the user might make sense then as well.

mdouchin:
> @mdouchin depending how busy you are - i have some time this weekend and could work on introducing a checkbox for the lizmap-plugin.

I am quite busy, but not confortable with the fact you work during your weekends. We could address this plugin modification next week if you prefer.

> The message when starting the selection tool sounds good - should i also add one for the snapping? When an editing session starts and snapping is enabled via config as default, a displayed message to inform the user might make sense then as well.

I think everything which helps the user is a progress 👌
You could use the 4th parameter to make the message disappear automatically after 10 seconds. Something like :

```javascript
let previousMessage = document.getElementById('lizmap-editing-message');
if (previousMessage) previousMessage.remove();			
lizMap.addMessage('SOME MESSAGE DISPLAYED TO THE USER, 'info', true, 10000).attr('id','lizmap-editing-message');
```

meyerlor:
@mdouchin hehe do not worry - it became kind of a passion to bring new features to lwc - and with the FOSSGIS2026 approaching i'm eager to show potentially interested people there what LWC is capable of! (also i only work 80% and have fridays always off - so work life balance is a thing i care about alot 😀 )

The auto-disapearing is a good idea - i will play with the timings a bit!

meyerlor:
I think 7sec is good, what do you think? We get now two messages for editing - the one which was already there + mine. Should we merge them into one IF both would be shown?

Below the demonstration of plugin + lwc:

Auto-Snap

mdouchin:
> Should we merge them into one IF both would be shown?

Not really needed IMHO. Thanks for your work !

mdouchin (approved):

mdouchin:
@rldhont Ok for you to merge ?

rldhont:
@meyerlor i have planned to merge it today. Why did you close it and remove your branch ?

meyerlor:
holy, that was a mistake - i was cleaning out my branches and misclicked. on it!

PR #6609, opened 2026-03-02
[Feature] Zoom the map to the full extent of the filtered features

mind84:
A new button has been added to the attribute table that allows you to zoom the map to the full extent of the filtered features

**SPECS**

- The extent is calculated on the totality of filtered features, ignoring the pagination
- It is not active on *non-geometric* layers

[vokoscreenNG-2026-03-02_10-09-54.webm](https://github.com/user-attachments/assets/5f066dd4-4da0-409e-825d-9611665ac5e5)

**NOTES**

The calculation of total extent is done in a very verbose way. Unfortunately I haven't been able to find a more reliable solution or a QGIS server method that supports this calculation the way I intended. I'd be happy to replace the calculation part with a more concise method, if you can suggest one. In any case, for now, I placed the calculation directly within the datatables endpoint, but it can be extracted in a separate function/file to create a reusable method, if needed.

Funded by Etra

**EDIT**

I realized the video might be misleading. Zoom works even without selecting the features; in the video, they've been selected to highlight the functionality.

### DIFF
--- a/assets/src/legacy/attributeTable.js
+++ b/assets/src/legacy/attributeTable.js
@@ -1070,37 +1070,57 @@ var lizAttributeTable = function() {
                 // and then fit the map extent accordingly. The fit follows OL View settings.
                 // The calculation is performed on all filtered features, ignoring pagination
                 $('#attribute-layer-'+ cleanName + ' button.btn-fit-filtered-extent').click(async function(){
+                    let targetExtent = [];
                     var tableName = attributeLayersDic[ $(this).val() ];
                     if (!DataTable.isDataTable(`#attribute-layer-table-${tableName}`)) return;
 
                     const dtTable = new DataTable.Api(`#attribute-layer-table-${tableName}`);
-                    const datatablesUrl = globalThis['lizUrls'].wms.replace('service', 'datatables/filteredFeaturesExtent');
-                    const params = {...globalThis['lizUrls'].params};
-                    params['layerId'] = config.attributeLayers[cleanName].layerId;
+                    const dtParams = dtTable.ajax.params();
+                    const hasFilter = typeof dtParams == 'object' &&
+                        'searchBuilder' in dtParams &&
+                        'criteria' in dtParams.searchBuilder &&
+                        Array.isArray(dtParams.searchBuilder.criteria) &&
+                        dtParams.searchBuilder.criteria.length > 0
+                    if (!hasFilter) {
+                        const layerConfig = lizMap.config.layers[$(this).val()];
+                        if (layerConfig && 'extent' in layerConfig && 'crs' in layerConfig) {
+                            targetExtent = lizMap.ol.proj.transformExtent(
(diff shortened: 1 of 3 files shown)


mdouchin:
Thanks for your proposal. I agree with you parsing the bbox of all features is not the best way, but there is no other way around at present which will respect the different filters. We cannot use the Lizmap expression methods to get this data (not possible yet)

Have you tried with different versions of QGIS Server and different providers ? @rldhont told me the bbox could be absent in the features properties.

mind84:
Hi @mdouchin,

> Have you tried with different versions of QGIS Server and different providers.

QGIS Server 3.40 with pgsql and shp providers. Are there other "critical" providers that should be tested?

mdouchin:
> QGIS Server 3.40 with pgsql and shp providers. Are there other "critical" providers that should be tested?

It seems for that for a SHP with geometry type "point", when we ask for "extent" and not for the full geometry, the bbox property is not returned. Perhaps for QGIS >= 3.44 ?
See https://github.com/3liz/lizmap-web-client/pull/6425

mind84:
@mdouchin,

I handled the case where the geometry is a point and adapted the tests to include also polygonal features as well,to cover both cases.

Now it's more verbose than ever :smile: 

It can probably be improved, but for now I've kept it this way to make it easier to read the calculation logic.

mdouchin:
Thanks
Last question : what's happening when no filter has been applied ? I think we should zoom to the layer extent is possible

mind84:
> Thanks Last question : what's happening when no filter has been applied ? I think we should zoom to the layer extent is possible

I confirm that the system zooms in on the currently filtered features. If no filter is applied, all features are considered in the calculation, so it essentially corresponds to zooming to the entire layer extent. The tests also cover this case.

Possibly, to avoid making server calls in this case, we could directly zoom to the layer extent using the traditional method.

Thanks!

mdouchin:
> Possibly, to avoid making server calls in this case, we could directly zoom to the layer extent using the traditional method.

I think this is needed to avoid bad performance for big datasets. Thanks ! 
See for example https://github.com/3liz/lizmap-web-client/blob/master/assets/src/legacy/switcher-layers-actions.js#L543 
@rldhont Is there a dedicated public method to zoom to a given layer extent ?

mind84:
> See for example https://github.com/3liz/lizmap-web-client/blob/master/assets/src/legacy/switcher-layers-actions.js#L543

Yep, that's what I'm talking about

mind84:
> I think this is needed to avoid bad performance for big datasets.

Done in last commit

rldhont (changes requested):

mind84:
> Thanks @mind84 ! Can you add tests in `requests-datatables.spec.js` ?
> 
> You will be able to test returned errors or empty final extent.

Done

rldhont (approved):

PR #6613, opened 2026-03-07
Rewire copy&paste functionallity to old "Copy Geometry" button, remove newly introduced button

meyerlor:
Remove separate CopyGeometryButton and wire copy&paste functionality into the existing lizmap-paste-geom component. The button now activates the geometry copy handler which enables the user to pick a feature from the map and applies it to the current editing layer in one step.

The old button had no working functionallity - removed the code artifacts. i hope this is ok?


mdouchin:
I am a bit lost here. Could you please add a animated GIF to demonstrate the changes ? Or 2 before/after screenshots ?
Thanks in advance

meyerlor:
Ah so sorry - really badly explained from my side! With introducing the copy & paste functionallity, I added a new button in the editing form:
grafik


With this PR, I removed the newly introduced button again so it now uses the button from before (which had no real functionality so far):
grafik

Or should we keep both ?

In any case, the size of the button is not good, will fix it once it's clear how to continue! :)

mdouchin:
Thanks for your explanation.

Ok to keep only one button, dedicated to the new behaviour.
But we must use another icon IMHO, since it is not a "paste" action anymore.

The icon can be something like [map-pin-add-fill in ](https://remixicon.com/icon/map-pin-add-fill) 
Not sure this one is better, but I would like to find a simple icon saying "find a geometry and use it". Since the action is more "get a geometry from the map layers by clicking" than "paste the copied geometry"

By the way, the old behaviour had been "thought" to be able to copy geometries from other tools, such as drawing or selection tool, the popup toolbar, lizmap-search result, etc. and then paste it elsewhere (editing form, drawing, etc.). 
We did not finish the set of tools (for example by adding "copy" buttons in the interface), but it was more like "Copy a geometry from anywhere possible and use it elsewhere", whereas your tool is more dedicated "Drill-down and get the geometry among the one under the click"
I think this behaviour could also be great to have in the future

meyerlor:
I agree, none of the used symbols matches the behaviour - I will search for a better one as replacement and post suggestions here when I find something promising.

Oh, this functionallity would also cool to have - if some kind of WKT geometry parser is included, we could also paste geoms from QGIS! That would be awesome. How far on top is it on your internal backlog? :)

meyerlor:
I asked a friend oft mine (media design student) to come up with something:
lorenz_icon

What do you think? He can change it if needed

mdouchin:
> I asked a friend oft mine (media design student) to come up with something
> What do you think? He can change it if needed

This is great, you can thank your friend. I have not tried it in a small size though.

meyerlor:
ah, i just uploaded via mobile last night - i'm away this weekend, will try to include into a build soonish and check how it looks in "button-size"

meyerlor:
here as 24x24 as it woul dbe in the sprite: 
lorenz_icon

meyerlor:
> By the way, the old behaviour had been "thought" to be able to copy geometries from other tools, such as drawing or selection tool, the popup toolbar, lizmap-search result, etc. and then paste it elsewhere (editing form, drawing, etc.). We did not finish the set of tools (for example by adding "copy" buttons in the interface), but it was more like "Copy a geometry from anywhere possible and use it elsewhere", whereas your tool is more dedicated "Drill-down and get the geometry among the one under the click" I think this behaviour could also be great to have in the future

I reinstated the previous code, but removed the button for it - so it can be worked on later.

This is how the button looks now:

CopyPast2

mdouchin:
Ok for me, could you just replace the tooltip message `Copy and paste an existing geometry` with `Copy the geometry from an existing map layer feature` if it is ok for you ? IMHO it is more precise

meyerlor:
sure!

mdouchin (approved):
Thanks a lot !

PR #6615, opened 2026-03-08
Auto-activate snapping via snap_on_start config

meyerlor:
## Summary
- Auto-activate snapping when editing starts if the layer has `snap_on_start: "True"` in the cfg
- Defers WFS snap requests until map tiles finish loading so QGIS Server doesnt get blocked on large projects
- Shows a info message for 7 seconds when snapping activtes
- Legacy configs without the key wont auto-activate, so backwards compatible

Split from #6603 as requested by @rldhont. The companion plugin PR for the `snap_on_start` checkbox is here: https://github.com/3liz/lizmap-plugin/compare/master...meyerlor:lizmap-plugin:feature/snap-on-start

## Test plan
Existing snap E2E test has been updated. Could probably add a test for a layer without `snap_on_start` to verify it doesnt auto-activate, and maybe one for the deferred loading behaviour on larger projects.

### DIFF
--- a/assets/src/modules/Snapping.js
+++ b/assets/src/modules/Snapping.js
@@ -7,7 +7,7 @@
 
 import { mainEventDispatcher } from '../modules/Globals.js';
 import Edition from './Edition.js';
-import { MapRootState } from './state/MapLayer.js';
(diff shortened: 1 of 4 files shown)


meyerlor:
Duplicate - snapping changes are in #6603 instead.

PR #6628, opened 2026-03-09
Feature/single wms layer opacity

meyerlor:
When "Load layers as single WMS request" was enabled, opacity controls were hidden and users couldn't adjust transparency for individual layers.

To overcome this:
  - Added the OPACITIES WMS parameter to the single WMS GetMap request, letting QGIS Server composite layers at their correct opacities server-side
  - Registered opacity change event listeners in SingleWMSLayer so changing opacity triggers a new WMS request
  - Removed the UI guard that hid opacity buttons for single WMS layers
  - Only sends OPACITIES when at least one layer differs from full opacity (255) to avoid 400 errors
  - Works for baselayers as well
 
Fixes: https://github.com/3liz/lizmap-web-client/issues/6061

### DIFF
--- a/assets/src/legacy/switcher-layers-actions.js
+++ b/assets/src/legacy/switcher-layers-actions.js
@@ -148,26 +148,24 @@ var lizLayerActionButtons = function() {
                 }
             }
             // Opacity
-            {
-                html+= '        <dt>'+lizDict['layer.metadata.opacity.title']+'</dt>';
-                html+= '<dd>';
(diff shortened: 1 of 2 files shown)


mind84 (commented):
I would consider adding a simple e2e test that takes into account the WMS `OPACITY` request, as done for example in the "Change layer style" test

meyerlor:
ah, good idea - added one!

PR #6601, opened 2026-02-26
[Bug] Re-enable the filtered feature selection functionality in the attribute table

mind84:
Here the functionality restored:

[vokoscreenNG-2026-02-26_10-46-03.webm](https://github.com/user-attachments/assets/408f966d-60d7-42d9-b39a-f82ada1c5692)

**NOTE ON TESTS**
The tests can be refined by adding filters using the search builder, but most of the methods for interacting with this component are contained in this PR https://github.com/3liz/lizmap-web-client/pull/6572. I would therefore proceed to integrate them after the latter's merge or, in any case, once it has been decided what to do with it.

Thanks!

Backport 3.10

Funded by Faunalia

PR #6632, opened 2026-03-10
[Feature] Select all filtered features server side

mind84:
Following  https://github.com/3liz/lizmap-web-client/pull/6601, this PR aims to follow the logical thread of server-side rendering of the attribute table. When a user wants to select all the filtered features, logically, in my opinion, this also means selecting the records that aren't rendered on the current page.

**FACT**

This can inevitably lead to the selection of many records. I believe this is the correct behavior from a logical standpoint. What can be done is to control the interface behavior so that it's possible to prevent the selection of many records if they exceed a certain (arbitrary) number, or to prompt the user for confirmation before proceeding with the selection.

What do you think?

Backport 3.10

Funded by Faunalia

### DIFF
--- a/lizmap/modules/lizmap/controllers/datatables.classic.php
+++ b/lizmap/modules/lizmap/controllers/datatables.classic.php
@@ -47,12 +47,12 @@ class datatablesCtrl extends jController
     private $DTLength;
 
     /**
-     * @var null|array Array of columns order
+     * @var mixed Array of columns order
      */
     private $DTOrder;
 
     /**
-     * @var null|array Array of columns name
+     * @var mixed Array of columns name
      */
     private $DTColumns;
 
--- a/lizmap/modules/lizmap/controllers/datatables.classic.php
+++ b/lizmap/modules/lizmap/controllers/datatables.classic.php
@@ -12,6 +12,7 @@
  */
 
 use GuzzleHttp\Psr7\StreamWrapper as Psr7StreamWrapper;
+use GuzzleHttp\Psr7\Utils as Psr7Utils;
 use JsonMachine\Items as JsonMachineItems;
 use Lizmap\DataTables\DataTables;
 use Lizmap\Project\UnknownLizmapProjectException;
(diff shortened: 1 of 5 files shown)


mdouchin:
> Following #6601, this PR aims to follow the logical thread of server-side rendering of the attribute table. When a user wants to select all the filtered features, logically, in my opinion, this also means selecting the records that aren't rendered on the current page.

+1 As a user, the filter is applied to the layer, not only the attribute table visible page

> **FACT**
> 
> This can inevitably lead to the selection of many records. I believe this is the correct behavior from a logical standpoint. 

I agree

> What can be done is to control the interface behavior so that it's possible to prevent the selection of many records if they exceed a certain (arbitrary) number, or to prompt the user for confirmation before proceeding with the selection.
> 
> What do you think?

It would be great to know in advance the number of features to display a message only if the number exceed 1000 or 10k features.
I do not think there is a performant way to count the layer features with a expression. If so, we could use the LWC requests to evaluate the expression before triggering the action.

mind84:
> It would be great to know in advance the number of features to display a message only if the number exceed 1000 or 10k features.

We can use the information returned from the datatable endpoint

image

Since the selection is always applied to the current results set, the `recordsFiltered` property should contain a reliable information, even if no filter is applied.

Am I wrong?

Thanks!

mdouchin:
> Since the selection is always applied to the current results set, the `recordsFiltered` property should contain a reliable information, even if no filter is applied.

Yep, I forgot this. This will do the job !

mind84:
@mdouchin,

I'm thinking about a prompt like this:

_You are about to select more than a 1000 features. This may take a few seconds to process. Are you sure you want to proceed?_

What do you think?

Thanks!

P.S. the number of rows is a value stored directly in the `datatable` object instance and it's coming from the datatable response, so I probably use directly that value to perform this check.

mdouchin:
> I'm thinking about a prompt like this:
> 
> _You are about to select more than a 1000 features. This may take a few seconds to process. Are you sure you want to proceed?_

+1
 
> P.S. the number of rows is a value stored directly in the `datatable` object instance and it's coming from the datatable response, so I probably use directly that value to perform this check.

Yep, seems the simplest solution.

Thanks !

lizmap/modules/lizmap/controllers/datatables.classic.php
```diff
@@ -41,94 +102,112 @@ protected function setErrorResponse(jResponseJson $rep, int $code, string $error
...
+     * @return null|string
+     */
+    private function getExpFilter()
+    {
+        $expFilter = $this->param('exp_filter');
+        if (count($this->filteredFeatureIDs) > 0) {
+            $filteredFeatureIDSFilter = '$id IN ('.implode(' , ', $this->filteredFeatureIDs).')';
+            // concat current exp_filter with filteredFeaturesIds filter
+            $expFilter = !$expFilter ? $filteredFeatureIDSFilter : "( {$expFilter} ) AND ( {$filteredFeatureIDSFilter} )";
+        }
```
mdouchin: We could use a map function to check the ids are integers. We cannot trust parameters sent by Javascript code.
If possible, we should add more security checks when using parameters. 
NB: I know you are not the original author of this file.

We could create another PR to improve this if you prefer ?

mdouchin (changes requested):
I left some comments about security, we could add more controls in another PR if you prefer

rldhont:
@mdouchin changes have been done, can you made a new review ?

rldhont (approved):

PR #6635, opened 2026-03-12
Fix print layout order to respect cfg configuration

meyerlor:
printTemplates was built by iterating printCapabilities in QGIS XML order, ignoring the user-configured order in layouts.list from the cfg file. Since the frontend matches these arrays by index, the print dropdown showed layouts in QGIS's internal order instead of the order set in the Lizmap plugin.

Reorder printTemplates to match the layouts.list order from the cfg.

### DIFF
--- a/lizmap/modules/lizmap/lib/Project/Project.php
+++ b/lizmap/modules/lizmap/lib/Project/Project.php
@@ -1984,6 +1984,8 @@ class Project
         $server = new Server();
         $serverMetadata = $server->getMetadata();
         $serverPlugins = $serverMetadata['qgis_server_info']['plugins'];
+        // Build a lookup of print templates by title
+        $printTemplatesByTitle = array();
         foreach ($this->printCapabilities as $printTemplate) {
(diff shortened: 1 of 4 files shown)


rldhont:
Can you add a test in `requests-getprojectconfig.spec.js` to check the template order ?

Can you also update `print.spec.js` to check the order ?

meyerlor:
In the .qgs XML and the .cfg both are in the same order - should a create copies of the project file and the cfg and change the order in the cfg?

Edit:
worked around it.
  - .cfg: print_overview moved from last to second position (before print_map), creating a mismatch between cfg order and the QGIS XML order
  - Tests: toContainText (unordered) replaced with toHaveText (strict order) for template assertions, selectOption indices updated to match new positions, fixed some stale comments

meyerlor:
ugh, so many references to update. now it's passing.

rldhont:
Thanks @meyerlor

rldhont (approved):

PR #6640, opened 2026-03-13
[Feature] Double clicking on layer legend propagates the checked state on symbology checkbox

mind84:
This PR is highly inspired by this one https://github.com/3liz/lizmap-web-client/pull/4179

It basically extends the double click functionality to the layer symbology.

[vokoscreenNG-2026-03-13_17-17-26.webm](https://github.com/user-attachments/assets/509feb39-c84a-475d-a26d-a2fe1a755bcf)

**SPECS**

- Double-clicking groups doesn't trigger the on/off state propagation to the underlying layer symbology. From a user perspective, I believe this is the most frequently requested behavior.
- It works recursively, traversing all symbology subtrees
- Double-clicking a symbol group doesn't propagate the on/off state to its child symbols. The feature is designed to act simultaneously on all layer symbols.
- Works with `singleWMS` mode as well

Backport 3.10

Funded by Etra


rldhont (changes requested):
Thanks for this contribution! 

Once `propagateCheckedState` is fixed in `Symbology` classes, can you add js unit tests in `symbology.test.js` and `layerTree.test.js` ? Thanks.

mind84:
Hi @rldhont,

I hope I understood what you meant by the comments in the review. With a fairly significant class inheritance mechanism, it's easy to get lost :smiley: 

In any case, let me know if there are any things I should add!

Thanks!

rldhont (approved):

PR #6683, opened 2026-03-31
Fix/atlas print icon missing sprite viewbox

meyerlor:
viewBox mismatch - for atlas print in a feature form, the symbol was gone / out of viewport


rldhont:
@meyerlor can you provide a printscreen ?

meyerlor:
Current master:
grafik


Fixed:

grafik



For backporting to 3.9 this pr alone is not enough:
Two files need changing:

  1. lizmap/www/assets/svg/sprite.svg
  The map-print symbol was given the wrong viewBox by svgo:

  2. assets/src/components/FeatureToolbar.js + JS rebuild
  The backport kept the old xlink:href attribute, which Chrome 88+ no longer supports for external SVG sprite
  references. All other components already use href:
Lines 140:
```
-<use xlink:href="${lizUrls.svgSprite}#map-print"></use>
+<use href="${lizUrls.svgSprite}#map-print"/>
```
  After that change, npm run build must be run to regenerate the JS bundle.

  The sprite.svg fix alone (as in the master PR) is not sufficient on 3.9 because of this second issue.

rldhont (approved):

mdouchin (approved):

rldhont:
Thanks @meyerlor to point the issue in release_3_9. I have fixed it https://github.com/3liz/lizmap-web-client/commit/c0a22d17437277399d9d606c1847c1f5f5c6ebef

Issue #6456, opened 2026-01-26
[Bug]: zoom to feature: inconsistent behavior between popup and table


josemvm:
### What is the bug? (in English)

when zooming to a point feature through a popup, the point is correctly centered on the map:

<img width="769" height="180" alt="Image" src="https://github.com/user-attachments/assets/1ba79c53-0d9f-471e-b87d-9923fcc06145" />

however, when zooming from the attribute table, the point appears slightly offset upwards from the center:
<img width="769" height="235" alt="Image" src="https://github.com/user-attachments/assets/ee77ba09-be4f-4231-a324-bd8c9fcbb5e3" />

### Steps to reproduce the issue

set map to 1:1 scale
zoom to the same feature via popup and via attribute table, and verify that the point's position on the map is not the same

### Versions, safeguards, check summary etc

**Versions :**

* Lizmap Web Client : 3.9.6-pre.9346 - commit 0679152b6 https://github.com/3liz/lizmap-web-client/commit/0679152b6
* Lizmap plugin : 4.5.7
* QGIS Desktop : 3.40.14
* QGIS Server : 3.40.14
* Py-QGIS-Server or QJazz : not used
* QGIS Server plugin atlasprint : 3.4.3
* QGIS Server plugin lizmap_server : 2.14.1
* QGIS Server plugin wfsOutputExtension : 1.8.3

<details>
<summary>List of Lizmap Web Client modules :</summary>
<br/>
* multiauth : 1.2.2
* altiProfil : 0.5.9
* altiProfilAdmin : 0.5.9
</details>

<details>
<summary>List of safeguards :</summary>
<br/>
* Mode : normal<br/>
* Allow parent folder : no<br/>
* Prevent other drive : no<br/>
* Prevent PG service : no<br/>
* Prevent PG Auth DB : no<br/>
* Force PG user&pass : no<br/>
* Prevent ECW : no<br/>
</details>

### Check Lizmap plugin

- [x] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

Ubuntu 24

### Browsers

Firefox

### Browsers version

147.0.1

### Relevant log output

```shell

```

rldhont:
I have created this PR https://github.com/3liz/lizmap-web-client/pull/6453

I am not able to reproduce your issue.

josemvm:
> I am not able to reproduce your issue.

@rldhont thank you for your response

[...]

PR #6684, opened 2026-04-02
Fix WFS PostGIS path SRSNAME handling and snap datum-shift error

meyerlor:
While working on the snapping migration to OL10, I just found a mismatch happening when snapping to layers which looks like a datum-shift error:

Snap_Mismatch

This PR fixes it, it is aimed for 3.9 and 3.10 - the Ol10 will handle snapping differently and i'll keep in mind to watch for this behaviour once i continue working on that for 3.11 (or 4.X ! :)  )

Snap_Match


mind84:
Hi @meyerlor 

I did some quick tests, I tried to enable snapping on the following project:

repository=testsrepository&project=**form_edition_value_relation_field** 

The snap is only on the `list` layer :

image

but the server returns an incorrect bbox/geometries for the `list` layer.

image

Project is in `3857`, `list` layer is in `4326`


Am I doing something wrong? Can you replicate the problem?

Thanks!

meyerlor:
@mind84 you are right — none of the test projects use a datum-shifted CRS, so the e2e test checks only the protocol change (WFS 1.1.0 + SRSNAME) rather than the coordinate values directly. Adding a project with e.g. EPSG:28992 or EPSG:25832 (my usual one) would also require NTv2 grid files on the test QGIS Server. The test only proves the new code path is taken, which eliminates the client-side proj4js transform by construction — the coordinate accuracy guarantee comes from QGIS Server/PROJ once SRSNAME is present.

mind84:
Hi @meyerlor,

I took a quick look, so please correct me if I'm wrong!

The main problem is that the WFS call is made to PostGIS for DB layers, see `getFeaturePostgres` method in `WFSRequest.php` class. The query that is made has a fixed CRS of `4326` for bboxes and geometries.

The following example shows the query generated with the new method:

```sql
WITH source AS (
         SELECT "id", "code", "label", "code_parent", "geom" AS "geosource" FROM "tests_projects"."form_edition_vr_list" WHERE True LIMIT 1000
        )
        --SELECT row_to_json(fc, True) AS geojson
        --FROM (
        --    SELECT
        --        'FeatureCollection' As type,
        --        array_to_json(array_agg(f)) As features
        SELECT row_to_json(f, True) AS geojson
            FROM (
                SELECT
                    'Feature' AS type,
         Concat(
            'list',
            '.',
            "id") AS id,
                        array_to_json(ARRAY[
                            ST_XMin(ST_Transform(ST_Envelope(lg.geosource::geometry), 4326)),
                            ST_YMin(ST_Transform(ST_Envelope(lg.geosource::geometry), 4326)),
                            ST_XMax(ST_Transform(ST_Envelope(lg.geosource::geometry), 4326)),
                            ST_YMax(ST_Transform(ST_Envelope(lg.geosource::geometry), 4326))
                        ]) As bbox,
            
                        ST_AsGeoJSON(ST_Transform(lg.geosource::geometry, 4326))::json As geometry,
            
                    row_to_json(
                        ( SELECT l FROM
                            (
                                SELECT "id", "code", "label", "code_parent"
                            ) As l
                        )
                    ) As properties
                FROM source As lg
            ) As f
        --) As fc
```

Geometries are always returned in 4326.

Also `bbox` is not considered, probably due to the BBOX sintax passed as param:

`-899310.3675109865,5331047.836042481,1412145.3675109865,6412173.163957519,EPSG:3857`

The CRS added as the fifth element seems to be a format not supported by QGIS Server (maybe is related to Geoserver?) -> **EDIT:** QGIS Server supports this format

Ignoring the BBOX causes all features to be requested, which can cause performance issues.

It would probably be necessary to modify this method to get a Postgis side conversion, or alternatively force the QGIS Server request with the WFS `force_qgis` parameter.

@rldhont @mdouchin what do you think?

Thanks!

meyerlor:
Hi @mind84, good catch — you're right on both. Both issues are in the PostGres direct path (getfeaturePostgres), which is what gets hit for PG-backed layers:

1. BBOX ignored: `getBboxSql` rejects anything that isn't exactly 4 comma-separated values (`line 751`). WFS 1.1.0 appends the CRS as a 5th element (x,y,x,y,EPSG:3857), so the check fails silently and no spatial filter is applied → all features get returned. That explain the perf issue you saw.

2. Geometry hardcoded to 4326: `Lines 1184–1185` always do `ST_Transform(..., 4326)` regardless of SRSNAME. Meanwhile the JS now assumes coordinates come back in map projection and skips reprojection — so on a 3857 map you get 4326 coords plotted as-is. Thats the mismatch in your screenshot.

Both are pre-existing limitation of the PG path.. Simplest fix imo: add `FORCE_QGIS: '1'` to the snap WFS call. That bypasses the direct SQL path and goes trough QGIS Server, which handle WFS 1.1.0 + SRSNAME + 5-element BBOX correctly and is also the right place for PROJ/NTv2 datum shifts anyway. Snap is already capped at 1000 features so the extra overhead doesn't really matter here.

Should I make that change and update the test? Fixing the PG path itself (BBOX parsing + SRSNAME awareness) would be worth doing but i'd keep that as a separate issue? I'm keen on migrating the whole snapping function to OL10 soonish.

I'll wait for @rldhont / @mdouchin ..! :)

rldhont:
@meyerlor you are just facing the fact that when QGIS Server returns GeoJSON in EPSG:4326, it truncates coordinates at 6 number after comma. Like I explain here : https://github.com/3liz/lizmap-web-client/issues/6456#issuecomment-3935901663

The solution is to request the GetFeature as GML or to add an SRSNAME and in PHP to force to QGIS Server when SRSNAME is provided.

mind84:
@meyerlor @rldhont 

I'm a little worried about bypassing PG in this case, it's a very useful method for increasing performance.

mdouchin:
> I'm a little worried about bypassing PG in this case, it's a very useful method for increasing performance.

I agree !

meyerlor:
Hi @rldhont @mind84 @mdouchin,
Thanks for the feedback. I worked on both issues raised — the BBOX parsing and the hardcoded EPSG:4326 — directly in the PostGIS path, so the performance benefit of the direct PG query is preserved.
  - BBOX with CRS (5th element): `getBboxSql `now recognises the WFS 1.1.0 format `x,y,x,y,EPSG:XXXX`. The CRS is extracted and the envelope is reprojected to the layer's native SRID via `ST_Transform(ST_MakeEnvelope(..., inputSrid), layerSrid).` An explicit `SRSNAME` parameter takes priority over the 5th BBOX element.
- Geometry output in requested `SRSNAME`: `setGeojsonSql` now transforms geometry and bbox to the requested output SRID instead of always using 4326.

Both changes have unit tests covering EPSG:4326, 3857, and 2154.

Is this ok to go forward like this?

My other assumptions about precision were wrong - i closed that PR. (PostGIS ST_AsGeoJSON defaults to 9 decimal digits regardless of CRS.

mind84:
Hi @meyerlor!

In my opinion, there are no significant issues with modifying the PG function to handle `SRSNAME`. I think it's worth considering that:

- Precision issues can't be completely avoided; the best solution is to reduce the geometry precision to a fixed grid.

- There are several methods in the code that perform WFS requests; some of these assume the returned geometry is in 4326 format, so it's probably worth doing a general consistency check on these methods.

Cheers

meyerlor:
Hi @mind84, thanks for bringin in that perspective!

Both of your points are fair, and I'd rather not widen the scope of this PR to adress them — I think an own issue / PR would be more apropriate?

Like:
- **Precision / fixed grid** — agreed. `ST_AsGeoJSON` already caps at 9 digits which is sub-mm for projected CRSes, but an explicit `ST_SnapToGrid` per
layer SRID would make output deterministic across reprojections. I can open a dedicated issue to it sooner or later.
- **Consistency across WFS callers** — also agreed. Several paths still implicitly assume 4326 output (e.g. `qgisVectorLayer::editableFeatures()`, `search`,
 `osm` controllers, plus a few JS consumers). I will start a new issue listing the suspects so we can work through them one at a time - both issues are not top-priority i think?

If that is ok with you, I guess this PR is kind of ready for a final pass. @rldhont @mdouchin — any remaining concerns?

rldhont (changes requested):
To verify that the changes you've made are correctly transforming the geometries, you can add tests in https://github.com/3liz/lizmap-web-client/blob/master/tests/end2end/playwright/requests-wfs.spec.js, thanks.

tests/end2end/playwright/snap.spec.js
```diff
@@ -8,14 +9,169 @@ test.describe('Snap on edition', () => {
...
+    test('Snap WFS GetFeature uses WFS 1.1.0 with SRSNAME in the map projection', async ({ page }) => {
+        const project = new ProjectPage(page, 'form_edition_multilayer_snap');
+
+        // Intercept the snap GetFeature request before opening the form —
+        // snapping is auto-activated on form display for this project.
+        const snapWfsRequestPromise = page.waitForRequest(
+            request => request.method() === 'POST'
+                && request.postData()?.includes('GetFeature') === true
+                && request.postData()?.includes('form_edition_snap_point') === true
+        );
```
rldhont: I invite you to enhance project's `waitForGetFeatureRequest` method to be able to pass a FEATURETYPE parameter like `waitForGetMapRequest` has a `LAYERS`  parameter.

tests/end2end/playwright/snap.spec.js
```diff
@@ -8,14 +9,169 @@ test.describe('Snap on edition', () => {
...
+        requestExpect(snapWfsRequest).toContainParametersInPostData({
+            SERVICE: 'WFS',
+            VERSION: '1.1.0',
+            REQUEST: 'GetFeature',
+            TYPENAME: 'form_edition_snap_point',
+            SRSNAME: 'EPSG:4326',  // the project's map projection
+        });
+
+        // Allow any in-flight requests to settle before checking the flag.
+        await page.waitForTimeout(300);
```
rldhont: Add a `responseExpect` test to check that `await snapWfsRequest.response()` is a GeoJSON. And you cwill be able to reduce the `waitForTimeout` with the certitude that the `DescribeFeatureType` request will not be send.

tests/end2end/playwright/snap.spec.js
```diff
@@ -8,14 +9,169 @@ test.describe('Snap on edition', () => {
...
+        const project = new ProjectPage(page, 'form_edition_snap_datum_shift');
+        project.waitForGetLegendGraphicDuringLoad = false;
+
+        await project.open();
+
+        const snapWfsRequestPromise = page.waitForRequest(
+            request => request.method() === 'POST'
+                && request.postData()?.includes('GetFeature') === true
+                && request.postData()?.includes('snap_datum_shift_target') === true
+        );
```
rldhont: Use your new `waitForGetFeatureRequest` method with `FEATURETYPE` parameter.

tests/end2end/playwright/snap.spec.js
```diff
@@ -8,14 +9,169 @@ test.describe('Snap on edition', () => {
...
+            REQUEST: 'GetFeature',
+            TYPENAME: 'snap_datum_shift_target',
+            SRSNAME: 'EPSG:3857',
+        });
+
+        // BBOX must be the 5-element WFS 1.1.0 format ending with the CRS code.
+        const bbox = postData.get('BBOX') ?? '';
+        expect(bbox, `BBOX must be 5-element WFS 1.1.0 format (x,y,x,y,EPSG:3857), got: "${bbox}"`
+        ).toMatch(/^-?[\d.]+,-?[\d.]+,-?[\d.]+,-?[\d.]+,EPSG:3857$/);
+    });
```
rldhont, line 97: Add a `responseExpect` test to check that `await snapWfsRequest.response()` is a GeoJSON.

tests/end2end/playwright/snap.spec.js
```diff
@@ -8,14 +9,169 @@ test.describe('Snap on edition', () => {
...
+        await project.open();
+
+        const snapWfsRequestPromise = page.waitForRequest(
+            request => request.method() === 'POST'
+                && request.postData()?.includes('GetFeature') === true
+                && request.postData()?.includes('snap_datum_shift_target') === true
+        );
+
+        const formRequest = await project.openEditingFormWithLayer('snap_datum_shift_edit');
+        await formRequest.response();
```
rldhont: Add a `responseExpect` to check that `await formRequest.response()` is an HTML.

tests/end2end/playwright/snap.spec.js
```diff
@@ -8,14 +9,169 @@ test.describe('Snap on edition', () => {
...
+        const project = new ProjectPage(page, 'form_edition_snap_datum_shift');
+        project.waitForGetLegendGraphicDuringLoad = false;
+
+        await project.open();
+
+        const snapWfsResponsePromise = page.waitForResponse(
+            response => response.request().method() === 'POST'
+                && response.request().postData()?.includes('GetFeature') === true
+                && response.request().postData()?.includes('snap_datum_shift_target') === true
+        );
```
rldhont: Use your new waitForGetFeatureRequest method with FEATURETYPE parameter.

tests/end2end/playwright/snap.spec.js
```diff
@@ -8,14 +9,169 @@ test.describe('Snap on edition', () => {
...
+
+        const formRequest = await project.openEditingFormWithLayer('snap_datum_shift_edit');
+        await formRequest.response();
```
rldhont, line 114: Check the response.

tests/end2end/playwright/snap.spec.js
```diff
@@ -8,14 +9,169 @@ test.describe('Snap on edition', () => {
...
+        const snapWfsResponse = await snapWfsResponsePromise;
+        const responseBody = await snapWfsResponse.text();
+
+        // Attach the full server response to the test report.
+        await testInfo.attach('snap-wfs-response-body', {
+            body: responseBody,
+            contentType: 'application/json',
+        });
+
+        const geojson = JSON.parse(responseBody);
```
rldhont:
```suggestion
        const geojson = await snapWfsResponse.json();
```

tests/end2end/playwright/snap.spec.js
```diff
@@ -8,14 +9,169 @@ test.describe('Snap on edition', () => {
...
+                && response.request().postData()?.includes('GetFeature') === true
+                && response.request().postData()?.includes('snap_datum_shift_target') === true
+        );
+
+        const formRequest = await project.openEditingFormWithLayer('snap_datum_shift_edit');
+        await formRequest.response();
+
+        await page.getByRole('tab', { name: 'Digitization' }).click();
+
+        const snapWfsResponse = await snapWfsResponsePromise;
```
rldhont: Check that the response is GeoJSON

tests/end2end/playwright/snap.spec.js
```diff
@@ -142,12 +296,10 @@ test.describe('Snap on edition', () => {
 
         let getSnappingPolygonFeatureRequestPromise = page.waitForRequest(
             request => request.method() === 'POST' && request.postData() != null && request.postData()?.includes('GetFeature') === true && request.postData()?.includes('form_edition_snap_polygon') === true);
-        let getSnappingPolygonDescribeFeatureRequestPromise = page.waitForRequest(
-            request => request.method() === 'POST' && request.postData() != null && request.postData()?.includes('DescribeFeatureType') === true && request.postData()?.includes('form_edition_snap_polygon') === true);
 
         await page.locator("#edition-point-coord-form-group").getByRole("button").nth(2).click()
 
-        await Promise.all([getSnappingPolygonFeatureRequestPromise, getSnappingPolygonDescribeFeatureRequestPromise])
+        await getSnappingPolygonFeatureRequestPromise;
```
rldhont: Wait for the response and check it to be able to remove `waitForTimeout`

rldhont (changes requested):
I have made a review of snap.spec.js.

Can you add a test in requests-wfs.spec.js, to check the ability to support SRSName for WFS 1.1.0 with Lizmap eb Client. You can reuse the request made by snap.

rldhont (approved):

PR #6699, opened 2026-04-06
fix: increase ST_AsGeoJSON precision to 9 decimal digits

meyerlor:
Fixes coordinate truncation in PostGIS WFS direct path. ST_AsGeoJSON defaults to 6 decimals for degrees (~11cm) and only 1 for meters (~10cm), which is insufficient for snapping accuracy. Using 9 decimals matches QGIS Server's own GeoJSON output precision.
Includes unit-test.

### DIFF
--- a/lizmap/modules/lizmap/lib/Request/WFSRequest.php
+++ b/lizmap/modules/lizmap/lib/Request/WFSRequest.php
@@ -1196,7 +1196,7 @@ class WFSRequest extends OGCRequest
 
             // Transform into GeoJSON
             $sql .= '
-                        ST_AsGeoJSON('.$geosql.')::json As geometry,
+                        ST_AsGeoJSON('.$geosql.', 9)::json As geometry,
             ';
         } else {
             $sql .= '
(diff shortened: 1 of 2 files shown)


u-cav:
Hi @meyerlor, I can't find any documentation about `ST_AsGeoJSON` having a low and variable `maxdecimaldigits` limit based on the unit of the SR. The official documentation states that the default value is already 9, does that sound right to you?

The limit for QGIS Server seems to be 6 https://github.com/3liz/lizmap-web-client/issues/6456#issuecomment-3935901663

Setting a lower maxdecimaldigits through PostGIS could also lead to invalid geometries (see https://postgis.net/docs/ST_AsGeoJSON.html).

Feel free to let me know if I'm missing anything!

meyerlor:
You're right — got confused here. PostGIS ST_AsGeoJSON already defaults to 9 decimal digits, so this PR is a no-op. The 6-digit truncation issue is specific to QGIS Server's GeoJSON output in EPSG:4326. PR #6684 fixes the PostGIS path by respecting SRSNAME and returning coordinates in the correct CRS (fixing a CRS mismatch, not a precision issue). As a side effect, requesting with SRSNAME from QGIS Server also returns projected coordinates instead of 4326, which avoids that truncation. Closing this.

PR #6733, opened 2026-04-13
Improve FTS search: ordering, result limits and 2-column display

meyerlor:
- Switch similarity() to word_similarity() so short search terms (e.g. "106") score equally for all labels containing the term as a whole word, letting the secondary ORDER BY item_label produce a clean alphabetical list
- Raise SQL fetch limit (40→500), per-layer cap (30→50) and total cap (60→100) to prevent late-alphabet results being silently dropped
- Widen search panel (max-width 70%→85%), cap height at 95%
- Apply 2-column layout via CSS :has() only when a category has 11+ results; each item is capped at 280px with word wrapping

grafik

### DIFF
--- a/lizmap/modules/lizmap/classes/lizmapFts.class.php
+++ b/lizmap/modules/lizmap/classes/lizmapFts.class.php
@@ -27,10 +27,14 @@ class lizmapFts
         // Build search query
 
         // SELECT
+        // word_similarity measures how well the search term matches the best
+        // word-level substring of item_label. Unlike similarity(), it gives
+        // equal scores to all labels that contain the search term as a whole
+        // word, so the secondary alphabetical sort on item_label is effective.
         $sql = "
         SELECT
         item_layer, item_label, concat('EPSG:', ST_SRID(geom)) AS item_epsg, ST_AsText(geom) AS item_wkt,
-        similarity(trim( :searchedString ), item_label) AS sim
+        word_similarity(trim( :searchedString ), item_label) AS sim
         ";
 
         // FROM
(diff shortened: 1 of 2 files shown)


meyerlor:
I admit this is maybe questionable, and especially the values how many results per layer / in total should be displayed is set kind of randomly. But imho, this improves the search a bit for imprecise search terms, giving more results in a 2-column layout. No hard feelings if you reject this!

mdouchin:
Love it !

mdouchin:
We should really activate autocompletion for these searches. Not trivial to do since there are different providers, but it would be awesome to have. 

I built a quick hack for this website : 
https://sig.gard.fr/index.php/view/map/?repository=risques&project=zi_gard

See the related files 
* JavaScript file: 
[autocomplete_search_dock.js](https://github.com/user-attachments/files/26830259/autocomplete_search_dock.js)

* CSS file: 
[autocomplete_search_dock.css.css](https://github.com/user-attachments/files/26830264/autocomplete_search_dock.css.css)


This is just to see "what it could do". The code is quite old and not really maintained (some ref to OpenLayers 2 functions ...)

meyerlor:
ouhw, autocomplete would be loveley, indeed. I had it not on my list but now that you mention it..!

What i had on my list was "upward pointing children search" (dunno how to explain it - search for a child of a 1:n related relation, get the parent feature as result)

mdouchin:
> ouhw, autocomplete would be loveley, indded. I had it not on my list but now that you mention it..!
> 
> What i had on my list was "upward pointing children search" (dunno how to explain it - search for a child of a 1:n related relation, get the parent feature as result)

Nice to have too ! 
NB: I edited my previous post with embedded files

meyerlor:
The site is down (or i can not access it?) but i'm going through your files now. I'll draft a dedicated PR for autocompletion if i can make it work!

PR #6735, opened 2026-04-13
Fix sqlite migration

laurentj:
- fix: value of creation date of accounts were not imported
- fix: importing data was not possible without deleting all existing data. It is now possible

Funded by 3Liz


rldhont (approved):
Thanks @laurentj

PR #6738, opened 2026-04-14
Remove x from btn-selectiontool-clear

ygorigor:
After this https://github.com/3liz/lizmap-web-client/pull/6675 it's not needed anymore

Fix this 
image

To this

image

Backport to 3.10


rldhont (approved):
Good catch. Thanks @ygorigor

PR #6765, opened 2026-04-23
[Feature] Short link permalink functionality

mind84:
Adds new functionality to generate a `sha256` short link for permalink.

The system uses a database table to generate and save the hash and relies on local storage to store permalink information. This allows for structured storage without altering the URL.

The advantage of this system is the ability to cleanly manage complex permalinks, generating smaller URLs with the ability to manage additional information beyond style and opacity. Furthermore, it would solve problems related to the presence of special characters in the URL, which can sometimes be misinterpreted by messaging systems during sharing.
On the other hand, it will no longer be possible to copy the permalink directly from the browser bar, and it will therefore be necessary to generate a database record for each permalink.

**WITHOUT AUTOMATIC PERMALINK**
[vokoscreenNG-2026-04-23_15-03-12.webm](https://github.com/user-attachments/assets/fe302643-faf3-440b-88db-66d612c28442)

**WITH AUTOMATIC PERMALINK**
[vokoscreenNG-2026-04-23_15-01-56.webm](https://github.com/user-attachments/assets/f1e2cff6-630d-4c2b-a5ac-eda107dde2c0)

**NOTES**
This feature is designed to work as an alternative mode to the classic one, so it can be enabled using the `short_link_permalink` option in the Lizmap plugin (still in development). 
Given the premises, this mode is ideal for large-scale projects, where the permalink can also assume considerable lengths.

Also:
- Geobookmarks handling has been updated accordingly
- Works with single WMS requests
- First theme activation at startup is respected
- Classic mode is left untouched 

Funded by Etra


meyerlor:
Hi @mind84 !
Great idea and PR!

Just one thought: "First theme at startup is respected" - is that really wanted when accessing a project via a permalink? This is the current behavior but from a user perspective this is not expected imo. Saving a browser bookmark with a permalink for example: i would expect when opening my browser & clicking on the bookmark to open the map in my configured way, not the Startup-Theme.

mind84:
Hi @meyerlor!

> is that really wanted when accessing a project via a permalink?

I agree, it seems a bit confusing to me too.
I believe these are two conflicting features, since only the BBOX information is respected when this option is enabled.
It would probably make sense for the permalink to take precedence over this option.

mind84:
@rldhont,

Changes implemented:

- use of the `fetchJSON` method in `Permalink` class and update docstring
- use `history.replaceState` instead of `history.pushState`
- ability to remove a single record from local storage and delete all permalink records from local storage
- added default value test for `short_link_permalink` option
- added an admin page to display/remove permalinks from the table and updated tests accordingly

[vokoscreenNG-2026-05-14_11-02-16.webm](https://github.com/user-attachments/assets/c5bc286b-a110-4ce7-a6cb-8f9ecc33da75)

rldhont:
Thanks @mind84 for the last changes.

rldhont (approved):

PR #6767, opened 2026-04-24
[Fix] Upgrade Alpine to 3.22 and Php to 8.3 for Docker

u-cav:
Fix docker builds, currently broken due to two issues:

* References to the old version (8.1) in https://github.com/3liz/lizmap-web-client/blob/93187c27d14e03af877070d728690cd474cc2335/docker/Dockerfile#L78 https://github.com/3liz/lizmap-web-client/blob/93187c27d14e03af877070d728690cd474cc2335/docker/lizmap-entrypoint.sh#L130 https://github.com/3liz/lizmap-web-client/blob/93187c27d14e03af877070d728690cd474cc2335/docker/lizmap-entrypoint.sh#L136

* Dependency issue in Alpine 3.21/3.22, where composer (installed in the Dockerfile) depends on php8.3 (see https://pkgs.alpinelinux.org/package/v3.21/community/x86_64/composer). This means that the image is built with:
    * php8.2 with extensions
    * php8.3 without extensions, and this is the default version

    So update-config.php (https://github.com/3liz/lizmap-web-client/blob/master/docker/lizmap-entrypoint.sh#L85) fails because it's run with php8.3, missing various necessary extensions.

The easiest solution is to upgrade to PHP 8.3, so that versions match. The upgrade to Alpine 3.22 is not strictly necessary, 3.21 works as well.

As far as I can tell, the latest docker image (3liz/lizmap-web-client:3.10.0-beta.3) is broken due to these issues.

Funded by Faunalia

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,46 +1,46 @@
 ARG REGISTRY_PREFIX=''
-FROM ${REGISTRY_PREFIX}alpine:3.21 AS build
+FROM ${REGISTRY_PREFIX}alpine:3.22 AS build
 
 COPY lizmap_web_client.zip .
 RUN unzip lizmap_web_client.zip
 
-FROM alpine:3.21
+FROM alpine:3.22
 LABEL org.opencontainers.image.authors="David Marteau <david.marteau@3liz.com>"
 LABEL Description="Lizmap web client" Vendor="3liz.org" Version="23.02.0"
 
 RUN apk update && apk upgrade && \
     apk --no-cache add git fcgi \
     icu-data-full \
-    php82 \
-    php82-ctype \
-    php82-curl \
-    php82-dom \
-    php82-exif \
-    php82-fileinfo \
-    php82-fpm \
-    php82-gd \
-    php82-gettext \
-    php82-iconv \
-    php82-intl \
-    php82-json \
-    php82-ldap \
-    php82-mbstring \
-    php82-opcache \
-    php82-openssl \
-    php82-pdo \
-    php82-pdo_sqlite \
-    php82-pdo_pgsql \
-    php82-pgsql \
-    php82-phar \
-    php82-redis \
-    php82-session \
-    php82-simplexml \
-    php82-sqlite3 \
-    php82-tokenizer \
-    php82-xml \
-    php82-xmlreader \
-    php82-xmlwriter \
-    php82-zip \
+    php83 \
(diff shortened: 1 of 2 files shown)


ygorigor:
Custom css theme is not working with this commit

u-cav:
Hi @ygorigor, thank you for the feedback! We are actively using this build with custom css, as far as I know the docker packaging changes in this PR should not have any effects on custom css. Do you have errors in the browser console or in the logs? Is there a way for us to replicate the issue?

ygorigor:
Hi @u-cav. In www/themes I created another dir with a custom_theme name. In LWC there is a environment variable LIZMAP_THEME where I set the theme name (LIZMAP_THEME: 'custom_theme'). 
With PHP8.1 custom theme is used, but with php8.3, theme is not used.
It seems that it use only default theme.

No error logs in docker or browser.
For now I switched to php8.1, with php8.2 LWC is not working.

u-cav:
Hi @ygorigor, I just tried 3.9.7 with PHP 8.1 and with PHP8.3 (from this PR), and I can see the custom theme in both. Is the issue on a specific Lizmap version? Can you share the docker images?

ygorigor:
@u-cav I'm on LWC 3.11.0-pre

u-cav:
Hi @ygorigor, I just tested the latest LWC 3.11.0-pre with PHP8.3, with docker and this PR, and I can confirm that the `LIZMAP_THEME` works correctly.

I am unable to start LWC 3.11.0-pre with PHP8.1 (`PHP Fatal error:  Composer detected issues in your platform: Your Composer dependencies require a PHP version ">= 8.2.0". You are running 8.1.22. in /www/lizmap/vendor/composer/platform_check.php on line 24`)

Therefore I can't reproduce the issue that you are experiencing. Are you running 3.11.0-pre in Docker with PHP8.1?

Just for reference, I uploaded an image: `ucav/lizmap-web-client:3.11.0-pre-php83`

ygorigor:
Hi @u-cav I tried LWC with your image, and I have the same issue. 
I will try to find what is is wrong.
Thank you

rldhont (approved):

rldhont:
@u-cav do you think it could be merged ?

u-cav:
@rldhont As far as I know yes, we are using this configuration with 3.9, 3.10 and master without issues.

PR #6769, opened 2026-04-27
[Feature] Locate by layers loaded event

mind84:
Added new event `lizmap.locatebylayers.loaded`. 

This allows for a convenient way to interact with the Locate By Layers panel via external js, since the event is fired once all the Locate By Layers features have been loaded.

I've done a small refactoring, leveraging Lizmap's `WFS` class and using the `async/await` pattern instead of the jQuery `$.post` API.

You can subscribe to the event in your `custom.js`:

```js
lizMap.subscribe(()=>{
    // code here
}, 'lizmap.locatebylayers.loaded')

// or

lizMap.subscribe(()=>{
    lizMap.subscribe(()=>{
        // code here
    }, 'lizmap.locatebylayers.loaded')
}, 'lizmap.uicreated')
```

Funded by Faunalia

### DIFF
--- a/assets/src/modules/LocateByLayer.js
+++ b/assets/src/modules/LocateByLayer.js
@@ -9,6 +9,7 @@
 import DOMPurify from 'dompurify';
 
 import GeoJSON from 'ol/format/GeoJSON.js';
+import WFS from './WFS.js';
 
 /**
  * @class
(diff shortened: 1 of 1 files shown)


rldhont (approved):
To enhance this module, it will be great to separate fetching features and displaying user interface in an other pull request.

PR #6818, opened 2026-05-11
Feat/make-default-expressions-work

meyerlor:
This PR aims to bring original QGIS behaviour for Lizmap. Default expressions set via "Attribute From" in QGIS did not work so far when either
a) a geometry expression was used (e.g. `$x`, `aggregate()` with filters based on geometry etc)
b) a nother field of the same layer was referenced ( `"firstname" || ' ' || "lastname"` )

With this PR following behaviour is introduced:

- Server-side (PHP): At form-open, dynamic default expressions are now batched into a single EXPRESSION/Evaluate call with form_scope=true and a form_feature GeoJSON payload containing the current geometry (from liz_geometryColumn) and field values. Literal/numeric defaults short-circuit this and are applied immediately without a server round-trip.
- New AJAX endpoint: GET/POST /edition/{repo}/{project}/{layerId}/evaluateDefaults — re-evaluates a specified set of default expressions given a geometry and current field values. Gated by the same ACL as the existing edition actions.
- Client-side (JS): Two triggers re-call the endpoint and fill the form:
  - Geometry change — fires after the user finishes drawing/moving a feature (lizmap.editionGeometryUpdated event from updateGeometryColumnFromFeature)
  - Referenced field change — a debounced change listener on each field that appears in another field's default expression

Fill policy matches QGIS Desktop:
On create: fill if the field is still empty
On modify: only fill if the QGIS attribute has applyOnUpdate=true; user-typed values are never overwritten

DefaultExpression

### DIFF
--- a/CHANGELOG-3.11.md
+++ b/CHANGELOG-3.11.md
@@ -14,6 +14,8 @@ with some extra keywords: backend, tests, test, translation, funders, important
 
 ### Added
 
+* Edition - Support QGIS dynamic default-value expressions in edit forms, including geometry-based (`$x`, `$y`, `$area`, `$length`, `$geometry`) and field-referencing expressions (e.g. `"firstname" || ' ' || "lastname"`). Defaults are re-evaluated when the geometry is drawn/edited and when a referenced field changes, honoring QGIS's `applyOnUpdate` flag.
+
 ### Changed
 
 ### Backend
--- a/CHANGELOG-3.10.md
+++ b/CHANGELOG-3.10.md
@@ -8,6 +8,9 @@ with some extra keywords: backend, tests, test, translation, funders, important
 
 ## Unreleased
 
+### Added
+
+* Edition - Support QGIS dynamic default-value expressions in edit forms, including geometry-based (`$x`, `$y`, `$area`, `$length`, `$geometry`) and field-referencing expressions (e.g. `"firstname" || ' ' || "lastname"`). Defaults are re-evaluated when the geometry is drawn/edited and when a referenced field changes, honoring QGIS's `applyOnUpdate` flag.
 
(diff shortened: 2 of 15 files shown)


rldhont (approved):

CHANGELOG-3.11.md
```diff
@@ -14,6 +14,8 @@ with some extra keywords: backend, tests, test, translation, funders, important
 
 ### Added
 
+* Edition - Support QGIS dynamic default-value expressions in edit forms, including geometry-based (`$x`, `$y`, `$area`, `$length`, `$geometry`) and field-referencing expressions (e.g. `"firstname" || ' ' || "lastname"`). Defaults are re-evaluated when the geometry is drawn/edited and when a referenced field changes, honoring QGIS's `applyOnUpdate` flag.
+
```
rldhont, line 18: Hi @meyerlor, can you move this into CHANGELOG-3.10.md ?

rldhont (changes requested):

CHANGELOG-3.11.md
meyerlor: yes, done! Should I do that in future as well or do you prefere filling the changelog.md files yourselve?

rldhont: Thanks. We would prefer that the developers update the changelog themselves for new features.

rldhont (approved):

CHANGELOG-3.11.md
meyerlor: ok, gotcha! I will fill the Changelog in the future!

mdouchin:
Thanks a lot ! This is a killer feature we would like to implement a long time ago !

PR #6821, opened 2026-05-12
Popup: place children features tables inside drag-and-drop relation 

meyerlor:
When 'Display children as features table' (children_lizmap_features_table) is enabled on a parent layer that uses a QGIS drag-and-drop form layout, each child <lizmap-features-table> is now placed inside the matching .popup_lizmap_dd_relation placeholder div - i.e. inside the tab the relation was placed into in the QGIS form - instead of being appended at the end of the popup.

Adds CSS scoped to div.popup_lizmap_dd_relation > lizmap-features-table:
- Drops the inner per-table` <h4>` title (the tab usually names the relation).
- Wraps the features-table in a 1px border on the non-table wrapper div (`.lizmap-features-table`), with `border-radius 4px` and `overflow:hidden` for clean corners. The border lives on the non-table wrapper so it bypasses `.table-bordered`'s split tr/td border model (which can sub-pixel round to zero on high-DPI 100% zoom) and Bootstrap 5's per-cell  trick (which paints over any outline or inset box-shadow declared on the `<table>`). <- this is not really working though, as you can see in the provided screenshot. Happy for some input how to handle this better, as it's not really looking nice imo
- Zero bottom margins so the list does not leave a large gap before the bottom of the surrounding box.
- Tightens the wrapping fieldset padding via `:has(...)` when the QGIS form wraps the relation in a GroupBox - other form fieldsets untouched.

Adds a Playwright test fixture children_features_table_in_relation_div covering the placement of features tables inside relation placeholders.

grafik


josemvm:
hi @meyerlor 
would you be in favor of adding zebra striping to the table rows?
i think it might help improve readability and make it easier to distinguish between rows

meyerlor:
Yes, I think the UI needs some fine-tuning in this regard - but I basically just moved the child-attribute-table to a different position, so any UI changes to that attribute table is out of scope here. But I have it on my list (I'm not really fond of UI work though tbh 😆 )

PR #6828, opened 2026-05-14
Migrate snapping from OL2 to OL10

meyerlor:
Picking up the work of my previous attempt to migrate the edition tools to OL10 - i tried to atomize the needed changes as much as possible to make reviewing easier. This is the first of 3 PR's in the initial pipeline, targeting only the snapping function. Opening those other two PR's as draft, they are based on this PR subsequently.
This PR's scope:
* Replace OpenLayers.Control.Snapping with the OL10 Snap interaction.
* Render snap features as a solid orange overlay while snapping is active.
* Drop the manual refresh button: getSnappingData() now runs automatically on moveend and on layer-tree visibility changes (the snapLayersRefreshable state and snapping.refreshable event are gone, no consumers remained).
* OL10 Snap exposes a single vertex flag and a single pixelTolerance, so snap_vertices is OR-ed with snap_intersections and pixelTolerance is max(snap_vertices_tolerance, snap_segments_tolerance). Worth flagging for projects that configure vertex vs intersection independently.
* Public reorderSnapInteraction() lets future callers re-add the Snap interaction so OL processes it before any recently-added Draw / Modify / Translate. No-op when snap is inactive.

Tests:
* snap.spec.js: remove three WFS 1.1.0 / SRSNAME / datum-shift tests that exercised the OL2 snap WFS path (replaced by OL10 GeoJSON parsing). Strip refresh-button click / disabled-state assertions from the snap panel test now that refresh is automatic.

This PR ships the snap infrastructure but does not visibly engage with the OL2 edition Draw / Modify yet. Snapping starts to actually snap once the follow-up OL10 edition migration lands.

Note: The provided gif shows the snapping function is working on my testsystem where the follow up PR's are merged already:
Snapping

### DIFF
--- a/assets/src/components/Snapping.js
+++ b/assets/src/components/Snapping.js
@@ -26,23 +26,12 @@ export default class Snapping extends HTMLElement {
             <h3>${lizDict['snapping.title']}</h3>
             <div class="form-group">
                 <div class="controls">
-                    <div class="btn-group">
-                        <button
-                            class="btn ${mainLizmap.snapping.active ? 'active btn-success' : ''}"
-                            @click=${() => mainLizmap.snapping.toggle()}
-                            >
-                            ${mainLizmap.snapping.active ? lizDict['geolocate.toolbar.stop'] : lizDict['geolocate.toolbar.start']}
-                        </button>
-                        <button
-                            class="btn ${mainLizmap.snapping._snapLayersRefreshable ? 'btn-warning' : ''}"
-                            ?disabled=${!mainLizmap.snapping._snapLayersRefreshable}
-                            @click=${() => mainLizmap.snapping.getSnappingData() }
-                            >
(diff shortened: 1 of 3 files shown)


mind84:
Hi @meyerlor!

We plan to implement the edition tool, introducing the ability to enable snapping for geometries drawn or imported via the digitization tool.

There are some draft PRs you proposed a while back; I wonder if they cover the functionality in question.

In any case, I believe this PR is useful for planning future developments regarding the snapping functionality.

@rldhont, @nboisteault do you plan to merge this PR into 3.10.0?

Thanks!

nboisteault:
Hi @mind84 @meyerlor 
Sorry but we lack time to make reviews currently. We want to finish 3.10 first.

mind84:
Thanks for the info @nboisteault!

PR #6830, opened 2026-05-14
Ol10 geometry copy

meyerlor:
Migrate Copy/Paste geometry to OL10
* Bind clicks to mainLizmap.map (OL10) in edition context; keep the OL2
  click handler for non-edition flows.
* Toggle mainLizmap.digitizing.toolSelected = deactivate / restore to
  pause drawing while copy mode is active, instead of poking the now
  removed OL2 mainLizmap.edition.drawControl.

Without this PR the <lizmap-paste-geom> button rendered by the OL10
digitizing toolbar would still call into the OL2 paths and silently fail
on click.


mdouchin:
Hi @meyerlor Is this PR still needed ? If so, could you please rebase so that we tackle this issue ? Thanks a lot

PR #6842, opened 2026-05-16
Fix/Avoid GeomFromGPB log flooding for GeoPackage filter extent

### LINKED ISSUE #6165: [Bug]: Lizmap filter with GeoPackage layer triggers "no such function: GeomFromGPB" in logs

When using the “Form filter” tool in Lizmap Web Client (v3.9.2) on a GeoPackage layer, the filter itself works and correctly filters the features on the map.

However, Lizmap repeatedly writes errors in the server log related to the SQL function GeomFromGPB.
The issue seems to be caused by Lizmap trying to compute the spatial extent (BBOX) of the filtered features through a direct SQLite query.

Because the SQLite engine used by Jelix/PHP does not include the QGIS-specific function GeomFromGPB, the query fails, and Lizmap logs multiple PHP notices and SQL errors.

The filter widget remains functional, but the logs are flooded with errors. The problem only occurs with GeoPackage (OGR provider) layers. With PostGIS layers, the filter runs without any error, since QGIS Server executes the spatial query correctly.

The root cause is that Jelix/PHP tries to compute a BBOX using:

`SELECT AsGeoJSON(Extent(Transform(GeomFromGPB("geom"), 2056)), 8, 1)`

but the GeomFromGPB() function is only available inside QGIS or GDAL, not in plain SQLite. The issue is harmless functionally, but pollutes the logs and may hide other errors.

1. Create a QGIS project with a GeoPackage (.gpkg) layer containing a geometry column.
2. Example: layer name actions_localisees, field Avancement.
3. Publish the project in Lizmap Web Client 3.9.2.
4. Configure a Form filter in the Lizmap plugin for the layer:

- Filter type: uniquevalues
- Field: Avancement
- Provider automatically set to ogr

5. Open the Lizmap project in the browser.
6. Use the filter to select one or several values.
7. Check the Lizmap server log (var/log/error.log).

**Versions :**

* Lizmap Web Client : 3.9.3 - commit 32bad131d https://github.com/3liz/lizmap-web-client/commit/32bad131d
* Lizmap plugin : 4.5.4
* QGIS Desktop : 3.40.10
* QGIS Server : 3.40.11
* Py-QGIS-Server : 1.9.6

### Check Lizmap plugin

### Operating system

Ubuntu 20.04

### Browsers

Firefox

### Browsers version

LTR

### Relevant log output

```shell
2025-10-10 09:30:18 192.168.200.1 lizmapadmin  Error in query (no such function: GeomFromGPB ( SELECT AsGeoJSON(Extent(Transform(GeomFromGPB("geom"), 2056)), 8, 1) AS bbox FROM actions_localisees WHERE 2>1 AND ( "Avancement" IN ( 'En cours' ) )))
2025-10-10 09:30:10 [2] Undefined array key "authcfg" /var/www/lizmap-web-client/lizmap/modules/lizmap/classes/qgisVectorLayerDatasource.class.php 166
2025-10-10 09:30:10 [2] Undefined array key "schema" /var/www/lizmap-web-client/lizmap/modules/lizmap/classes/qgisVectorLayerDatasource.class.php 166
2025-10-10 09:30:10 [2] Undefined array key "tablename" /var/www/lizmap-web-client/lizmap/modules/lizmap/classes/qgisVectorLayerDatasource.class.php 166
2025-10-10 09:30:10 [2] SQLite3::query(): Unable to prepare statement: 1, no such function: GeomFromGPB /var/www/lizmap-web-client/vendor/jelix/jelix/lib/jelix/plugins/db/sqlite3/sqlite3.dbconnection.php 129
```

meyerlor:
Fixes #6165

### DIFF
--- a/CHANGELOG-3.9.md
+++ b/CHANGELOG-3.9.md
@@ -46,6 +46,7 @@ with some extra keywords: backend, tests, test, translation, funders, important
 * Attribute table - export with filter
 * Filter - Hide the form filter panel export button when the user is not allowed to export the layer (#6642)
 * Filter - Date filter calendar now shows all years within the data range (#6548)
+* Filter - Skip the GeoPackage filtered-extent query when SpatiaLite lacks GeoPackage support, instead of flooding the logs (#6165)
 
 ### Tests
 
--- a/lizmap/modules/filter/classes/filterDatasource.class.php
+++ b/lizmap/modules/filter/classes/filterDatasource.class.php
@@ -128,6 +128,28 @@ class filterDatasource
         return $q;
     }
 
+    /**
+     * Check whether a SQL function is available on the SQLite/SpatiaLite connection.
+     *
+     * @param string $functionName the function name to look up (case-insensitive)
+     *
+     * @return bool true if available, or if availability cannot be determined
+     */
+    private function sqliteFunctionExists($functionName)
+    {
+        try {
+            $sql = 'SELECT count(*) AS c FROM pragma_function_list';
+            $sql .= ' WHERE lower(name) = '.$this->cnx->quote(strtolower($functionName));
+            $res = $this->cnx->query($sql)->fetch();
+
+            return $res && (int) $res->c > 0;
+        } catch (Exception $e) {
+            // pragma_function_list unavailable: assume the function exists
+            // and let the caller attempt the query as before.
+            return true;
+        }
+    }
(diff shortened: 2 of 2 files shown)


rldhont:

meyerlor:
done, moved to changelog 3.9!

PR #6843, opened 2026-05-16
Fix/Drop NULL entry in split-value form filters 

### LINKED ISSUE #6166: [Bug]: Filter widget displays unwanted NULL value when using “splitter” in form filtering

When configuring a form filter (uniquevalues type) with a splitter character (e.g. comma , or semicolon ;), Lizmap automatically adds a NULL (empty) entry in the list of possible values — even though there are no NULL or empty values in the source data.

This behavior occurs systematically as soon as a splitter is defined, regardless of the data content.
If the splitter is removed, the NULL entry disappears, but Lizmap then lists all full combinations of multi-values instead of splitting them, which is not usable in practice.

This suggests an internal logic issue when Lizmap builds the list of unique values after splitting strings — likely because an empty token ('') is added to the list during the PHP explode() process.

The list of available filter values should only include real unique tokens from the field, excluding NULL or empty strings.

1. In QGIS, create a GeoPackage layer with a text field containing multi-values separated by commas, for example:

> agenda
> -------
> 2026,2027
> 2028,2029
> avant
> futur

→ No NULL values, no empty strings, no trailing commas.

2. Publish the layer in Lizmap 3.9.3 and configure a form filter:

> "formFilterLayers": {
>     "0": {
>         "layerId": "my_layer_id",
>         "title": "Agenda",
>         "type": "uniquevalues",
>         "field": "agenda",
>         "format": "checkboxes",
>         "splitter": ","
>     }
> }

3. Open the Lizmap map viewer, expand the “Agenda” filter list.

4. Observe that an additional (NULL) (or empty) entry appears in the list, even though the field never contains any NULL or empty values.

5. If you remove the splitter parameter, the (NULL) entry disappears, but then all combinations appear literally (e.g. 2026,2027).

**Versions :**

* Lizmap Web Client : 3.9.3 - commit 32bad131d https://github.com/3liz/lizmap-web-client/commit/32bad131d
* Lizmap plugin : 4.5.4
* QGIS Desktop : 3.40.10
* QGIS Server : 3.40.11
* Py-QGIS-Server : 1.9.6

### Check Lizmap plugin

### Operating system

Ubuntu 20.04

### Browsers

Firefox

### Browsers version

LTR

### Relevant log output

```shell
No error is logged in the Lizmap or QGIS Server logs.
Behavior seems internal to Lizmap’s PHP processing of filter values (likely during explode() or aggregation step in FormFilter.php).
```

meyerlor:
Fixes #6166

### DIFF
--- a/CHANGELOG-3.9.md
+++ b/CHANGELOG-3.9.md
@@ -47,6 +47,7 @@ with some extra keywords: backend, tests, test, translation, funders, important
 * Filter - Hide the form filter panel export button when the user is not allowed to export the layer (#6642)
 * Filter - Date filter calendar now shows all years within the data range (#6548)
 * Filter - Skip the GeoPackage filtered-extent query when SpatiaLite lacks GeoPackage support, instead of flooding the logs (#6165)
+* Filter - No longer show a spurious (NULL) entry in split-value form filters without NULL data (#6166)
 
 ### Tests
 
--- a/lizmap/modules/filter/classes/filterDatasource.class.php
+++ b/lizmap/modules/filter/classes/filterDatasource.class.php
@@ -309,6 +309,7 @@ class filterDatasource
                 $sql .= ' SELECT cat AS v, sum(nb) AS c';
                 $sql .= ' FROM source';
                 $sql .= ' GROUP BY cat';
+                $sql .= ' HAVING sum(nb) > 0';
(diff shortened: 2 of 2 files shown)


rldhont:

meyerlor:
doen, moved to changelog 3.9!

PR #6848, opened 2026-05-17
Fix/Timemanager: Add layername

meyerlor:
Added the layername(s) under control of TimeManager to the PopUp:

grafik


rldhont:

meyerlor:
done, moved to changelog 3.9!

PR #6405, opened 2026-01-05
Add geometry "copy&paste" functionality for digitizing workflow

meyerlor:
This commit implements a new feature that allows users to copy existing feature geometries while digitizing new features, improving the workflow when creating features with similar shapes to existing ones.

This is on my personal "wish list" since quite some time :)
**Demonstration**

CopyPaste

**Description**

1. **GeometryCopyHandler** (assets/src/modules/GeometryCopyHandler.js)
   - Manages the geometry copy workflow
   - Queries visible layers via WMS GetFeatureInfo using standard Lizmap popup mechanism (text/html format)
   - Parses HTML response to extract layer names (lizmapPopupTitle) and geometries (WKT format from hidden inputs)
   - Filters features by geometry type compatibility (Point, LineString, Polygon, including Multi* variants)
   - Temporarily deactivates draw control during selection to prevent accidental vertex creation
   - Applies copied geometry to current editing session

2. **FeaturePickerPopup** (assets/src/modules/FeaturePickerPopup.js)
   - Displays selectable list of features found at click position
   - Shows layer names in a clean popup table
   - Highlights features on hover using temporary vector layer
   - Handles feature selection and popup cleanup

3. **CopyGeometryButton** (assets/src/components/CopyGeometryButton.js)
   - Web component providing UI button in digitizing toolbar
   - Enabled when edition session is active (drawFeatureActivated or layerId present)
   - Listens to edition state changes for proper enable/disable state
   - Toggles copy mode on/off with visual feedback (active state)

4. **FeatureStorage** (assets/src/modules/FeatureStorage.js)
   - Enhanced with copy() method for storing geometries with metadata
   - Stores geometry type, source layer, and CRS information
   - Dispatches 'featureStorage.copy' event for UI synchronization

5. **PasteGeom** (assets/src/components/edition/PasteGeom.js)
   - Updated to support all geometry types (Point, LineString, Polygon, MultiPoint, MultiLineString, MultiPolygon)
   - Implements proper OL6 to OL2 geometry conversion for each type

1. User starts editing a layer (clicks "Add feature")
2. User clicks "Copy existing geometry" button in digitizing toolbar
3. Draw control is temporarily deactivated, cursor changes to crosshair
4. User clicks on map to query features at that location
5. System queries all visible popup-enabled layers via WMS GetFeatureInfo
6. Popup displays found features, filtered by geometry type compatibility
7. User hovers over table rows to preview geometry highlights
8. User clicks a row to copy that geometry
[...]

PR #6888, opened 2026-05-26
Fix copy-paste geometry

meyerlor:
Three issues reported on PR #6405 after testing:

1. The copy/paste button showed two tooltips on hover (one in English, one in French). Removed the inner button's tooltip attributes and the redundant $('button', this).tooltip(...) call.

2. Clicking on the map returned "no compatible geometry" even on layers that visually had features. QGIS Server only emits the geometry attribute in GetFeatureInfo when WITH_GEOMETRY=true *and* the layer is published as WFS; without that, WMSRequest.php never inserts the hidden .lizmap-popup-layer-feature-geometry input that GeometryCopyHandler reads. Candidate layers were filtered only by visibility + popup, so popup-only layers were silently skipped one feature at a time. Added an explicit WFS filter using mainLizmap.initialConfig.vectorLayerFeatureTypeList (same source DxfExport uses) and an upfront message when none of the visible popup-enabled layers are WFS-published.

3. The "no compatible features" / "geometry copied" messages stacked up, because addMessage() just appends. Added a _notify() helper that removes any previous #lizmap-copy-geometry-message before appending and auto-dismisses after 5s — same pattern legacy/edition.js uses for lizmap-edition-message. All four user-facing strings now go through it and read from lizDict (edition.geom.copyPaste.*) with English fallbacks; the matching keys are added to the EN dictionary so Transifex can pick them up.


rldhont:
@meyerlor thanks to fix these issues. Can you add end2end to test `#lizmap-copy-geometry-message` displayed and to check not duplicated messages provided by `_notify` ?

meyerlor:
done!

rldhont (approved):

rldhont:
@meyerlor thanks!

PR #6889, opened 2026-05-27
JS dependencies update rspack to v2

rldhont:
Update @rspack/core and @rspack/cli to 2.0.x


rldhont:
Superseeded by https://github.com/3liz/lizmap-web-client/pull/7038

PR #6900, opened 2026-05-29
Fix startup popup true highlight

nboisteault:
Funded by Argentan's paid support

### DIFF
--- a/tests/end2end/playwright/startup.spec.js
+++ b/tests/end2end/playwright/startup.spec.js
@@ -1,6 +1,8 @@
 // @ts-check
 import { test, expect } from '@playwright/test';
 import { ProjectPage } from "./pages/project";
+import { expect as requestExpect } from './fixtures/expect-request.js';
+import { expect as responseExpect } from './fixtures/expect-response.js';
 
 test.describe('Startup', () => {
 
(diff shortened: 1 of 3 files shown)


rldhont (approved):

tests/end2end/playwright/startup.spec.js
```diff
@@ -41,6 +41,27 @@ test.describe('Startup', () => {
...
+            'layer': 'sousquartiers',
+            'filter': '"quartmno" = \'PA\' OR "quartmno" = \'HO\'',
+            'popup': 'true',
+        });
+
+        // Wait for the GetFeatureInfo request triggered by popup=true
+        const getFeatureInfoRequest = await getFeatureInfoPromise;
+        const postData = getFeatureInfoRequest.postData() ?? '';
+        expect(postData).toContain('GetFeatureInfo');
+        expect(postData).toContain('sousquartiers');
```
rldhont: Please use `toContainParametersInPostData` from `expect-request`.

tests/end2end/playwright/startup.spec.js
```diff
@@ -41,6 +41,27 @@ test.describe('Startup', () => {
...
+
```
rldhont, line 65: Here wait for response, and use `toBeHtml` (if it is an HTML response) from `expect-response`.

rldhont (changes requested):

PR #6902, opened 2026-06-01
fix(edition): persist read-only QGIS default-value expressions on save

meyerlor:
Read-only form controls never receive their value from the request: jFormsBase::initFromRequest() skips them on purpose, so a dynamic default (e.g. \ or round(\, 2)) rendered into a non-editable widget was filled client-side for display but silently dropped on save, persisting NULL instead.

Re-evaluate the dynamic default expressions server-side just before saving, using the posted geometry and field values as context, and inject the result into the form data. This mirrors QGIS Desktop, which evaluates such defaults at commit time regardless of widget editability.

Fill policy: on insert, only fill fields left empty; on update, only re-evaluate fields flagged applyOnUpdate. Applied fields are added to the controls persisted on update so they are written even when untouched.

### DIFF
--- a/CHANGELOG-3.10.md
+++ b/CHANGELOG-3.10.md
@@ -8,6 +8,10 @@ with some extra keywords: backend, tests, test, translation, funders, important
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


mind84:
LGTM

Thanks!

rldhont (approved):

PR #6905, opened 2026-06-02
[Bugfix] Add upgrade script for short link permalink

mind84:
This PR completes https://github.com/3liz/lizmap-web-client/pull/6765

@laurentj there are a couple of things that aren't completely clear to me on Jelix upgrade module logic:

- variable `$targetVersions`
  As far as I understand, if the current upgrading version matches one of the version listed in this variable, then the upgrade script may be executed, depending on other subsequent checks. On the other hand, if the version doesn't match, the script never runs.

- variable `$date`
  What is the purpose of this field?

In general, is there a way to force a module to update?
I updated Lizmap to version `3.10.0-rc.1`, but I forgot to add the `upgrade_permalink` script. I tried adding it later and then execute `installer.php`, but it seems there's no way to run the script once the main module has been updated.

Thanks for the support!

Funded by Faunalia


mind84:
@rldhont

I will rebase this PR after https://github.com/3liz/lizmap-web-client/pull/6908 is merged. 
Same for https://github.com/3liz/lizmap-web-client/pull/6907 and https://github.com/3liz/lizmap-web-client/pull/6906

Thanks

lizmap/modules/admin/install/upgrade_permalink.php
```diff
@@ -3,9 +3,10 @@
 class adminModuleUpgrader_permalink extends jInstallerModule
 {
     public $targetVersions = array(
-        '3.10',
+        '3.10.0-rc.1',
+        '3.10.0',
```
rldhont:
```suggestion
        '3.10.0-pre',
        '3.10.0-rc.2',
        '3.10.0',
```
`-rc.1` is already passed.
The actual is `-pre`
The futur is -`rc.2` and `.0`

@laurentj are these target versions rights ?

lizmap/modules/lizmap/install/upgrade_permalink.php
```diff
@@ -0,0 +1,20 @@
...
+    public $date = '2026-05-26';
+
+    public function install()
+    {
+        // Adding permalink table
+        if ($this->firstDbExec()) {
+            // Add permalink table
+            $this->useDbProfile('jauth');
+            $this->execSQLScript('sql/lizpermalink');
+        }
```
rldhont: I think, you can merge it with `admin/install/upgrade_permalink.php` to keep only one `upgrade_permalink.php` file.

@laurentj What do you think about this?

lizmap/modules/admin/install/upgrade_permalink.php
laurentj: `$targetVersions` must contain the next version of the module. So you don't have to put all these `3.10.*` versions. Only one is enough. You can put several versions, when the upgrade script should be executed for different branches of the software. For example, if it should be applied for 3.9.25, and 3.10.0-rc.1, you put these two versions.

Then, if the user upgrade from 3.9.20 to 3.9.26, the script will be executed. And then if he upgrades to 3.10, the script won't be executed again. If he upgrade from 3.9.20 to 3.10 or from 3.9.26 to 3.10, the script will always be executed once. The upgrade system knows that by comparing the date of the upgrade script and the date of the releases. So the date of the script should be the date of the creation of the script.

In development, to force the execution of the script, you can change the `version` of the module and its  `version.date` into `var/config/installer.ini.php`, with a previous version. And then run again the `install/installer.php`. 

In production, it is not recommended. And if an upgrade script was missing in a released version, you must release a new version of the application, and indicate this version into the `targetVersion`

laurentj: you should indicate only `3.10.0-rc.2`.

The origin of the issue about why it was not executed, it is because the target version was `3.10`, so it won't be executed for 3.10-rc.1 etc, and it is because, **the version indicated into the module.xml is still 3.10.0-pre**, which is lower than 3.10.0-rc*.  (pre < alpha < beta < rc < stable)

After the release of 3.10.0-rc.1, the version into module.xml was changed to 3.10.0-pre, whereas it should be 3.10.0-rc.2...

lizmap/modules/lizmap/install/upgrade_permalink.php
laurentj: technically, no matter. But if it is about the same functionality, probably it is better to have a single upgrade script.

laurentj (changes requested):

mind84:
@laurentj, @rldhont,

I merged the upgrade scripts. This also required moving the `sql` scripts to the `admin` module, and consequently, to avoid duplication, I also moved the installation logic from the `lizmap` module to the `admin` module.

Are you ok with this changes?

Thanks!

laurentj (approved):
LGTM

PR #6906, opened 2026-06-02
[Bugfix] Prevent edition from removing hash information

mind84:
When you click on the add feature button from the edition panel, the hash is overwritten

[vokoscreenNG-2026-06-02_14-58-06.webm](https://github.com/user-attachments/assets/74d193a1-8fa4-4d82-a76d-ee8860458584)

Funded by Faunalia


rldhont (approved):

rldhont:
Thanks @mind84 !

PR #6916, opened 2026-06-08
fix(map): close measure & geolocation minidocks via native click

meyerlor:
The #mapmenu .nav click handler was refactored to native addEventListener, but the X buttons on the measure (#measure-stop) and geolocation (.btn-geolocation-close) minidocks still triggered the toggle via jQuery .click(). jQuery's .trigger()/.click() does not dispatch a real DOM event to native listeners, so the close became a silent no-op. Switch both handlers to native addEventListener + HTMLElement.click().

### DIFF
--- a/assets/src/legacy/map.js
+++ b/assets/src/legacy/map.js
@@ -795,9 +795,8 @@ window.lizMap = function() {
 
         if (('geolocation' in configOptions)
       && configOptions['geolocation'] == 'True'){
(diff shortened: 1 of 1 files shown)


nboisteault:
Thanks @meyerlor 
I've made two separated commits instead:
5c4c960a1d2c56dff0a163a54311b75119a0c15e
e80251b39ea4e444b1eaa73a002a30f0f2287152

PR #6917, opened 2026-06-09
Actions - Add the ability to send dynamic parameters from JavaScript

Elioooooott:
Allow passing custom options (e.g. `buffer_size`) to Lizmap actions via the `options` argument of `runLizmapAction`. These options can be set dynamically from a JS custom form or any JS function.

Example:
```javascript
lizMap.mainLizmap.action.runLizmapAction(
    "project_map_center_buffer", // actionId
    'project',                   // scope
    null,                        // layerId
    null,                        // featureId
    null,                        // wkt
    {"buffer_size": 2000}        // options
);
``` 

Funded by 3Liz

### DIFF
--- a/lizmap/modules/action/controllers/service.classic.php
+++ b/lizmap/modules/action/controllers/service.classic.php
@@ -164,7 +164,6 @@ class serviceCtrl extends jController
 
         // Compute the parameters to pass to the PostgreSQL action function
         // depending on the scope and given parameters
-        $data = array();
         $action_params = array(
             'lizmap_repository' => $repository,
             'lizmap_project' => $project,
@@ -192,23 +191,58 @@ class serviceCtrl extends jController
             $action_params['feature_id'] = $featureId;
         }
 
-        // Get the additional options from the JSON config file
-        // Not for any requests parameters !!!
-        foreach ($action->options as $k => $v) {
-            $action_params[$k] = $v;
+        // Run the action
+        $i = 1;
+        $sqlParts = array();
+        $sqlValues = array();
+        foreach ($action_params as $key => $value) {
+            $caster = ($key == 'feature_id') ? 'integer' : 'text';
+            $sqlParts[] = "'{$key}', (\${$i})::{$caster}";
+            $sqlValues[] = $value;
+            ++$i;
         }
 
-        // Run the action
-        $sql = "SELECT lizmap_get_data('";
-        $sql .= json_encode($action_params);
-        $sql .= "') AS data";
+        // Get the additional options from the JSON config file
+        // We need to be cautious with this values as they are sent from the client
+        $options = $this->param('options');
+        foreach ($action->options as $key => $config_value) {
+            $sqlParts[] = "'{$key}', (\${$i})::text";
+            $sqlValues[] = $options[$key] ?: $config_value;
+            ++$i;
+        }
 
+        $sql = '
+            SELECT public.lizmap_get_data(
+                json_build_object(
+                '.implode(",\n", $sqlParts).'
+                )
+            ) AS data
+        ';
+        jLog::log(
+            "SQL =
+            {$sql}"
+        );
+
+        // Prepare returned data
+        $resultset = null;
+
+        // Start SQL transaction
+        $cnx->beginTransaction();
+
+        // Run the lizmap_get_data function with a prepared statement
+        // to avoid SQL injections
+        // $sql = "SELECT json_build_object('un', $1::text, 'deux', $2::text)";
         try {
-            $res = $cnx->query($sql);
-            foreach ($res as $r) {
-                $data = json_decode($r->data);
+            $resultset = $cnx->prepare($sql);
+            $execute = $resultset->execute($sqlValues);
+            if ($resultset && $resultset->id() === false) {
+                $errorCode = $cnx->errorCode();
+
+                throw new Exception($errorCode);
             }
+            $cnx->commit();
         } catch (Exception $e) {
+            $cnx->rollback();
             jLog::log(
                 'Error in project '.$repository.'/'.$project.', layer '.$layerId.', '.
                 'while running the action with the PostgreSQL query : '.$sql.' → '.$e->getMessage(),
@@ -220,6 +254,14 @@ class serviceCtrl extends jController
             );
 
             return $this->error($errors);
+        }
+        $data = array();
+        if ($resultset !== null) {
+            foreach ($resultset as $r) {
+                if ($r->data) {
+                    $data = json_decode($r->data);
+                }
+            }
         }
 
         // Send response
--- a/lizmap/modules/action/controllers/service.classic.php
+++ b/lizmap/modules/action/controllers/service.classic.php
@@ -1,5 +1,6 @@
 <?php
 
+use Lizmap\ActionQuery\ActionQuery;
 use Lizmap\App\WktTools;
 use Lizmap\Project\UnknownLizmapProjectException;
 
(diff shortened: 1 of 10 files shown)


lizmap/modules/action/controllers/service.classic.php
```diff
@@ -221,6 +255,14 @@ public function index()
 
             return $this->error($errors);
         }
+        $data = array();
```
mdouchin: We should probably have a default returned data to the client, so that the JS can display a message such as "An error occurred while running this action"

lizmap/modules/action/controllers/service.classic.php
```diff
@@ -192,23 +191,58 @@ public function index()
...
+
         // Get the additional options from the JSON config file
-        // Not for any requests parameters !!!
-        foreach ($action->options as $k => $v) {
-            $action_params[$k] = $v;
+        // We need to be cautious with this values as they are sent from the client
+        $options = $this->param('options');
+        foreach ($action->options as $key => $config_value) {
+            $sqlParts[] = "'{$key}', (\${$i})::text";
+            $sqlValues[] = $options[$key] ?: $config_value;
```
mdouchin: We should add another layer of security here. Probably pass the values to something like for the filter tool
https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/filter/classes/filterDatasource.class.php#L92

But better move the filter sanitation tools to a dedicated library of the lizmap module. The code is already duplicated between filter and WFS request (see https://github.com/3liz/lizmap-web-client/blob/db240957097c4bdb65b23f729389253ab9f5d57f/lizmap/modules/lizmap/lib/Request/WFSRequest.php#L52 )

mdouchin (changes requested):
Thanks for this PR ! Please add a security layer and sanitize the content of the passed options.

lizmap/modules/lizmap/lib/App/SqlTools.php
```diff
@@ -0,0 +1,39 @@
...
+     * @param string $filter The expression filter to validate
+     *
+     * @return bool returns if the expression does not contains dangerous chars
+     */
+    public static function validateExpressionFilter(string $filter): bool
+    {
+        $block_items = array();
+        $pattern = '#'.implode('|', array_map(fn ($w): string => preg_quote($w, '#'), static::$blockSqlWords)).'#i';
+
+        return !preg_match($pattern, $filter, $block_items);
```
mdouchin:
```suggestion

        return array(!preg_match($pattern, $filter, $block_items), $block_items);
```

This way each caller would be able to log the error for its context, for example

```php
jLog::log('The EXP_FILTER param contains dangerous chars : '.implode(', ', $block_items), 'lizmapadmin');
```

lizmap/modules/filter/classes/filterDatasource.class.php
```diff
@@ -87,11 +75,8 @@ private function validateFilter($filter)
...
             $filter = str_replace(' ILIKE ', ' LIKE ', $filter);
         }
-        $block_items = array();
-
-        if (preg_match('#'.implode('|', $this->blockSqlWords).'#i', $filter, $block_items)) {
-            jLog::log('The EXP_FILTER param contains dangerous chars : '.implode(', ', $block_items), 'lizmapadmin');
 
+        if (!SqlTools::validateExpressionFilter($filter)) {
             return null;
         }
```
mdouchin, line 84: Please keep the log here by returning the $block_items from the `validateExpressionFilter` method

lizmap/modules/lizmap/lib/Request/WFSRequest.php
```diff
@@ -654,7 +639,7 @@ function ($n) {
             $expFilter = '$id IN ('.implode(', ', $pks).')';
         }
 
-        if ($expFilter && $this->validateExpressionFilter($expFilter)) {
+        if ($expFilter && SqlTools::validateExpressionFilter($expFilter)) {
             return $expFilter;
         }
```
mdouchin: We should call the `SqlTools::validateExpressionFilter($expFilter))` before the `if`
and reverse the logic.

If the expression is not valid, add a `jLog` then return ''. Else return `$expFilter`

lizmap/modules/lizmap/lib/Request/WFSRequest.php
```diff
@@ -1172,7 +1138,7 @@ protected function validateExpressionFilter(string $filter): bool
      */
     protected function validateFilter(string $filter): false|string
     {
-        if (!$this->validateExpressionFilter($filter)) {
+        if (!SqlTools::validateExpressionFilter($filter)) {
             return false;
         }
```
mdouchin: Idem

mdouchin (changes requested):
Some changes needed. Thanks !

mdouchin (approved):
Thanks a lot

rldhont:
@Elioooooott 

Some PHP Stan errors to fix:

```
Error: Parameter $qgisLayer of method Lizmap\ActionQuery\ActionQuery::buildParams() has invalid type Lizmap\ActionQuery\qgisVectorLayer.
Error: Call to method getDatasourceParameters() on an unknown class Lizmap\ActionQuery\qgisVectorLayer.
Error: Call to method getName() on an unknown class Lizmap\ActionQuery\qgisVectorLayer.
Error: PHPDoc tag @throws with type Lizmap\ActionQuery\Exception is not subtype of Throwable
Error: Left side of && is always true.
 ------ --------------------------------------------------------------------- 
  Line   lizmap/lib/ActionQuery/ActionQuery.php                               
 ------ --------------------------------------------------------------------- 
  39     Parameter $qgisLayer of method                                       
         Lizmap\ActionQuery\ActionQuery::buildParams() has invalid type       
         Lizmap\ActionQuery\qgisVectorLayer.                                  
  62     Call to method getDatasourceParameters() on an unknown class         
         Lizmap\ActionQuery\qgisVectorLayer.                                  
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
  63     Call to method getName() on an unknown class                         
         Lizmap\ActionQuery\qgisVectorLayer.                                  
         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
  129    PHPDoc tag @throws with type Lizmap\ActionQuery\Exception is not     
         subtype of Throwable                                                 
  136    Left side of && is always true.                                      
         💡 Because the type is coming from a PHPDoc, you can turn off this   
            check by setting treatPhpDocTypesAsCertain: false in your         
            ../phpstan-ci.neon.                                               
 ------ --------------------------------------------------------------------- 
```

The e2e tests about action failed:

```
2 failed
    [end2end] › playwright/action.spec.js:14:5 › Lizmap actions @readonly › should display working project action 
    [end2end] › playwright/feature-toolbar.spec.js:354:5 › Feature toolbar in popup @readonly › should display working custom action
```

lizmap/modules/lizmap/lib/Request/WFSRequest.php
```diff
@@ -654,11 +639,17 @@ function ($n) {
...
-            return $expFilter;
+        [$validFilter, $block_items] = SqlTools::validateExpressionFilter($expFilter);
+        if (!$expFilter || !$validFilter) {
+            $this->appContext->logMessage(
+                'The EXP_FILTER param contains dangerous chars : '.implode(', ', $block_items),
+                'lizmapadmin'
+            );
+
+            return '';
         }
```
mdouchin, line 653:
```suggestion
        if (!$expFilter) {
            return '';
        }
        [$validFilter, $block_items] = SqlTools::validateExpressionFilter($expFilter);
        if (!$validFilter) {
            $this->appContext->logMessage(
                'The EXP_FILTER param contains dangerous chars : '.implode(', ', $block_items),
                'lizmapadmin'
            );

            return '';
        }
```

mdouchin (changes requested):

PR #6572, opened 2026-02-12
[Feature] Management of Value relation/Value Map fields in the new datatables Search Builder panel

mind84:
Added an implementation of the new Search Builder feature. This development includes:

- search builder panel now recognizes the `Value relation/Value map` fields and renders conditions with an appropriate drop-down menu showing key-value mapping.
- Large dropdown menus add a lot of nodes to the DOM, which can slow down performance during DOM reflow. To address this, I've introduced a new Lizmap component, `TypeAHead`. For now, this component is very basic; it only handles displaying filtered options and knows nothing about the filtering method. Of course, it's a starting point, and I'd be happy to discuss potential evolutions of this component with you. The ultimate goal is to move the filtering of options to the server side, but that's another story :D
- I introduced also a "**huge_table**" for testing purpose. Not so huge actually (5k records), but it can be useful for testing large datasets
- add corresponding tests

Here you can see the functionality in action. 
- `Large lookup` field --> condition rendered with new `TypeAHead` component since the lookup has more than 50 elements
- `Small lookup` field --> condition rendered with traditional dropdown (lookup with <= 50 elements)
- `Value Map field`--> condition rendered with traditional dropdown (lookup with <= 50 elements)

[vokoscreenNG-2026-02-13_08-35-06.webm](https://github.com/user-attachments/assets/769447c4-4743-485a-a04e-4ab018dcb9a3)

Backport 3.10

Funded by Etra

PR #6919, opened 2026-06-11
[Bugfix] Searchbuilder: key-value description to string

mind84:
It takes into account cases where the layer's `keyValue` configuration has a numeric type description.

Related to https://github.com/3liz/lizmap-web-client/pull/6572

Funded by Faunalia


mdouchin (approved):
Thanks

Issue #5932, opened 2025-07-09
[Bug]: Project with no unit in projection definition leads to a fatal error


nworr:
### What is the bug? (in English)

When a project use a projection without unit in the definition (exemple : [OGC:CRS84](https://spatialreference.org/ref/ogc/CRS84/)  : `+proj=longlat +datum=WGS84 +no_defs +type=crs` 

The `lizmapTiler` class fails to compute the Tile Matrix Set (see logs)

https://github.com/3liz/lizmap-web-client/blob/d8735083cb33050917507dead5e07408cb2aa248/lizmap/modules/lizmap/classes/lizmapTiler.class.php#L296

### Steps to reproduce the issue

define a project with a projection without explicit units in its defintion

### Versions, safeguards, check summary etc

Lizmap 3.8

### Check Lizmap plugin

- [x] I have done the step just before in the Lizmap QGIS desktop plugin before opening this ticket. Otherwise, my ticket is not considered valid and might get closed.

### Operating system

whatever

### Browsers

Firefox

### Browsers version

whatever

### Relevant log output

```shell
Undefined array key "unit"
  /srv/lizmap/lib/lizmap-web-client/lizmap_web_client-3.8.11.8624/lizmap/modules/lizmap/classes/lizmapTiler.class.php   296
 /index.php/lizmap/service?repository=foo&project=bar&SERVICE=WMTS&REQUEST=GetCapabilities&VERSION=1.0.0
 array (   'repository' => 'foo',   'project' => 'bar',   'SERVICE' => 'WMTS',   'REQUEST' => 'GetCapabilities',   'VERSION' => '1.0.0',   'module' => 'lizmap',   'action' => 'service:index', )
 

        0       lizmapTiler::getTileMatrixSetList()     /srv/lizmap/lib/lizmap-web-client/lizmap_web_client-3.8.11.8624/lizmap/modules/lizmap/classes/lizmapTiler.class.php : 107
        1       lizmapTiler::getTileCapabilities()      /srv/lizmap/lib/lizmap-web-client/lizmap_web_client-3.8.11.8624/lizmap/modules/lizmap/lib/App/JelixContext.php : 361
        2       Lizmap\App\JelixContext->getTileCaps()  /srv/lizmap/lib/lizmap-web-client/lizmap_web_client-3.8.11.8624/lizmap/modules/lizmap/lib/Request/WMTSRequest.php : 52
[...]

PR #6944, opened 2026-06-19
check for undefined unit

nworr:
Ticket : #5932 

Funded by 3Liz


rldhont (approved):

nworr:
degree as default unit is relevant ?

rldhont:
@dmarteau degree as default unit in the case of proj4 string projection definition ?

dmarteau:
> @dmarteau degree as default unit in the case of proj4 string projection definition ?

Depends on the projectiion, for geographical projection it is degrees, for most other projections it is metres.

PR #6955, opened 2026-06-23
Optimize the getMedia controller

laurentj:
- reduce memory usage when sending http ranges content, by streaming the content instead of copying it into the memory
- Content-Length header is set by jResponseBinary, so no need to set it in the controller
- remove useless code about specific cases on html and text content

Funded by 3liz


rldhont (approved):
Thanks @laurentj !

PR #6959, opened 2026-06-25
log key for login should not contain folder

nworr:
Login event are logged on a dedcicated table if this feature is enabled
But since various entry point can performe login (webddav for instance) the code was checking the entrypoint file name

When lizmap installation is on an URL with subfolder (ie `https://domain/cartes/` the log is filled with folder name : 

Screenshot 2026-06-25 at 11-30-43 Administration 

The pr fix the behaviour : classical admin login will be showned as `login` , other entrypoint login will be prefixed by entry point name
Ticket : #

Funded by 3Liz


mdouchin (approved):
Thanks

PR #6976, opened 2026-07-03
fix: remove leftover Bootstrap tab classes from bottom-dock template

nboisteault:
map_bottomdock.tpl was the only outer dock template still wrapping its panels in Bootstrap's tabbable/tab-content/tab-pane/active classes, unlike dock/mini-dock/right-dock which already switched to the plain .hide convention driven by vanilla JS. This caused the bottom-dock "hide sibling panes" step to rely on a wrong #bottomdock-content selector (real id is #bottom-dock-content) and let Bootstrap's .tab-content>.active rule outrank the .hide utility class, so a docked popup and the attribute table could end up shown at the same time.

Funded by 3liz

### DIFF
--- a/assets/src/legacy/map.js
+++ b/assets/src/legacy/map.js
@@ -3592,7 +3592,9 @@ window.lizMap = function() {
                 const parentElement = linkClicked.parentElement;
                 const wasActive = parentElement.classList.contains('active');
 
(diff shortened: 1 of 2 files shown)


rldhont:
LGTM, @mdouchin I would like your point of view.

nboisteault:
Tests are passing locally

PR #6993, opened 2026-07-07
dataviz: load plotly.js via npm and lazy-load it on first plot draw

nboisteault:
Plotly.js was vendored as a static custom-built file, injected server-side only when dataviz was enabled for a project. It is now a regular npm dependency (same custom trace bundle: scatter, bar, box, pie, histogram, histogram2d, scatterpolar, sunburst), exposed as lizMap.plotlyjs and always available, but its chunk (plus one small chunk per locale) is only downloaded the first time a plot is actually drawn.

Funded by 3Liz


lizmap/modules/dataviz/controllers/service.classic.php
```diff
@@ -351,15 +351,6 @@ protected function GetPlot()
             'layout' => $dplot->getLayout(),
         );
 
-        // We also add the URL to access the Plotly JavaScript file
-        // to let the client use the same one
-        $basePath = jApp::config()->urlengine['basePath'];
-        $locale = substr(jApp::config()->locale, 0, 2);
-        $plot['plotly'] = array(
```
nworr, line 358: this line seems necessary for the plugin
https://github.com/3liz/lizmap-plugin/blob/ab68208c4526e35de6eb7d3d410b8e777386e188/lizmap/table_manager/dataviz.py#L277

PR #6994, opened 2026-07-09
Fix WFS/filter selection: translate all geometry operators to PostGIS

meyerlor:
The selection tool and the filter form let the user pick any of the QGIS geometry predicates (intersects, contains, within, crosses, overlaps, touches, disjoint). On the direct PostgreSQL query path the expression filter is rewritten to SQL, but only "intersects" was mapped to its ST_* function. Every other operator was passed through verbatim, producing invalid SQL (e.g. within(...) instead of ST_Within(...)), so the query failed and the selection returned nothing on PostGIS layers.

Rewrite the translation to map all seven predicates to their PostGIS equivalents, matching the function name case-insensitively and only when followed by "(" so plain text and column names are left untouched. The same fix is applied to the filter module which shared the incomplete mapping.

### DIFF
--- a/lizmap/modules/filter/classes/filterDatasource.class.php
+++ b/lizmap/modules/filter/classes/filterDatasource.class.php
@@ -82,10 +82,8 @@ class filterDatasource
 
             return null;
         }
-        $filter = str_replace('intersects', 'ST_Intersects', $filter);
-        $filter = str_replace('geom_from_gml', 'ST_GeomFromGML', $filter);
 
-        return str_replace('$geometry', '"'.$this->datasource->geocol.'"', $filter);
+        return SqlTools::translateExpressionToPostgis($filter, $this->datasource->geocol);
     }
 
     /**
--- a/lizmap/modules/lizmap/lib/App/SqlTools.php
+++ b/lizmap/modules/lizmap/lib/App/SqlTools.php
@@ -1,5 +1,16 @@
 <?php
 
+/**
+ * SQL tools for Lizmap.
+ *
+ * @author    3liz
+ * @copyright 2026 3liz
+ *
+ * @see      http://3liz.com
+ *
+ * @license Mozilla Public License : http://www.mozilla.org/MPL/
+ */
+
 namespace Lizmap\App;
 
 class SqlTools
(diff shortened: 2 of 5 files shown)


mdouchin:
Nice catch. Do you think you could move the duplicated logic into a library, for example https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/lib/App/SqlTools.php ? This way there would be only one place to modify in case of future improvements

meyerlor:
stumbled across it because of a typo in the german translation, so it's kind of a lucky catch :D

mdouchin:
Beware I just merged a work on the action module which created the sqlTools file. Please rebase on master 🙏 
See https://github.com/3liz/lizmap-web-client/blob/master/lizmap/modules/lizmap/lib/App/SqlTools.php

meyerlor:
yes, just saw that - i have moved everything, should be OK now!

mdouchin (approved):
Thanks a lot !

rldhont:
@meyerlor can your rebase your branch ?

It is better to keep your branch clean of merged to let Github Action created backport branches.

meyerlor:
ofc, sorry, got a bit mixed up!

rldhont:
@meyerlor can your rebase your branch ?

rldhont:
@meyerlor can you rebase instead of merge master ? Merging master will broke backport.

meyerlor:
ah sorry, true! Done!

CHANGELOG-3.10.md
```diff
@@ -8,6 +8,10 @@ with some extra keywords: backend, tests, test, translation, funders, important
 
 ## Unreleased
 
+### Fixed
+
+* Selection tool & DB filter - Fix spatial selection on PostgreSQL layers with the `contains`, `within`, `crosses`, `overlaps`, `touches` and `disjoint` operators, which were not translated to their PostGIS functions and returned no result
+
```
rldhont: It will blocked backport to release_3_9.

CHANGELOG-3.11.md
```diff
@@ -16,6 +16,10 @@ with some extra keywords: backend, tests, test, translation, funders, important
 
 * Edition - Support QGIS dynamic default-value expressions in edit forms, including geometry-based (`$x`, `$y`, `$area`, `$length`, `$geometry`) and field-referencing expressions (e.g. `"firstname" || ' ' || "lastname"`). Defaults are re-evaluated when the geometry is drawn/edited and when a referenced field changes, honoring QGIS's `applyOnUpdate` flag.
 
+### Fixed
+
+* Selection tool & DB filter - Fix spatial selection on PostgreSQL layers with the `contains`, `within`, `crosses`, `overlaps`, `touches` and `disjoint` operators, which were not translated to their PostGIS functions and returned no result
+
```
rldhont: It will blocked backport to release_3_10 and release_3_9.

rldhont (changes requested):
Remove changes in CHANGELOD-3.10.md and CHANGELOG-3.11.md to be able to easily backport changes.

meyerlor:
> Remove changes in CHANGELOD-3.10.md and CHANGELOG-3.11.md to be able to easily backport changes.

uhm , this confuses me a bit - in the past this was the approved way. Where should i add changelog-worthy changes?

rldhont:
> > Remove changes in CHANGELOD-3.10.md and CHANGELOG-3.11.md to be able to easily backport changes.
> 
> uhm , this confuses me a bit - in the past this was the approved way. Where should i add changelog-worthy changes?

Because, the changelog is seprated by version, you have to update the changelog for the lower version to backport to.

rldhont (approved):

meyerlor:
so in future commits, i use the lowest version the commit should be backported to? will do!

PR #6925, opened 2026-06-12
Admin page listing projects : filter by repository / UI fix + inspection file tests

nworr:
- Add a title to the page
- Use zone in template instead of controller
- New feature : allow to filter projects by repository
- Add `VersionTools::strVersionToMajMinInt` returning str version as int with only MajMin version (3.05.12 → 305)
- Remove `FIXME` in zone using  `VersionTools` methods
- fix for lizmap cloud hosted instance , unformated lizmap version

Screenshot 2026-07-02 at 14-17-36 Admin - Lizmap projects - Administration
is now 
Screenshot 2026-07-02 at 14-16-49 Admin - Lizmap projects - Administration

Ticket : # 

Funded by 3liz

PR #6996, opened 2026-07-09
feat: serve the admin QGIS projects list as JSON to DataTables

nboisteault:
The projects list page rendered a full HTML table server-side and let DataTables upgrade it in place. Switch to an AJAX/JSON source instead: a new QgisProjectsListData class (lib/, LizmapAdmin namespace) builds each row as raw values plus a precomputed CSS class/tooltip per cell, exposed by a new qgis_projects:data JSON action. The table template is now a static shell and activate_datatable.js loads rows via ajax with explicit columns, replacing the old DOM-sourced init (and fixing its hardcoded column-index bugs along the way).

Funded by 3Liz


rldhont:
@nboisteault I have merged #6925 I think, you have to update your PR.

nboisteault:
@rldhont it is a PR we did with @nworr to test Claude Code. @nworr what do we do now?

PR #6997, opened 2026-07-10
Fix Services URL to be compatible with some external clients (lizmap 3.9 backport)

laurentj:
WMS/WMTS/etc services URL should not have the repository and the project as query parameters.

Funded by 3Liz


laurentj:
I don't know why some end2end failed.

laurentj:
All tests are green !

tests/end2end/playwright/wmts.spec.js
rldhont:
```suggestion
        await page.unroute('**/wmts_test*');
```
unroute the right rule will help to get strong tests

rldhont (changes requested):
Thanks, just one update

rldhont:
@laurentj sorry I have merged it!

PR #7008, opened 2026-07-14
fix(search): resolve relative nominatim URL against the page URL

meyerlor:
The jQuery->vanilla refactor (6284d4530) passes the server-generated nominatim endpoint, a root-relative path (/index.php/lizmap/osm/nominatim), to the URL constructor. new URL() throws on a relative URL without a base, so external nominatim search crashes with "URL constructor: ... is not a valid URL" and never fires the request. jQuery $.get previously accepted the relative path.

Pass window.location.origin as the base so the path resolves. IGN uses an absolute URL and is unaffected.

### DIFF
--- a/assets/src/modules/Search.js
+++ b/assets/src/modules/Search.js
@@ -225,7 +225,7 @@ export default class Search {
         const searchQuery = document.getElementById('search-query').value;
         switch (searchConfig.service) {
             case 'nominatim': {
-                const nominatimUrl = new URL(service);
+                const nominatimUrl = new URL(service, window.location.href);
                 nominatimUrl.searchParams.set('query', searchQuery);
                 nominatimUrl.searchParams.set('bbox', extent.toBBOX());
                 Utils.fetchJSON(nominatimUrl.toString()).then(data => {


meyerlor:
@mdouchin sorry for beeing pushy, but would it be possible to merge this particular PR sooner then later? Would be highly appreciated!

mdouchin:
Sorry for the delay. Quick answer, have you tested with a Lizmap application URL like https://acme.org/some_path/index.php ?
MDN docs shows the base path could be removed : https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

```javascript
new URL("/en-US/docs", "https://developer.mozilla.org/fr-FR/toto");
// returns { href: "https://developer.mozilla.org/en-US/docs", origin: "https://developer.mozilla.org/", protocol: "https:", username: "", password: "", host: "developer.mozilla.org", hostname: "developer.mozilla.org", port: "", pathname: "/en-US/docs", search: "" }
```

meyerlor:
Thanks @mdouchin! Yep, tested with a subpath URL too — works fine.

The trick is that service comes from jUrl::get('lizmap~osm:nominatim'), which already includes the basePath (e.g. /some_path/index.php/lizmap/osm/nominatim?…). Since it starts with /, only the origin matters for resolving it, like your MDN example where the base path gets dropped. So the subpath is preserved.

Happy to use window.location.href instead of .origin if you prefer — same result here, just makes the intent clearer.

mdouchin:
Thanks for checking !

> Happy to use window.location.href instead of .origin if you prefer — same result here, just makes the intent clearer.

If possible, it would make sense IMHO

meyerlor:
done.


> Sorry for the delay. 

no sweat! There is no need to hurry and i know you have a lot on the table. I was just a bit too brave and installed 3.10-rc on a productive server, and now ppl are missing the adress search, so that's why there is a certain urgency (for me) to have this fixed 😄 

The failing e2e test is unrelated I think.

Just saw that you will be involved in several talks at the UC! Nice, really looking forward to it!

mdouchin:
> no sweat! There is no need to hurry and i know you have a lot on the table. I was just a bit too brave and installed 3.10-rc on a productive server, and now ppl are missing the adress search, so that's why there is a certain urgency (for me) to have this fixed 😄

Ok !

> The failing e2e test is unrelated I think.

Not sure, @rldhont any hint ?

> Just saw that you will be involved in several talks at the UC! Nice, really looking forward to it!

Yep, looking forward too !

rldhont (approved):

meyerlor:
@rldhont thanks alot, highly appreciated!

PR #7009, opened 2026-07-15
Fix Services URL to be compatible with some external clients

laurentj:
WMS/WMTS/etc services URL should not have the repository and the project as query parameters.

Funded by 3Liz


tests/end2end/playwright/wmts.spec.js
rldhont:
```suggestion
        await page.unroute('**/wmts_test*');
```

rldhont (changes requested):
Only one change

rldhont (approved):

PR #7010, opened 2026-07-16
Fix/honor external wms toggle

meyerlor:
`getProjectConfig` unconditionally set `externalWmsToggle='True' `and `externalAccess` for every unauthenticated raster WMS layer, overriding the per-layer "get images directly from the WMS server" option written to the CFG by the plugin. WMS layers could no longer cascade through QGIS Server: 
For example, a server that only serves its native CRS could not be reprojected, and the browser issued cross-origin GetMap requests that failed with `OpaqueResponseBlocking`.

Direct client access is now exposed only when the CFG opts in
(externalWmsToggle truthy) or for xyz/WMTS EPSG:3857 tile layers.

To keep external-WMS legends working independently of tile routing, a
dedicated wmsLayer flag now drives PNG legend rendering for every WMS layer,
so cascaded (toggle-off) WMS layers still get a proper legend instead of
empty JSON GetLegendGraphic icons.

### DIFF
--- a/assets/src/modules/action/Symbology.js
+++ b/assets/src/modules/action/Symbology.js
@@ -32,8 +32,11 @@ export async function updateLayerTreeLayersSymbology(treeLayers, method=HttpRequ
 
     if (method.toUpperCase() == HttpRequestMethods.GET) {
         for (const treeLayer of treeLayers) {
-            // Check if this is an external WMS layer using the backend flag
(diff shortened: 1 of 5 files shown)


rldhont:
@meyerlor can you rebase instead of merge master ? Merging master will broke backport.

meyerlor:
done!

rldhont (approved):

PR #7013, opened 2026-07-16
[CI] retry all curl errors when checking QGIS Server status in e2e wo…

nboisteault:
…rkflow

The nginx container can refuse/reset the first connection right after it starts, before it is actually ready to serve requests. curl's --retry only covers timeouts and HTTP 5xx codes, not low-level connection errors like "Connection reset by peer" (exit code 56), so the check failed immediately instead of retrying.


rldhont (approved):

nboisteault:
manually backported to release_3_10 and release_3_9

rldhont:
Thanks @nboisteault

PR #7018, opened 2026-07-21
Quote column names in DataTables search filter expressions

meyerlor:
The DataTables SearchBuilder filter built expressions like `Flur = 510` with the column identifier left unquoted. On the PostgreSQL direct query path in WFSRequest (getfeaturePostgres), this is passed as raw SQL, where PostgreSQL folds unquoted mixed-case identifiers to lower case, producing `column "flur" does not exist` errors for any layer whose columns contain uppercase letters.

Enclose the column name in double quotes (doubling any embedded quote), which is valid both as a QGIS Server expression and as PostgreSQL SQL, so mixed-case and special-character column names resolve correctly.

<

### DIFF
--- a/lizmap/modules/lizmap/lib/DataTables/DataTables.php
+++ b/lizmap/modules/lizmap/lib/DataTables/DataTables.php
@@ -124,7 +124,13 @@ class DataTables
                 break;
         }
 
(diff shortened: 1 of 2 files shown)


mdouchin (approved):
Seems good to men thanks

rldhont (approved):

PR #7027, opened 2026-07-22
[GA] E2E: launched @readonly if QGIS Server is available

rldhont:
Launched Playwright end2end tests only if QGIS Server is available.


rldhont:
Manually backported to 3.10 e91bcae30

rldhont:
Manually backported to 3.9 fcb5624c1

Issue #6876, opened 2026-05-20
Filtering layers - how to move forward


meyerlor:
As brough up by @mdouchin I'd like to start here the discussion/exchange on how to move forward regarding the filters of layers - the concern about filter attribution is real and should be adressed properly. 

Let me play devil's advocate first: I would say the three open PRs ([#6840](https://github.com/meyerlor/lizmap-web-client/issues/6840), [#6846](https://github.com/meyerlor/lizmap-web-client/issues/6846), [#6847](https://github.com/meyerlor/lizmap-web-client/issues/6847)) are safe to land as-is (with minor adjustments), and do not make the existing fragility meaningfully worse. My reasonings is:

[#6840](https://github.com/meyerlor/lizmap-web-client/issues/6840) (keep other layers' filters when switching) only removes a deactivateFilter() call that was incorrectly wiping a layer's filter when the user selected a different layer in the form panel. Each layer's filter already lives independently in its WMS request_params.filter. Nothing really new is introduced, it's just a regression fixed.

[#6847](https://github.com/meyerlor/lizmap-web-client/issues/6847) (combine TimeManager with existing filter) is semantically/logically correct: a feature should only appear if it passes all active filters. The old behavior (TimeManager silently overwriting the form filter) was the bug (imo). I could add a guard so that when TimeManager deactivates, it restores the pre-existing form filter rather than clearing the combined string entirely.

[#6846](https://github.com/meyerlor/lizmap-web-client/issues/6846) (legend panel filter removal) — here your concern is most valid. I suggest to scope the per-layer remove button to only appear when the filter on that layer was set directly by the form filter tool (not via cascading from a parent relation). That prevents the buttn from showing on cascade-filtered child layers where the source of truth is the parent. That would require small changes, but then it could be merged relatively safely.

--------------------------------------------------------------------------------------------------------------------

That said, you are absolutely right that the root architectural problem is remainign: there is no ownership "metadata" on filters. When `request_params.filter` is set on a layer, nothing records if it came from the form filter, the time manager, the attribute table, or a cascading relationship. That makes independent activation/deactivation possibly fragile across tools.

The fix I'd like to propose — and implement — is a centralized FilterState map, keyed by layerId, where each tool writes and clears its own slot like:

```
filterState[layerId] = {
  formFilter: "field = 'value'",
  timeManager: "date >= '2024-01-01'",
  attributeTable: null,
  cascading: null
}
[...]

PR #6847, opened 2026-05-17
Fix/Filter combined with TimeManager

meyerlor:
Fixes: https://github.com/3liz/lizmap-web-client/issues/6773

FilterTimeManager

PR #7055, opened 2026-07-29
[BUGFIX] Time manager: keep the filter already set on the layer

### LINKED ISSUE #6773: [Bug]: Time manager does not work on layers with an active filter

When activating time manager and the layer that is under is control has a filter enabled, then nothing happens, as it will show all (filtered) features at all times/dates.

See above.

**Versions :**

* Lizmap Web Client : 3.9.7 - commit 800219e95 https://github.com/3liz/lizmap-web-client/commit/800219e95
* Lizmap plugin : 4.5.9
* QGIS Desktop : 3.44.9
* QGIS Server : 3.44.9
* Py-QGIS-Server or QJazz : not used

<details>
<summary>List of Lizmap Web Client modules :</summary>
<br/>
* lizmapdemo : 3.6.0a1
</details>

<details>
<summary>List of safeguards :</summary>
<br/>
* Mode : safe<br/>
* Allow parent folder : no<br/>
* Prevent other drive : yes<br/>
* Prevent PG service : yes<br/>
* Prevent PG Auth DB : yes<br/>
* Force PG user&pass : yes<br/>
* Prevent ECW : yes<br/>
</details>

### Check Lizmap plugin

### Operating system

Ubuntu 24.04

### Browsers

Chrome

### Browsers version

Latest

### Relevant log output

```shell

```

nboisteault:
## Problem

Fixes #6773.

When a controlled layer already has a filter — typically set from the form filter — the time manager and that filter do not compose: the date range never restricts the features that the layer filter selected.

## Root cause

`setLayersFilter()` passed only its own date filter:

```js
filter = buildDateFilter(config.timemanagerLayers[l], lowerBoundary, upperBoundary);
lizMap.triggerLayerFilter(l, filter);
```

`lizMap.triggerLayerFilter()` *replaces* `request_params.exp_filter` on the layer, so the filter already applied was dropped at the first date step. Symmetrically, closing the time manager called `deactivateMaplayerFilter()`, wiping the layer filter along with the date one.

Reproduced with a form filter keeping `gid <= 2`: the GETFILTERTOKEN sent when opening the time manager contained only `"test_date"` conditions, the `"gid"` condition was gone.

## Fix

The filter each controlled layer has when the time manager is opened is captured in `tmBaseFilters` and combined with the date filter of every step. Closing the time manager restores that filter instead of clearing the layer, so only the date part is removed.

When the layer had no filter, the filter string is byte-for-byte unchanged, which keeps the existing (quite precise) GETFILTERTOKEN assertions in `time-manager.spec.js` and `time-manager-years.spec.js` valid.

**Known limitation:** the base filter is captured when the time manager is opened, so changing the form filter while the time manager is running is not picked up until it is reopened. Handling that properly needs the layer filters to be tracked per origin rather than as a single `exp_filter` string, which is a larger change.

## Test

Added a test in `time-manager.spec.js`: it filters the layer with the form filter, then asserts the time manager's GETFILTERTOKEN carries both the `"gid"` and the `"test_date"` conditions, and that closing the time manager restores the `"gid"` condition alone.

The `time_manager` test project had `formFilterLayers: {}`, so no layer filter could coexist with the date filter; a numeric form filter on `gid` is added to the fixture.

Verified locally: without the fix the test fails on the `"gid"` assertion (the filter contains only `"test_date"`), and passes with it.

Unrelated pre-existing local failures, identical on a clean `master` on my QGIS 3.44 stack: `Manual play`, `Let's play` and `Auto play uses full ISO dates in filter` (my stack resolves the time frame type to months, showing `January 2007` where the tests expect `2007`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/assets/src/legacy/timemanager.js
+++ b/assets/src/legacy/timemanager.js
@@ -63,6 +63,10 @@ var lizTimemanager = function() {
             var filter = null;
             var tmAnimationTimer;
             var tmCurrentDate;
+            // Filter each controlled layer already had when the time manager was
+            // opened, by layer name. The date filter is combined with it, instead
+            // of replacing it.
+            var tmBaseFilters = {};
             var tmStartDate = -Infinity; // lower bound of when values
             var tmEndDate = Infinity; // upper value of when values
 
@@ -83,6 +87,16 @@ var lizTimemanager = function() {
 
                 // hourglass
                 //$('#loading').dialog('open');
+
+                // Keep the filter the layers may already have, e.g. set from the
+                // form filter, so that the date filter is combined with it (#6773)
+                tmBaseFilters = {};
+                for (var tmLayer in config.timemanagerLayers) {
+                    var tmLayerRequestParams = config.layers[tmLayer]?.['request_params'];
+                    if (tmLayerRequestParams?.['exp_filter']) {
+                        tmBaseFilters[tmLayer] = tmLayerRequestParams['exp_filter'];
+                    }
+                }
 
                 // Get min and max timestamps from layers
                 var minTime = Infinity, maxTime = -Infinity ;
@@ -352,6 +366,15 @@ var lizTimemanager = function() {
                 // Set filter for each vector layer
                 for (var l in config.timemanagerLayers){
                     filter = buildDateFilter(config.timemanagerLayers[l], lowerBoundary, upperBoundary);
+
+                    // Combine the date filter with the filter the layer already had,
+                    // otherwise the time manager would drop it (#6773)
+                    if (tmBaseFilters[l]) {
+                        filter = filter
+                            ? '( ' + tmBaseFilters[l] + ' ) AND ' + filter
+                            : tmBaseFilters[l];
+                    }
+
                     lizMap.triggerLayerFilter(l, filter);
                 }
             }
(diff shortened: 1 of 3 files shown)


meyerlor:
Hey @nboisteault, that is almost the same fix I did in #6847, and it's basically the same `tmBaseFilters capture/combine/restore` shape, so nice that we (or our Claudes 😄  )landed on the same idea. Your Playwright test + gid fixture are a solid addition mine didn't have.

One thing before it lands: this captures the base filter from `exp_filter `only. That covers the form filter panel, but the attribute-table filter tool stores its filter in` request_params['filter'] ` instead (see `getDataAndFillAttributeTable`, `request_params['filter'] = filter)`, not `exp_filter `— so a filter set from the attribute table gets dropped again once the time manager kicks in. In #6847 I read `request_params['filter']` first (stripping the `wmsName`: prefix) and fall back to `exp_filter`, which handles both origins. Happy to fold that into this PR if useful.

Bigger picture though, this is exactly the mess I flagged in #6876: `request_params.filter` vs `exp_filter`, no ownership metadata, every tool guessing what was already there. This PR is a fine as a tactical fix — I'd just keep the per-layer `FilterState `idea from #6876 as the real target and treat the filter-key case as the one gap to close here - what do you think?

PR #7067, opened 2026-08-07
New parser for layer datasources

laurentj:
Fix several potential issues when the connection strings contain special characters into parameters values.

It has a better support of parameter values and quotes than a parser based on regexp. Support of escaped quote in all values etc.

It is probably faster than the previous parser, as it doesn't have to execute a regexp for each parameters.

Funded by 3liz.


rldhont:
@laurentj this datasource `$datasource = "dbname='test_dbname' host=test_host port=5432 user='test_user' password=\"test password\" sslmode=prefer";` failed with the message `syntax error, missing equal sign after parameter password` instead of to be valid

Issue #4075, opened 2024-01-17
Actions enhancement


arno974:
While the relatively new [action feature](https://docs.lizmap.com/next/en/publish/configuration/actions.html) is really useful, it can be limited in certain situations where an action has to be allowed only for specific Lizmap users/groups. Would it be possible to have such enhancement in a future release ?

Gustry:
Coincidence, I was just busy working on the "Actions" chapter, your link is now 404 :)
I moved it in the "plugin" section : [new link](https://docs.lizmap.com/current/fr/publish/lizmap_plugin/actions.html).

Check latest news on "Actions" in 3.7 : https://github.com/3liz/lizmap-web-client/releases/tag/3.7.0

Having some user&groups ACL is a good addition.

arno974:
Another suggestion related to this feature would be to know the user who performed the action. This would be available as another parameter in the public.lizmap_get_data function  like "action_scop", "map_center", etc.

PR #7079, opened 2026-08-27
[FEATURE] Actions: restrict action visibility by user and groups

arno974:
sponsored development
funded by GEOLAB - Arnaud Vandecasteele

Add two optional properties on each action of the .action configuration file:
- lizmap_user_groups: list of the allowed ACL group ids
- lizmap_user: list of the allowed user logins

They combine with an OR: the action is kept if the user belongs to one of the allowed groups or if their login is listed. If no property is set, the action stays visible to everybody, so existing .action files are unaffected.

The filtering happens in actionConfig, which every consumer goes through, so the actions are neither sent to the browser nor executable through the action service. There is no admin bypass, consistently with the print layouts "allowed_groups" behaviour.

New tests have been added and have been successfully passed.

Note: this only implements the server side. The UI to fill lizmap_user_groups and lizmap_user still has to be added to the Lizmap plugin for QGIS, and the .action file format documented on docs.lizmap.com.

Note 2: While implementing this, I looked for the existing convention for group-based ACL checks and found the same logic in about ten places, with four different normalizations 
 - options.acl for instance never trims its values. 
 - The DXF export check is even duplicated verbatim between the dock display (Project.php:2569) and  the export itself (service.classic.php:644).

Happy to open a dedicated issue with the full inventory if that's of interest.

Ticket : https://github.com/3liz/lizmap-web-client/issues/4075


mdouchin:
Thanks for this proposal, which is a great addition.

> Note: this only implements the server side. The UI to fill lizmap_user_groups and lizmap_user still has to be added to the Lizmap plugin for QGIS, and the .action file format documented on docs.lizmap.com.

Could you please add an issue in the Lizmap documentation Github repository ? Same for the lizmap plugin repository ?

> Note 2: While implementing this, I looked for the existing convention for group-based ACL checks and found the same logic in about ten places, with four different normalizations
> 
>     * options.acl for instance never trims its values. 
>     * The DXF export check is even duplicated verbatim between the dock display (Project.php:2569) and  the export itself (service.classic.php:644).

> Happy to open a dedicated issue with the full inventory if that's of interest.

We would love to have this report in a dedicated issue. Thanks for proposing it !

I think this PR would benefit for a refactoring of the list of groups (or users) parsing, somewhere in a dedicated lib, to avoid each module to have its own methods such as your `getAclProperty` function

arno974:
Thanks a lot for the review!

I have opened the three issues you asked for:

- Documentation : https://github.com/3liz/lizmap-documentation/issues/296
- QGIS Plugin https://github.com/3liz/lizmap-plugin/issues/761
- Refactoring export : https://github.com/3liz/lizmap-web-client/issues/7130

Finally, two practical points on this PR:

- I have rebased it on the current `master` to resolve the conflict in `action.spec.js` introduced by [786722d72](https://github.com/3liz/lizmap-web-client/commit/786722d725937b0bf7c24af6c2b6fcfd0ee2c267) ("Action: feature button can be display in single feature popup"). The rebase is complete and the conflict markers are gone.
The remaining local failure is is related to the feature action test, which currently expects two popup action buttons but receives none in the rendered popup. It seems to be a behavioral/test expectation issue rather than a Git conflict resolution problem. To be confirmed by @rldhont.
- The CI workflows are still awaiting maintainer approval (first contribution).  Could someone trigger them? Locally, on chromium: 14 Playwright tests pass in `action.spec.js`, PHPStan reports no error, and php-cs-fixer is clean.

arno974:
I have pushed the Rector fix (SimplifyIfReturnBoolRector on checkActionAcl), which resolves the Lint failure. Since the force-push created a new commit, the workflows are awaiting approval again  could someone trigger them on 40326af7? Locally: Rector clean, php-cs-fixer clean, 14 Playwright tests pass on chromium.

arno974:
I noticed that the integration test WFS GetFeature FILTERTOKEN is failing with a 400 Bad Request error. Looking at the logs, it seems the WFS service is rejecting the FILTERTOKEN parameter, which appears to be a backend issue.

Since my changes are unrelated to the WFS service or token handling, could you please let me know if this is a known issue or if I should investigate it further? It seems to be an existing failure in the CI environment.

rldhont:
The commits with `FILTERTOKEN` tests have been revert. You can rebase your branch.

arno974:
Rebased on `87e1ee34c`, which includes the two FILTERTOKEN reverts (`375976a1f`, `f304d1d5e`) thanks for the heads-up. The rebase was conflict-free: none of the four files touched by this PR moved upstream. The diff is unchanged.
Local validation on the new QJazz stack (QJazz 0.9.1, QGIS 3.40.15, `lizmap_server` 2.15.3):
  - php-cs-fixer: clean
  - Rector: clean
  - Playwright `action.spec.js`: 14/14 passed on chromium
 
Could someone approve the workflows?

**Side note: three issues in #7142 that break the local dev stack.** None of them can show up in CI, so I am reporting them here as I hit all three while rebasing:

  1. **`tests/Makefile:98`** : `build:: build-plugins` is still declared, but the `build-plugins` target was removed along with `add_server_plugins.sh`. `make build` fails with *No rule to make target 'build-plugins'*.  CI no longer calls `make build`, so it is not covered.
     
  2. **`tests/Makefile:18`** : `LZMQGSRVVERSION ?= qgis-ltr-eager`, but that tag does not exist on Docker Hub (`3liz/qjazz` publishes `qgis-3.40`, `qgis-3.44`, `qgis-ltr-unstable`,  `qgis-4.2`…). `make up` fails with *manifest unknown*. The e2e matrix overrides the variable (`LZMQGSRVVERSION: ${{ matrix.qgis-server }}`), so the default value is never exercised.
     
  3. **`tests/Makefile:77`** : `up`, `stop` and `reset` now force `--profile=dev`, which pulls in the `openldap` service. Its Dockerfile is based on `debian:buster-slim`, now EOL and removed from `deb.debian.org`, so the build fails on `404 Not Found` for every apt repository. CI runs a plain `docker compose up -d` with no profile, so this Dockerfile is never built there.
     
With `LZMQGSRVVERSION=qgis-3.40 make env` followed by `docker compose up -V --force-recreate -d` (no profile), everything comes up fine and QJazz auto-installs the three server plugins the migration itself works well.

Happy to open a dedicated issue if that helps.

mdouchin:
@dmarteau could you please have a look at @arno974 previous comment about the current status of the Makefile ?

dmarteau:
>  **`tests/Makefile:98`** : `build:: build-plugins` is still declared, but the `build-plugins` target was 
> removed  along with `add_server_plugins.sh`. `make build` fails with _No rule to make target 'build-plugins'_.  
> CI no longer calls `make build`, so it is not covered.

Fixed in 10927fb22a19f77dbf126a502f0f97de17d82cd3

>   **`tests/Makefile:18`** : `LZMQGSRVVERSION ?= qgis-ltr-eager`, but that tag does not exist on Docker Hub (`3liz/qjazz` publishes `qgis-3.40`, `qgis-3.44`, `qgis-ltr-unstable`,  `qgis-4.2`…). `make up` fails with _manifest unknown_. The e2e matrix overrides the variable (`LZMQGSRVVERSION: ${{ matrix.qgis-server }}`), so the default value is never exercised.

Fixed in 61130822420a81048556614828d47fd516ae0202,  use `qgis-3.44` now as default tag.
 
>   **`tests/Makefile:77`** : `up`, `stop` and `reset` now force `--profile=dev`, which pulls in the `openldap` service. Its Dockerfile is based on `debian:buster-slim`, now EOL and removed from `deb.debian.org`, so the build fails on `404 Not Found` for every apt repository. CI runs a plain `docker compose up -d` with no profile, so this Dockerfile is never built there.

We are checking it: it seems to be an older issue.

dmarteau:
> We are checking it: it seems to be an older issue.

Fixed

PR #7096, opened 2026-08-31
Fix/popup compact table ragged rows

meyerlor:
The compact table listing all the features of a children layer built its header from the first feature only, while each body row skipped the cells whose value was empty for that very feature. As soon as one feature had more empty fields than the first one, its row had fewer cells than the header, and the DataTables instance created on that table in map.js raised "Requested unknown  parameter 'N' for row N, column N".

Features having the same number of non empty fields, but not the same ones, were silently rendered under the wrong columns.

Compute the displayed columns from all the features, keeping a column as soon as one feature has a value for it, and give every row one cell per column so all the rows have the same length.

### DIFF
--- a/CHANGELOG-3.11.md
+++ b/CHANGELOG-3.11.md
@@ -18,4 +18,8 @@ with some extra keywords: backend, tests, test, translation, funders, important
 
 ### Changed
 
+### Fixed
+
+* Popup - Compact table of children features: build the columns from all the displayed features instead of the first one only, so features having more empty fields no longer produce rows with missing cells and a DataTables `Requested unknown parameter` warning
+
 ### Backend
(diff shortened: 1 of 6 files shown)


CHANGELOG-3.11.md
```diff
@@ -18,4 +18,8 @@ with some extra keywords: backend, tests, test, translation, funders, important
 
 ### Changed
 
+### Fixed
+
+* Popup - Compact table of children features: build the columns from all the displayed features instead of the first one only, so features having more empty fields no longer produce rows with missing cells and a DataTables `Requested unknown parameter` warning
+
```
rldhont: To help backport, update CHANGELOG-3;9.md

rldhont (changes requested):
Good catch

rldhont (approved):

PR #7097, opened 2026-08-31
fix(edition): save the parent foreign key of a read-only child field

meyerlor:
When the foreign key of a 1:n relation is not editable in the QGIS project, the control is set read-only. jForms then renders a disabled select, which the browser does not submit, and skips read-only controls in initFromRequest() anyway. Creating a child feature from a popup silently stored an empty foreign key, leaving the child unlinked without any error.

The child form now sends the value in a dedicated liz_parent_fk parameter, and the controller applies it after initFromRequest(), only when inserting. The field name is checked against the relations of the project, so that an arbitrary column cannot be set through this parameter, and the value still goes through the usual form check.

### DIFF
--- a/assets/src/legacy/edition.js
+++ b/assets/src/legacy/edition.js
@@ -1676,6 +1676,25 @@ var lizEdition = function() {
 
                     }
                 } else {
+                    // Always send the foreign key value in a dedicated hidden input.
+                    // The control itself cannot be relied on: when the field is not
+                    // editable in the QGIS project, jForms renders a disabled select,
+                    // which the browser does not submit, and jForms ignores request
+                    // values of read-only controls anyway. The child feature would
+                    // then be created with an empty foreign key, without any error.
+                    if ( editionType == 'createFeature'
+                      && parentFeatProp !== null && parentFeatProp !== undefined ) {
+                        const hiddenFields = form[0].querySelector('div.jforms-hiddens');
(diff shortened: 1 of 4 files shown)


assets/src/legacy/edition.js
```diff
@@ -1676,6 +1676,21 @@ var lizEdition = function() {
...
+                    // values of read-only controls anyway. The child feature would
+                    // then be created with an empty foreign key, without any error.
+                    if ( editionType == 'createFeature'
+                      && parentFeatProp !== null && parentFeatProp !== undefined ) {
+                        var parentFkInput = $('<input type="hidden"></input>')
+                            .attr('id', 'liz_parent_fk_hidden')
+                            .attr('name', 'liz_parent_fk')
+                            .attr('value', relationRefField + ':' + parentFeatProp);
+                        form.find('div.jforms-hiddens').append(parentFkInput);
+                    }
```
mdouchin: Please use vanilla JS instead of jQuery
I know the existing code still uses it, but let not add more ;-)

mdouchin (changes requested):
Good catch.
Some minor changes needed.
I will add another reviewer

meyerlor:
> Good catch. Some minor changes needed. I will add another reviewer

lost some of my (already sparse) hair about this :D

rldhont (approved):

PR #7120, opened 2026-09-03
Fix WMTS Layer without TileMatrixSetLimits

rldhont:
The WMTS Capabilities build by lizmap could contain the Layer even if the layer's extent is outside the project's extend that define the tile matrices. In this case, the layer's TileMatrixSetLimits may be empty which causes an OpenLayers error that freezes the Lizmap's map building.

Funded by [SBE](https://www.sbe-engineering.com/)


mdouchin (approved):
Looks good to me !

PR #7141, opened 2026-09-14
Action - Fix action parameter value not passed to lizmap_get_data if empty (or 0)

mdouchin:
We sometimes need to pass action parameter values which are empty strings or 0 integer.

Funded by 3Liz


mdouchin:
@rldhont all checks are green

rldhont (approved):

PR #7143, opened 2026-09-15
Security fixes

dmarteau:
Test lizmap-servers security fixes


rldhont:
Tests with expression failed

dmarteau:
@rldhont @mdouchin There is two errors surfacing:

* Playwright test tagged `@readonly` [39]:  I cannot tell what this test does and why it is failing.
* Playwright test tagged `@write` [14]:  Connected to the fact that  now errors on expression always return a 400. I don't know is the test a LWC feature or an isolated test  that can be fixed.

dmarteau:
> Tests with expression failed

I know that.    

The question is:  It is something thas can be fixed by changing only the test or is this a LWC feature that need to be refactored in LWC ?

dmarteau:
@rldhont Let's put things that way:  Is LWC ready to handle response 400 on the whole request instead dealing with partial response ?

dmarteau:
And I absolutely cannot figure at all where are the paramaters for the requests in the tests......

dmarteau:
## VIRTUALFIELD test failure.

I have fixed the `VIRTUAFIELD` test failure by keeping the legacy behavior of the `VIRTUALFIELD`  request: not allowed expressions are reported in the response instead triggering a 400 Bad Request error.

## Form widget test failure

The test `Text widget in a form` (`tests/end2end/playwright/edition-form.spec.js`)  has been disabled because its
assumption is wrong: QGIS fail silently when replacing tex.

Security fixes in Lizmap server does not allow to fail silently and return an error 400 Bad Requestt with expression evaluation and return the template expression instead issuing an error.   

This test must be fixed accordingly and LWC must handle the error correctly.

dmarteau:
@rldhont @mdouchin Can you check the  `Form widget` test (see above) on how to adapt it for handling correctly a error response or remove the invalid expression fra the test data/project and replace it with a valid one ?   

 This test is actually flawed by the fact it handles a response which is invalid.

PR #7174, opened 2026-09-22
Proxy - Increase client timeout for requests sent to QGIS Server

mdouchin:
The Proxy class in charge of sending the requests to QGIS Server used a timeout calculated from the PHP `max_execution_time` (default 30s) .

Since the network request time is not part of the PHP execution time, we can set a timeout higher than the PHP `max_execution_time`, by calculating it based on the `default_socket_timeout` (default 60s).

Funded by 3liz


rldhont (approved):

rldhont:
Good catch, thanks !

